PR #306, opened 2025-08-26
bug(#305): update phino to 0.0.0.40 + added max-cycles flag

maxonfjvipon:
Closes: #305


src/main/resources/org/eolang/hone/scaffolding/entry.sh
```diff
@@ -130,6 +130,7 @@ fi
     "-Dexec.phino.timeout=${TIMEOUT}" \
     "-Dexec.phino.threads=${THREADS}" \
     "-Dexec.phino.max-depth=${MAX_DEPTH}" \
+    "-Dexeclphino.max-cycles=${MAX_CYCLES}" \
```
yegor256, line 133: @maxonfjvipon there is a typo

Issue #449, opened 2026-03-23
default grep-in `(66-69-6C-74-65-72|6D-61-70)` permits classes without `map` and `filter`


volodya-lombrozo:
I'm trying to optimise a java projext using `hone-maven-plugin`. I'm using the following grep-in config to avoid unnecessary attempts to optimise irrelevant classes:

```java
.set("grepIn","(66-69-6C-74-65-72|6D-61-70)");
```

It works fine, that said not always. For example, I have the following java class:

```java
class X {
    public static void main(String[] a) {
      byte[] r = new byte[] {(byte) 0x01, (byte) 0x02};
      System.out.println(r);
    }
}
```

and `phino` is trying to optimise it:

```
[INFO] >> [INFO] --- exec:3.6.3:exec (default-cli) @ hone ---
[INFO] >> Hone version: 0.0.0
[INFO] >> Phino version: 0.0.0.43
[INFO] >> Timeout: 999 seconds
[INFO] >> Using 44 rewriting rule(s)
[INFO] >> Grep-in: (66-69-6C-74-65-72|6D-61-70)
[INFO] >> Starting to rewrite 1 file(s)...
[INFO] >> No changes in 1/1 X.phi:  24K, 313 lines, 2685 lps
[INFO] >> Finished rewriting 1 file(s) in 0 seconds
```

But we don't have any `map` and `filter` methods here.
Most probably, this happens due the following content in `X.xmir` (jeo-disassemble):

```xml
<o base="Φ.org.eolang.string" name="name">
  <o as="α0" base="Φ.org.eolang.bytes">
    <o as="α0">6D-61-70-70-65-64-2F-58</o>
  </o>
</o>
```

Maybe we can invent a bit more sophisticatged filtering, to avoid such cases?

bibonix:
Submitted #487 with a fix. The default `hone.grep-in` regex was unanchored, so its hex bytes for `"map"` (`6D-61-70`) and `"filter"` (`66-69-6C-74-65-72`) matched as a prefix/suffix of any longer string whose bytes happened to contain them — including `"mapped/X"` and `"filtered"`, exactly the case reported here.

The PR:

1. Exposes the default as `OptimizeMojo.DEFAULT_GREP_IN` and adds four unit tests in `OptimizeMojoTest` — two happy-path (`matches…`) and two that reproduce the false positives (`ignores…`). The two `ignores…` tests fail against current `master`.
[...]

PR #487, opened 2026-04-23
#449: anchor default grep-in at XML tag boundaries

bibonix:
Addresses #449.

## Root cause

The default `grep-in` regex was `(66-69-6C-74-65-72|6D-61-70)` — the hex bytes of `"filter"` and `"map"` with no boundary anchors. In XMIR, strings are emitted as byte sequences inside `<o>` elements, e.g.
```xml
<o as="α0">6D-61-70-70-65-64-2F-58</o>  <!-- "mapped/X" -->
```
The unanchored alternation matches `6D-61-70` as a prefix of `"mapped/X"`, so phino happily rewrites classes that never call `Stream.map()` / `Stream.filter()` — which is exactly the scenario @volodya-lombrozo reported.

## Fix

Anchor each alternative between `>` and `<` so a match only fires when the hex bytes are the entire content of an `<o>` element — i.e. a standalone UTF-8 method name in XMIR. A byte block that merely starts or ends with the target bytes now has a trailing/leading `-` that breaks the match.

`DEFAULT_GREP_IN` is now `>(66-69-6C-74-65-72|6D-61-70)<`. The new Javadoc on the constant spells out the reasoning for the next reader.

## Tests

The default is exposed as the package-private constant `OptimizeMojo.DEFAULT_GREP_IN` so that unit tests can assert on it directly (the `@Parameter` annotation is `RetentionPolicy.CLASS` and not visible via reflection at runtime).

Four new unit tests in `OptimizeMojoTest`:
- `matchesStandaloneMapByteSequenceInDefaultGrepIn` — a standalone `<o>6D-61-70</o>` still matches (happy path).
- `matchesStandaloneFilterByteSequenceInDefaultGrepIn` — same for `"filter"` bytes.
- `ignoresMapBytesEmbeddedInLongerStringInDefaultGrepIn` — `<o>6D-61-70-70-65-64-2F-58</o>` (the `"mapped/X"` case from #449) does **not** match. Fails on `master`, passes with the fix.
- `ignoresFilterBytesEmbeddedInLongerStringInDefaultGrepIn` — the symmetric `"filtered"` case.

The first commit adds the tests (with the old regex still in place) so the bug is reproducible on the tip; the second commit flips the constant to the anchored regex and the tests go green.

## Verification

- `mvn test` — 26 tests, all green locally (1 skipped because Docker isn't available in this environment).
- `mvn -Pqulice verify -DskipTests -DskipITs` — clean.

### DIFF
--- a/src/main/java/org/eolang/hone/OptimizeMojo.java
+++ b/src/main/java/org/eolang/hone/OptimizeMojo.java
@@ -68,12 +68,17 @@ public final class OptimizeMojo extends AbstractMojo {
     /**
      * Default value for {@link #grepIn}.
      *
-     * <p>Exposed as a constant so that tests can reference the exact regex
-     * that is applied when the user does not override <tt>hone.grep-in</tt>.</p>
+     * <p>Anchors the hex-byte alternatives between <tt>&gt;</tt> and
+     * <tt>&lt;</tt> (the XML tag boundaries that delimit a single PHI bytes
+     * literal inside a {@code .xmir} file) so that the bytes of
+     * <tt>"map"</tt> or <tt>"filter"</tt> only match when they are the
+     * <em>entire</em> content of an {@code <o>} element — not when they
+     * happen to appear as a prefix or suffix of a longer byte sequence
+     * (e.g. <tt>"mapped/X"</tt> or <tt>"filtered"</tt>). See issue #449.</p>
      *
      * @since 0.19.0
      */
-    static final String DEFAULT_GREP_IN = "(66-69-6C-74-65-72|6D-61-70)";
+    static final String DEFAULT_GREP_IN = ">(66-69-6C-74-65-72|6D-61-70)<";
 
     /**
      * Location of <tt>.class</tt> files to optimize inside
--- a/src/main/java/org/eolang/hone/OptimizeMojo.java
+++ b/src/main/java/org/eolang/hone/OptimizeMojo.java
@@ -66,6 +66,16 @@ import org.cactoos.text.TextOf;
 public final class OptimizeMojo extends AbstractMojo {
 
     /**
+     * Default value for {@link #grepIn}.
+     *
+     * <p>Exposed as a constant so that tests can reference the exact regex
+     * that is applied when the user does not override <tt>hone.grep-in</tt>.</p>
+     *
+     * @since 0.19.0
+     */
+    static final String DEFAULT_GREP_IN = "(66-69-6C-74-65-72|6D-61-70)";
+
+    /**
      * Location of <tt>.class</tt> files to optimize inside
      * the {@code target} directory.
      *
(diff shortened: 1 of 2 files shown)


bibonix:
@yegor256 — ready for review.

**TL;DR:** the default `grep-in` regex was missing boundary anchors, so the hex bytes of `"map"` matched as a prefix of `"mapped/X"` (and `"filter"` as a prefix of `"filtered"`), making phino rewrite classes that don't use `Stream.map()` or `Stream.filter()`. This PR anchors the two alternatives between `>` and `<` — the XML tag boundaries that delimit a standalone UTF-8 byte sequence inside an `<o>` element in XMIR — so a match only fires when the bytes are the entire content of the element.

**Why this shape of fix:**

- `jeo-disassemble` emits UTF-8 constants (method names, class names, string literals) as one `<o as="α0">XX-XX-...</o>` element per constant, with the hex bytes hyphen-joined. A prefix-only match against the bytes of `"map"`/`"filter"` therefore fires spuriously on any longer string whose bytes happen to start with those three/six bytes — which is exactly what @volodya-lombrozo reported for `"mapped/X"`.
- The Javadoc on `grepIn` already documented the intended form (`<o>...</o>`), so wrapping the alternation in `>...<` brings the default in line with its own documentation.
- Anchoring keeps the cheap-filter property: it still needs a single `grep -qE` pass over the XMIR and rejects files that don't contain a `map`/`filter` method-name constant.

**Commits:**

- [`c574835b`](https://github.com/objectionary/hone-maven-plugin/pull/487/commits/c574835b) — expose `OptimizeMojo.DEFAULT_GREP_IN` and add four unit tests (`matches…` / `ignores…`). The two `ignores…` tests fail on `master`, pinning the bug.
- [`f09d2a79`](https://github.com/objectionary/hone-maven-plugin/pull/487/commits/f09d2a79) — flip the constant to `>(66-69-6C-74-65-72|6D-61-70)<`; all four tests pass. Javadoc on the constant now spells out why the anchors are there.

**Verification:**

- `mvn test` locally — 26/26 green (1 Docker test skipped locally because Docker isn't available in the sandbox).
- `mvn -Pqulice verify -DskipTests -DskipITs` — clean.
- CI on this PR — 17/17 green (shellcheck, bashate, copyrights, typos, xcop, markdown-lint, make, mvn 17, mvn 22, pdd, mvn default, qulice, docker, reuse, hadolint, yamllint, actionlint).

No other parameter defaults changed; users that override `grepIn` are unaffected.

0crat:
@bibonix Hey! Nice work on that contribution – you've earned **+16 points** this time! 🎉 The policy rewards solid code contributions like yours, and you're doing great by keeping things clean and reviewable. Your running score is now [+32](https://www.eolang.org/zerocracy/objectionary-vitals.html), so you're building up nicely! Just remember that staying under 100 hits-of-code and getting proper reviews helps maximize your points. [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the contributions flowing, and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates!

PR #493, opened 2026-04-29
Fix #492: use getgid() for Docker --user, not geteuid()

### LINKED ISSUE #492: `whoami()` passes `geteuid()` instead of `getgid()` in Docker `--user` flag
## What Happens

`whoami()` in `OptimizeMojo.java` passes `geteuid()` (effective user ID) as the group ID in the Docker `--user` flag, instead of `getgid()` (group ID).

The method at line 590:

```java
private static String whoami() {
    return String.format(
        "%d:%d",
        OptimizeMojo.CLibrary.INSTANCE.getuid(),
        OptimizeMojo.CLibrary.INSTANCE.geteuid()  // wrong: should be getgid()
    );
}
```

The method's own Javadoc says `@return A string in the format "uid:gid"`, but `geteuid()` is the effective user ID, not the group ID. Docker receives `--user uid:euid` instead of `--user uid:gid`.

On most systems `euid == uid`, so the bug is silent. On systems where `euid != uid` (setuid processes, some CI environments), Docker gets `--user 1000:0` or similar — wrong group, causing permission errors on group-owned files inside the container.

Additionally, `getgid()` is not declared in `CLibrary` at all:

```java
public interface CLibrary extends Library {
    int getuid();
    int geteuid();
    // getgid() is missing
}
```

## What Should Happen

`getgid()` should be declared in `CLibrary` and used in `whoami()`, consistent with the standard Docker idiom `--user $(id -u):$(id -g)`:

```java
public interface CLibrary extends Library {
    int getuid();
    int geteuid();
    int getgid();
}
```

```java
private static String whoami() {
    return String.format(
        "%d:%d",
        OptimizeMojo.CLibrary.INSTANCE.getuid(),
        OptimizeMojo.CLibrary.INSTANCE.getgid()
    );
}
```

bibonix:
@yegor256 this PR fixes #492 — `OptimizeMojo.whoami()` was passing `geteuid()` as the *group* component of the Docker `--user` flag, despite the method's own Javadoc saying the result is `uid:gid`.

## What changed

- `OptimizeMojo.CLibrary` now declares `int getgid()` (it was missing)
- `whoami()` is split into two: a `private static` production entry point and a package-private overload `whoami(CLibrary)` so the format can be exercised without `Native.load("c", ...)`
- the format now reads `lib.getgid()` for the second component
- `OptimizeMojoTest#formatsWhoamiAsUidColonGid` uses a fake `CLibrary` with distinct values for `uid` (1000), `euid` (2000), and `gid` (3000) — so the regression is caught even on the common case where `euid == uid` and the bug is silent

## Why it matters

Docker's `--user` flag expects `uid:gid`, not `uid:euid`. On most boxes `euid == uid` so nothing breaks, but on systems where they differ (setuid binaries, certain CI runners) Docker gets a wrong group — typically `1000:0` — which causes permission errors on group-owned files inside the container. This change makes the behaviour consistent with the standard `--user $(id -u):$(id -g)` idiom.

## Commits

1. `52ea3f9d` — refactors `whoami` for testability and adds the failing regression test (the `getgid()` declaration is added in this commit so the fake `CLibrary` can compile)
2. `801b7393` — one-line fix swapping `lib.geteuid()` for `lib.getgid()`

## Test plan

- [x] `mvn -Pqulice clean install -DskipTests -Dinvoker.skip` passes locally (no new lint violations)
- [x] `mvn test` passes locally — 27 unit tests, 1 skipped (Docker-only)
- [x] `formatsWhoamiAsUidColonGid` fails on commit 1 (`Expected: "1000:3000" but: was "1000:2000"`) and passes on commit 2
- [x] All 17 PR CI jobs green: actionlint, bashate, copyrights, docker, hadolint, make, markdown-lint, mvn (matrix 17, 22, rultor-image), pdd, qulice, reuse, shellcheck, typos, xcop, yamllint

### DIFF
--- a/src/main/java/org/eolang/hone/OptimizeMojo.java
+++ b/src/main/java/org/eolang/hone/OptimizeMojo.java
@@ -337,7 +337,7 @@ public final class OptimizeMojo extends AbstractMojo {
         return String.format(
             "%d:%d",
             lib.getuid(),
-            lib.geteuid()
+            lib.getgid()
         );
     }
 
--- a/src/main/java/org/eolang/hone/OptimizeMojo.java
+++ b/src/main/java/org/eolang/hone/OptimizeMojo.java
@@ -328,6 +328,20 @@ public final class OptimizeMojo extends AbstractMojo {
     }
 
     /**
+     * Return the user and group IDs of the current user, using the given C library.
+     *
+     * @param lib The C library to query for the IDs
+     * @return A string in the format "uid:gid"
+     */
+    static String whoami(final OptimizeMojo.CLibrary lib) {
+        return String.format(
+            "%d:%d",
+            lib.getuid(),
+            lib.geteuid()
+        );
+    }
+
+    /**
     * Check if the classes directory is absent or doesn't have classes.
     *
     * @return True if there are no class files, false otherwise.
@@ -588,11 +602,7 @@ public final class OptimizeMojo extends AbstractMojo {
      * @return A string in the format "uid:gid"
      */
     private static String whoami() {
-        return String.format(
-            "%d:%d",
-            OptimizeMojo.CLibrary.INSTANCE.getuid(),
-            OptimizeMojo.CLibrary.INSTANCE.geteuid()
-        );
+        return OptimizeMojo.whoami(OptimizeMojo.CLibrary.INSTANCE);
     }
 
     private String rulesAsString() {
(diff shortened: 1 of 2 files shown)


0crat:
@bibonix Hey! Nice work on that contribution – you just scored **+16 points**! 🎉 The bonus system rewards quality code with up to 24 points based on factors like code size, review feedback, and overall contribution value, and you nailed it. Your running score is now [+250](https://www.eolang.org/zerocracy/objectionary-vitals.html), so [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) those quality contributions flowing! Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for the latest updates.

PR #501, opened 2026-05-08
Default `threads` to 0 so the plugin uses all CPU cores out of the box

### LINKED ISSUE #500: default number of threads should be zero
Let's set it to zero and fix it with a test.

yegor256:
## Summary
- Closes #500.
- Change the default value of the `threads` parameter from `1` to `0`. With `0`, `rewrite.sh` already falls back to `$(nproc)`, so the plugin now uses every available CPU by default — a much more sensible choice for most users.
- Refresh the Javadoc accordingly.
- Add a unit test that reads the generated Maven plugin descriptor (`META-INF/maven/plugin.xml`) and asserts the default value of `threads` is `0`, guarding against future regressions.

## Test plan
- [x] `mvn -DexcludedGroups=deep -DskipITs test` — all 28 unit tests pass.
- [x] `mvn -Pqulice -DskipTests -DskipITs verify` — qulice and jtcop pass clean.
- [x] New test `OptimizeMojoTest#hasZeroDefaultThreadsInPluginDescriptor` was first committed alone and confirmed to fail against the old `defaultValue = "1"`, then passes after the source change.

### DIFF
--- a/src/main/java/org/eolang/hone/OptimizeMojo.java
+++ b/src/main/java/org/eolang/hone/OptimizeMojo.java
@@ -243,17 +243,16 @@ public final class OptimizeMojo extends AbstractMojo {
     /**
      * How many threads to use for rewriting?
      *
-     * <p>By default, it's a single-threaded process. However, it may be
-     * rather slow for large projects. It is recommended to use values
-     * that are close to the number of CPUs you have on your machine.</p>
-     *
-     * <p>If you set this parameter to zero, the number of threads
-     * will be set to the number of CPUs on the machine.</p>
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 **Great work on your contribution!** 🎉 You've earned **+4 points** (+16 base, -8 for missing code review, -4 for 31 hits-of-code). Your current score is [+89](https://www.eolang.org/zerocracy/objectionary-vitals.html) - [keep the momentum going](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) and don't forget to check your [Zerocracy account](https://www.zerocracy.com/)!

PR #522, opened 2026-05-15
#521: Handle header-only hone-statistics.csv without crashing

### LINKED ISSUE #521: CSV.java crashes with IndexOutOfBoundsException when hone-statistics.csv has only a header
## Summary

`org.eolang.hone.CSV` throws `IndexOutOfBoundsException: Index: 0, Size: 0` whenever `target/hone-statistics.csv` contains only the header line and no data rows. This happens any time a build runs the `optimize` mojo but `rewrite.sh` writes no statistics rows — e.g. when every input class is skipped by `grep-in`, or when there is nothing to rewrite.

## Reproduce

Run the deep test suite:

```
mvn test -Pdeep
```

Several tests fail with the same root cause, for example:

- `OptimizeMojoTest.optimizesExecutableJavaApp` — uses `rules=none`, the Hello/Main classes contain no `map`/`filter`, so the default `grep-in` skips them.
- `OptimizeMojoTest.optimizesTwice` — same pattern.
- `OptimizeMojoTest.optimizesAsSpecifiedInYamlPack[4]` (the `streams-rand.yml` pack) — the fixture explicitly asserts `No grep-in match for 1/1 X.xmir`, so by design no rows are written.

Maven error:

```
[ERROR] Failed to execute goal org.eolang:hone-maven-plugin:...:optimize (default) on project test: ... Index: 0, Size: 0
Caused by: java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
    at java.util.LinkedList.checkElementIndex (LinkedList.java:567)
    at java.util.LinkedList.get (LinkedList.java:488)
    at org.cactoos.list.ListEnvelope.get (ListEnvelope.java:46)
    at org.eolang.hone.CSV.<init> (CSV.java:59)
    at org.eolang.hone.CSV.<init> (CSV.java:49)
    at org.eolang.hone.OptimizeMojo.optimize (OptimizeMojo.java:386)
```

## Root cause

`src/main/java/org/eolang/hone/CSV.java:57-62`:

```java
private CSV(final List<CSVRecord> records) {
    this(
        new ListOf<>(records.get(0).toMap().keySet()),
        CSV.rows(records)
    );
}
```

`records.get(0)` is invoked unconditionally. `CSV.from` parses the CSV with `CSVFormat.DEFAULT.builder().setHeader().setSkipHeaderRecord(false)`. With `setHeader()` (no args) the first row is consumed as the header map, so a file containing only `ID,Before,After,Changed,LinesPerSec\n` parses to zero records and `get(0)` throws.

The script intentionally writes the header even when no files are rewritten (see `statistics_header` in `src/main/resources/org/eolang/hone/scaffolding/rewrite.sh`), and skips `statistics_row` for every file filtered out by `grep-in`. Producing a header-only CSV is a normal, expected state — the Java side should handle it.

## Suggested fix

Either:

1. In `CSV` derive the header list from `CSVParser.getHeaderNames()` instead of from the first record, so an empty body is fine; or
2. In `OptimizeMojo.optimize` (`src/main/java/org/eolang/hone/OptimizeMojo.java:384`), skip CSV parsing and the `Optimized %d/%d files` log line when the CSV has only the header.

Option 1 is preferable — it keeps the responsibility inside `CSV` and matches the obvious semantics ("a CSV with headers and no rows is still a valid CSV with size 0").

A regression test belongs in `CSVTest` covering the header-only file case.

## Impact

Six `@Tag("deep")` tests are blocked by this on the current `517` branch, and the same scenarios will hit any consumer whose build skips every file via `grep-in`.

yegor256:
## Summary

Fixes #521.

`CSV` previously derived its header list from `records.get(0).toMap().keySet()`. When `hone-statistics.csv` contains only the header row (a normal state when every input class is filtered out by `grep-in`, or there is nothing to rewrite), the records list is empty and `get(0)` threw `IndexOutOfBoundsException`, breaking the `optimize` mojo and the `@Tag("deep")` tests.

The fix follows option 1 in the issue: headers now come from `CSVParser.getHeaderNames()`, so a header-only CSV parses to a `CSV` with zero data rows — which is the obvious semantics ("a CSV with headers and no rows is still a valid CSV with size 0").

## Changes

- `src/main/java/org/eolang/hone/CSV.java`
  - `from(Path)` now returns a fully-built `CSV` whose headers come from `CSVParser.getHeaderNames()` and whose rows come from `parser.getRecords()`. The parser is closed via try-with-resources alongside the reader.
  - Replaced the private `CSV(List<CSVRecord>)` constructor with a delegating `CSV(CSV)` so `CSV(Path)` stays a one-liner.
- `src/test/java/org/eolang/hone/CSVTest.java`
  - Added `parsesHeaderOnlyCsv` covering the previously-crashing case.

## Test plan

- [x] `mvn test -Dtest=CSVTest` — 3/3 pass, including the new `parsesHeaderOnlyCsv`.
- [x] `mvn test` — full unit suite green (30 run, 1 skipped due to no Docker socket in this env).
- [ ] `mvn test -Pdeep` — should now unblock `OptimizeMojoTest.optimizesExecutableJavaApp`, `optimizesTwice`, and the `streams-rand.yml` pack (requires Docker; not runnable here).

---
_Generated by [Claude Code](https://claude.ai/code/session_018w2ZwTq6CNHYL13ZjShd4V)_

### DIFF
--- a/src/main/java/org/eolang/hone/CSV.java
+++ b/src/main/java/org/eolang/hone/CSV.java
@@ -19,6 +19,7 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVParser;
 import org.apache.commons.csv.CSVPrinter;
 import org.apache.commons.csv.CSVRecord;
 import org.cactoos.list.ListOf;
@@ -51,14 +52,11 @@ public final class CSV {
 
     /**
      * Constructor.
-     * @param records CSV records
-     * @checkstyle ConstructorsCodeFreeCheck (5 lines)
+     * @param other The CSV to copy fields from
+     * @checkstyle ConstructorsCodeFreeCheck (3 lines)
      */
-    private CSV(final List<CSVRecord> records) {
-        this(
-            new ListOf<>(records.get(0).toMap().keySet()),
-            CSV.rows(records)
-        );
+    private CSV(final CSV other) {
+        this(other.headers, other.records);
     }
 
     /**
@@ -158,20 +156,21 @@ public final class CSV {
     }
 
     @SuppressWarnings("PMD.UnnecessaryLocalRule")
-    private static List<CSVRecord> from(final Path csv) {
+    private static CSV from(final Path csv) {
         try (
             Reader reader = new InputStreamReader(
                 Files.newInputStream(csv),
                 StandardCharsets.UTF_8
-            )
+            );
+            CSVParser parser = CSVFormat.DEFAULT.builder()
+                .setHeader()
+                .setSkipHeaderRecord(false)
+                .get()
+                .parse(reader)
         ) {
-            return new ListOf<>(
-                CSVFormat.DEFAULT.builder()
-                    .setHeader()
-                    .setSkipHeaderRecord(false)
-                    .get()
-                    .parse(reader)
-                    .iterator()
+            return new CSV(
+                new ListOf<>(parser.getHeaderNames()),
+                CSV.rows(parser.getRecords())
             );
         } catch (final IOException exception) {
             throw new IllegalStateException(
--- a/src/test/java/org/eolang/hone/CSVTest.java
+++ b/src/test/java/org/eolang/hone/CSVTest.java
@@ -62,4 +62,28 @@ final class CSVTest {
             Matchers.is(0)
         );
     }
+
+    @Test
+    void parsesHeaderOnlyCsv(@Mktmp final Path temp) throws Exception {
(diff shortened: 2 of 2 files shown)


0crat:
@yegor256 Hey there! Nice work on the contribution – you've scored **+8 points** this time: **+16** as the base reward, but **-8** got deducted since there wasn't a code review (we really need those reviews to keep quality high!). [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the contributions flowing, and remember that getting your code reviewed can save you those penalty points. Your running score is now [+142](https://www.eolang.org/zerocracy/objectionary-vitals.html) – don't forget to peek at your [Zerocracy account](https://www.zerocracy.com/) when you get a chance! 🚀

Issue #537, opened 2026-05-18
Optimize .peek(Consumer) — pass-through observer in stream pipelines


yegor256:
## The operation

`Stream.peek(Consumer<T>)`, `IntStream.peek(IntConsumer)`, etc. — a non-terminal pass-through that invokes the consumer for each element and re-emits the same element downstream. Signature shape:

```java
Stream<T> peek(Consumer<? super T> action);
```

## Current behavior

No rule recognizes `.peek(...)`. The `Φ.hone.lambda` produced by 111 has `stream.method = "peek"` and flows untouched through the entire 2xx–5xx pipeline. At 7xx it's emitted back as raw `invokedynamic` + `invokeinterface Stream.peek`. So a chain like:

```java
.map(...).peek(System.out::println).filter(...)
```

…stays as three separate stream stages instead of fusing into one `mapMulti`.

## Why it matters

`peek` is the canonical idiom for **inline logging and side-effects** in stream pipelines. Real codebases that use streams heavily almost always have at least one `.peek(...)` for observability. Today it's an opaque wall that prevents fusion of operations on either side of it.

## Sketch

Add `Φ.hone.peek` (analogous to `Φ.hone.map`), then a `peek-to-distill` rule that folds it into the uniform `Φ.hone.distill` form. The distill body just invokes the consumer with the input element and then re-pushes the same element to the downstream consumer — no value transformation. With that, `401-fuse` collapses adjacent map/filter/peek chains uniformly.

PR #544, opened 2026-05-18
#539: Recognize Stream.takeWhile lambdas as Φ.hone.take-while

### LINKED ISSUE #539: Optimize .takeWhile(Predicate) — short-circuit prefix
## The operation

`Stream.takeWhile(Predicate<T>)`, `IntStream.takeWhile(IntPredicate)`, etc. — added in Java 9. Emits elements until the predicate first returns false, then short-circuits. Signature shape:

```java
Stream<T> takeWhile(Predicate<? super T> predicate);
```

## Current behavior

No recognition rule. The `Φ.hone.lambda` with `stream.method = "takeWhile"` propagates untouched through the pipeline. Adjacent transforms can't fuse across it.

## Why it matters

`takeWhile` is increasingly common for **ordered-stream early-termination** — sorted/iterate sources, log scanning, paginated inputs. It's a peer of `.filter` semantically (also takes a `Predicate`) but with sticky state (once-false-always-false). Streams ending in `.takeWhile(...).findFirst()` or `.takeWhile(...).count()` are a common idiom this pipeline currently optimizes only partially.

## Sketch

Add `Φ.hone.take-while` recognized analogously to `.filter` (201). Folding into `distill` is non-trivial because the operation has internal mutable state (a one-shot \"done\" flag). One option is a dedicated `Φ.hone.distill-with-flag` variant; another is keeping `take-while` as its own distill family that doesn't fuse with stateless ops but DOES fuse with adjacent `.take-while` / `.drop-while` of compatible direction.

yegor256:
## Summary

Resolves #539. Adds takeWhile recognition AND lowering to `Stream.mapMulti` with a captured guard cell.

### Recognition / round-trip safety (recognise → unrecognise)

- **213-lambda-to-take-while.phr** — recognizes `Φ.hone.lambda` whose stream call is `Stream.takeWhile(Predicate)` and rewrites it as a `Φ.hone.take-while` pragma (analog of 201 for `.filter`). _Renamed from 209 after master added `209-lambda-to-peek.phr`._
- **214-lambda-to-primitive-take-while.phr** — same for `{Int,Long,Double}Stream.takeWhile(...)` → `Φ.hone.primitive-take-while` (analog of 203). _Renamed from 210._
- **605-take-while-to-lambda.phr** — fallback unrecognise (fires if 351 didn't consume the pragma).
- **606-primitive-take-while-to-lambda.phr** — same for the primitive pragma.

### Lowering to mapMulti (the actual optimization)

- **351-take-while-to-mapMulti.phr** — folds object-Stream `Φ.hone.take-while` into:
  - Raw setup bytecode (`iconst_1` + `newarray boolean`) allocating a fresh `boolean[1]` guard on the operand stack
  - A `Φ.hone.lambda` that creates a `BiConsumer<Item, Consumer<Item>>` capturing that guard, then invokes `Stream.mapMulti(BiConsumer)`
  - A synthetic private static wrapper method that implements once-false-always-false semantics: short-circuits if `guard[0]` is set; on a `false` predicate result sets `guard[0] = true` and drops; on `true` forwards to the `Consumer`.
  - `bridge-signature` is the instantiated SAM `(L<item>;Ljava/util/function/Consumer;)V`; `target.signature` is the wrapper impl descriptor `([ZL<item>;Ljava/util/function/Consumer;)V` (captures live only in `interface` per `LambdaMetafactory`).

The primitive variant (214/606) is currently only recognise/unrecognise; lowering for primitive streams is left to a follow-up.

### Tests

- Phino unit tests for rules 213, 214, 605, 606.
- **streams-take-while.yml** end-to-end test confirming the substitution: the source contains `100` mid-stream so `takeWhile(n -> n < 10)` must yield `count = 3` (a filter rewrite would yield 5). `newarray`/`baload`/`bastore = 1` each prove the guard cell was allocated and consulted.
- **streams-all-ops.yml** opcode counts updated for the `+1 invokestatic` and `+2 return` from the wrapper method (combined with master's #537 peek deltas).

## Test plan

- [ ] `src/test/phino/run.sh` passes (CI: phino workflow)
- [ ] `streams-take-while.yml` passes (mvn deep test)
- [ ] `streams-all-ops.yml` passes with updated counts
- [ ] All other lint / build workflows pass


0crat:
@yegor256 Thanks for your contribution! You've earned **+4 points** for this submission: **+16** as the base reward, **-4** for exceeding 100 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (644 total), **-8** for exceeding 400 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html), and **-8** for lacking code review, with a final adjustment to meet the minimum threshold. Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+425](https://www.eolang.org/zerocracy/objectionary-vitals.html) - don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #545, opened 2026-05-18
#537: Optimize .peek(Consumer) in stream pipelines

### LINKED ISSUE #537: Optimize .peek(Consumer) — pass-through observer in stream pipelines

yegor256:
Closes #537.

## Summary

Before, `Stream.peek(Consumer)` was unrecognized: the `Φ.hone.lambda`
emitted by 111 had `stream.method = "peek"` and flowed untouched through
the entire 2xx–5xx pipeline, leaving an opaque `invokedynamic` + raw
`invokeinterface Stream.peek` between any adjacent map/filter stages and
blocking fusion across it.

This patch teaches the pipeline how to fold peek into the uniform
`Φ.hone.distill` form so it can fuse with its neighbours, exactly the way
map and filter already do.

## Changes

- `streams/209-lambda-to-peek.phr` — recognises a `Stream.peek(Consumer)`
  lambda and rewrites it to `Φ.hone.peek`, analogous to `202-lambda-to-map`.
  Restricted to `static = "true"` targets (which is what 111 + 121 always
  produce for non-capturing lambdas; capturing peek lambdas stay raw, as
  before).
- `streams/307-peek-to-distill.phr` — folds `Φ.hone.peek` into the
  canonical `Φ.hone.distill` shape. The body is `dup; invokeXXX target`,
  so the element is preserved on the stack for the downstream stage while
  the consumer is invoked with a duplicate.
- `streams/281-insert-dup-before-filter.phr` and
  `streams/282-insert-dup-before-transformed-filter.phr` — add `peek` to
  the allowed `𝜏-one` operations so the pre-filter dup is still inserted
  when a `peek` sits immediately before a filter.
- `src/test/phino/209-lambda-to-peek.yml` and
  `src/test/phino/307-peek-to-distill.yml` — phino-level unit tests for
  the new rules.
- `src/test/resources/org/eolang/hone/optimize/streams-all-ops.yml` —
  bump expected `invokestatic` (22→23) and drop `invokedynamic` (16→15)
  to reflect the peek lambda being absorbed into the adjacent mapMulti.

## Test plan

- [ ] `phino` workflow exercises the two new phino tests.
- [ ] `deep` workflow re-checks the updated opcode tally for
      `streams-all-ops` once the PR lands on master.
- [ ] No existing test fixture changes are expected to break: peek is
      only used in `streams-all-ops.yml`, and `Foo.java` in the make
      workflow does not exercise peek.

https://claude.ai/code/session_01PtnoH3KL5kBcESPnGKBe81

---
_Generated by [Claude Code](https://claude.ai/code/session_01PtnoH3KL5kBcESPnGKBe81)_

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/209-lambda-to-peek.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/209-lambda-to-peek.phr
@@ -1,0 +1,87 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# Recognizes a Φ.hone.lambda whose stream call is Stream.peek(Consumer)
+# and converts it to Φ.hone.peek — analogous to lambda-to-map. peek is
+# a pass-through observer: the consumer is invoked with each element and
+# the same element is re-emitted downstream. The bridge-output, returned
+# and bridge-input are therefore all identical to the input type.
+name: lambda-to-peek
+pattern: |
+  ⟦
+    φ ↦ Φ.hone.lambda,
+    class ↦ 𝑒-class,
+    method ↦ "accept",
+    interface ↦ "()Ljava/util/function/Consumer;",
+    lambda-signature ↦ "(Ljava/lang/Object;)V",
+    bridge-signature ↦ 𝑒-bridge-signature,
+    static ↦ 𝑒-static,
+    opcode ↦ 𝑒-opcode,
+    target ↦ ⟦
+      handle ↦ 𝑒-target-handle,
+      class ↦ 𝑒-target-class,
+      method ↦ 𝑒-target-method,
+      signature ↦ 𝑒-target-signature,
+      is-interface ↦ 𝑒-target-is-interface
+    ⟧,
+    stream ↦ ⟦
+      class ↦ "java/util/stream/Stream",
+      method ↦ "peek",
+      signature ↦ "(Ljava/util/function/Consumer;)Ljava/util/stream/Stream;"
+    ⟧
+  ⟧
+result: |
+  ⟦
+    φ ↦ Φ.hone.peek,
+    opcode ↦ 𝑒-opcode,
+    class ↦ 𝑒-class,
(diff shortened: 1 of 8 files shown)


0crat:
@yegor256 Thanks for your contribution! You've earned **+4 points** for this submission: **+16** as the base reward, **+14.2** for your 284 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (capped at +16), **-4** for exceeding 100 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html), and **-8** for the lack of code review, with the final score capped at the minimum of **+4 points**. Your running score is [+429](https://www.eolang.org/zerocracy/objectionary-vitals.html) - please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the contributions coming and don't forget to check your [Zerocracy account](https://www.zerocracy.com/)!

PR #553, opened 2026-05-18
Recognize Stream.sorted(Comparator) as a barrier pragma

### LINKED ISSUE #549: Optimize .sorted() and .sorted(Comparator) — full-buffer stateful sort
`Stream.sorted()` and `Stream.sorted(Comparator)` (Java 8+) have no recognition rule in `src/main/resources/org/eolang/hone/rules/streams/`. The `Φ.hone.lambda` with `stream.method = "sorted"` propagates untouched through the pipeline, so adjacent transforms can't fuse across it. The call is exercised but not rewritten in `src/test/resources/org/eolang/hone/optimize/streams-all-ops.yml:15`.

`.sorted` is one of the most common stream operations — present in nearly every introductory tutorial — and shows up routinely in ranked or ordered pipelines (`stream.filter(...).sorted().limit(k)` for top-k, `stream.map(...).sorted(byDate)`). The `Comparator` overload is also one of the most-fingerprinted uses of method references (`Comparator.comparing(...)`, `Comparator.reverseOrder()`). A chain like `.map(...).sorted().map(...)` is three pipeline stages today; even a partial fold would collapse the surrounding stateless ops on each side.

`.sorted` is fully buffering — no element can be emitted until the entire upstream has been consumed — so it cannot be carried through the running `distill` chain the way a `.skip` counter or a `.distinct` seen-set can. The smallest viable fix is to recognize `.sorted` as a barrier in the `2xx` stage (a new `Φ.hone.sorted` carrying the comparator handle when the overload is present, captured the same way `209-lambda-to-peek.phr` captures its `Consumer`) so that `3xx` and `4xx` can fuse runs of stateless ops upstream and downstream of the barrier independently. Related: #539, #540, #546, #547.

yegor256:
Adds a recognition rule (212) that maps `Φ.hone.lambda` whose stream call is `Stream.sorted(Comparator)` to a fresh `Φ.hone.sorted` pragma, and the inverse rule (607) that converts any leftover `Φ.hone.sorted` back to a `Φ.hone.lambda` so the 7xx phase can lower it to `invokedynamic + invokeinterface`. Matching phino unit tests cover the lambda → sorted and sorted → lambda transforms; an end-to-end fixture exercises a real `.sorted(Comparator)` call through the whole pipeline.

`.sorted` is fully buffering and cannot fold into the running distill chain, but leaving the call as a generic `Φ.hone.lambda` meant the surrounding stateless ops in 3xx/4xx had no stable pragma name to recognize it as a barrier. Giving it its own pragma is the smallest step that lets adjacent runs of stateless ops be reasoned about independently on each side, the same shape we already use for `.peek` (209) and `.takeWhile` (213).

This change targets only the `.sorted(Comparator)` overload — the no-arg `Stream.sorted()` does not produce a lambda and is left for follow-up. Closes #549.

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/212-lambda-to-sorted.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/212-lambda-to-sorted.phr
@@ -1,0 +1,54 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# Recognizes a Φ.hone.lambda whose stream call is Stream.sorted(Comparator)
+# and converts it into a Φ.hone.sorted pragma — analogous to 213
+# (lambda-to-take-while). Sorted is a fully-buffering stateful barrier:
+# every upstream element must be consumed before any downstream element
+# can be emitted, so the operation cannot be folded into the running
+# distill chain the way a stateless map/filter/peek can. Recognising it
+# gives the operation a stable pragma name so the surrounding 3xx/4xx
+# phases can fuse runs of stateless ops on each side of the barrier
+# independently. Any Φ.hone.sorted left after the recognise/fuse phases
+# is lowered back to a Φ.hone.lambda by 607 so that 7xx can emit valid
+# bytecode. See issue #549.
+name: lambda-to-sorted
+pattern: |
+  ⟦
+    φ ↦ Φ.hone.lambda,
+    class ↦ 𝑒-class,
+    method ↦ "compare",
+    interface ↦ "()Ljava/util/function/Comparator;",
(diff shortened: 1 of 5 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this work. Here's the breakdown based on our [policy](https://www.yegor256.com/2014/11/14/hits-of-code.html): **+16** as a basis, **-4** for exceeding 100 hits-of-code (227), and **-8** for lacking code review. While this contribution shows good effort, consider breaking larger changes into smaller, reviewable pieces to maximize your points and maintain code quality. Your running score is [+441](https://www.eolang.org/zerocracy/objectionary-vitals.html) - [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming and don't forget to check your [Zerocracy account](https://www.zerocracy.com/)!

PR #555, opened 2026-05-18
Optimize Stream.dropWhile via mapMulti with boolean[1] guard

### LINKED ISSUE #540: Optimize .dropWhile(Predicate) — short-circuit suffix
## The operation

`Stream.dropWhile(Predicate<T>)` (Java 9+). Skips elements while the predicate is true, then emits everything from the first false onward. The mirror of `.takeWhile`. Signature shape:

```java
Stream<T> dropWhile(Predicate<? super T> predicate);
```

## Current behavior

No recognition rule. Flows through unoptimized as raw `invokedynamic` + `invokeinterface Stream.dropWhile`.

## Why it matters

`dropWhile` pairs with `.takeWhile` for **windowed slicing** of ordered streams (e.g., header-skipping in log/CSV pipelines, prefix-trimming in time-series). It also gates fusion of adjacent operations: a `.map(...).dropWhile(...).map(...)` chain that the existing pipeline would happily collapse into a single mapMulti is currently three stages.

## Sketch

Same shape as `.takeWhile` (#539) but with the inverted state machine. Like `takeWhile` it has internal state (a one-shot "dropping" flag), so naive distill-folding doesn't apply — a state-carrying distill variant or a dedicated fusion family is needed.

Related: #539.

yegor256:
Closes #540.

This adds optimization for `Stream.dropWhile(Predicate)` analogous to what
#539 did for `takeWhile`. Recognition rules `217` and `218` lift a
`Φ.hone.lambda` whose stream call is `dropWhile` into a dedicated
`Φ.hone.drop-while` / `Φ.hone.primitive-drop-while` pragma. The fold rule
`352` lowers the object-Stream variant into a `Stream.mapMulti()` call that
captures a `boolean[1]` guard cell — a synthetic private static wrapper
drops elements while the original predicate returns true, flips the cell
once the predicate first returns false, and from that moment on forwards
every element to the mapMulti `Consumer`. Unrecognise fallbacks `607` and
`608` turn unconsumed pragmas back into `Φ.hone.lambda` so the 7xx phase
lowers them to standard `invokedynamic + invokeinterface` pairs.

While wiring this up I hit a bug in rule `351` (the `takeWhile` fold): its
result drops the class-level `ρ ↦ ∅` marker that the pattern requires.
That made the all-ops test break with `Unknown opcode name: distill`
because `501-distill-to-mapMulti` could no longer match the modified
class. The fix restores `ρ ↦ ∅` in `351` and adds the same marker to the
new `352`. I also tweaked the `streams-take-while` end-to-end test (and
shaped the new `streams-drop-while` test the same way) so the default
`grep-in` filter — which targets classes containing `map` or `filter` —
actually matches, otherwise the rewriter skips the file and the rule
never runs. The all-ops opcode counts are bumped by `+1 invokestatic` and
`+1 return` for the new dropWhile wrapper.

The phino test suite covers each of the four new pattern rules in
isolation; the new end-to-end fixture confirms the rewritten bytecode
yields the correct `dropWhile` semantics (it returns elements only after
the first predicate failure, never just the matching ones a naive filter
would emit).

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/217-lambda-to-drop-while.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/217-lambda-to-drop-while.phr
@@ -1,0 +1,43 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+name: lambda-to-drop-while
+pattern: |
+  ⟦
+    φ ↦ Φ.hone.lambda,
+    class ↦ 𝑒-class,
+    method ↦ "test",
+    interface ↦ "()Ljava/util/function/Predicate;",
+    lambda-signature ↦ "(Ljava/lang/Object;)Z",
+    bridge-signature ↦ 𝑒-bridge-signature,
+    static ↦ 𝑒-static,
+    opcode ↦ 𝑒-opcode,
+    target ↦ ⟦
+      handle ↦ 𝑒-target-handle,
+      class ↦ 𝑒-target-class,
+      method ↦ 𝑒-target-method,
+      signature ↦ 𝑒-target-signature,
+      is-interface ↦ 𝑒-target-is-interface
+    ⟧,
+    stream ↦ ⟦
+      class ↦ "java/util/stream/Stream",
+      method ↦ "dropWhile",
+      signature ↦ "(Ljava/util/function/Predicate;)Ljava/util/stream/Stream;"
+    ⟧
+  ⟧
+result: |
+  ⟦
+    φ ↦ Φ.hone.drop-while,
+    opcode ↦ 𝑒-opcode,
+    class ↦ 𝑒-class,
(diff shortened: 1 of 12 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this submission: +16 as a basis; +0.05×662=+33.1 for [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (capped at +16); -4 for exceeding 100 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (662 >= 100); -8 for exceeding 400 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (662 >= 400); -8 for lack of code review; final score adjusted to minimum threshold of +4. Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+445](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #559, opened 2026-05-18
feat(streams): recognize Stream.limit(long) via ldc

### LINKED ISSUE #548: Optimize .limit(long) — no recognizer in src/main/resources/org/eolang/hone/rules/streams/
`Stream.limit(long)` and its primitive peers flow through the pipeline unrecognized. The `2xx` stage in `src/main/resources/org/eolang/hone/rules/streams/` (around `201-lambda-to-filter.phr`..`215-lambda-to-mapMulti.phr`) has recognizers for `filter`, `map`, `flatMap`, `unbox`, `box`, but none for `limit`, so a `Φ.hone.lambda` whose `stream.method = "limit"` reaches the `3xx` fold as a raw lambda and blocks fusion across it.

The cost is that `.skip(k).limit(n)` pagination chains and `Stream.iterate(...).limit(n)` bounded-source idioms leave `.limit` unfused between optimized neighbours — `.map(...).limit(k).map(...)` stays three stages instead of one mapMulti.

The smallest concrete step is a new `.phr` file under `streams/` at an unused `2xx` prefix (e.g. `216-lambda-to-limit.phr`), modelled on `201-lambda-to-filter.phr`, that rewrites a `Φ.hone.lambda` with `stream.method = "limit"` into a `Φ.hone.limit` pragma carrying the `long` argument. Folding into `distill` can follow in a separate issue; recognition alone unblocks neighbour rules.

Related: #539, #540, #546, #547.

yegor256:
Adds a recognizer for `.limit(long)` so the byte pair `ldc(Φ.jeo.long N) + invokeinterface java/util/stream/Stream.limit:(J)Ljava/util/stream/Stream;` is abstracted into a single `Φ.hone.limit` pragma. A companion 7xx rule lowers the pragma back to the same byte pair when nothing else consumes it, so the round-trip is observationally a no-op.

The motivation is on the ticket: the 2xx stage has recognizers for filter, map, flatMap, unbox, box, skip, sorted, takeWhile and dropWhile, but no recognizer for limit. As a result `.limit(N)` sits in the body as raw bytecode and blocks 3xx/4xx fusion across it — for example `.map(...).limit(k).map(...)` cannot collapse into one mapMulti. Recognition alone is enough to unblock the neighbour rules; the two new files mirror the existing skip family one-for-one (216 mirrors 220, 722 mirrors 720).

Folding `Φ.hone.limit` into the distill chain is deliberately out of scope here — that's the state-carrying fusion path and it deserves its own ticket. The mirror in 722 keeps the bytecode valid until that lands.

Closes #548

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/216-recognize-limit-ldc.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/216-recognize-limit-ldc.phr
@@ -1,0 +1,49 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# Recognizes a stream `.limit(N)` call where the count is loaded via a
+# generic `ldc` carrying a `Φ.jeo.long` literal (the canonical javac
+# encoding for any long argument such as `20L` or `5L`). The pair is
+# abstracted into Φ.hone.limit with count captured as the underlying
+# Φ.number. Mirror rule 722 lowers it back to ldc + invokeinterface
+# when no fold matches; future state-carrying limit optimizations will
+# consume the recognized pragma instead. Sibling of 220 for skip; see
+# issue #548.
+name: recognize-limit-ldc
+pattern: |
+  ⟦
+    𝐵-before,
+    𝜏-push ↦ ⟦
+      φ ↦ Φ.jeo.opcode.ldc,
+      𝜏1 ↦ ⟦
+        φ ↦ Φ.jeo.long,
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Great work on your contribution! 🎉 You've earned **+4 points** this time (+16 base, -4 for 148 hits-of-code ≥ 100, -8 for missing code review). [Keep up the momentum](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) with quality contributions! Your total score stands at **[+461](https://www.eolang.org/objectionary-vitals.html)** - don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #561, opened 2026-05-18
Optimize Stream.distinct() via mapMulti with HashSet

### LINKED ISSUE #547: Optimize .distinct() — deduplicating filter
## The operation

`Stream.distinct()` (Java 8+). Emits the elements of the stream in encounter order, suppressing every value already seen (per `Object.equals`). Signature shape:

```java
Stream<T> distinct();
```

Primitive specializations (`IntStream.distinct`, `LongStream.distinct`, `DoubleStream.distinct`) have the same shape with primitive equality.

## Current behavior

No recognition rule. The `Φ.hone.lambda` with `stream.method = \"distinct\"` propagates untouched through the pipeline. Adjacent transforms can't fuse across it.

## Why it matters

`.distinct()` is one of the canonical stream operations — present in essentially every introductory tutorial — and shows up routinely in deduplication pipelines (`stream.map(...).distinct().sorted()`, `stream.flatMap(...).distinct().count()`). It is also a peer of `.filter` semantically: takes no extra lambda, accepts/rejects each element based on internal state. A chain like `.map(...).distinct().map(...)` would collapse into a single mapMulti if a recognition+distill story existed, but today it is three stages.

## Sketch

Same shape as `.takeWhile` / `.dropWhile` / `.skip` (#539, #540, #546). Recognize `.distinct()` into `Φ.hone.distinct` at the `2xx` stage, then fold into a state-carrying distill variant at `3xx`. The state here is unbounded — a `HashSet` seen-set — which is heavier than the bit/counter state of the other stateful ops; the distill variant must carry an object reference, not a scalar. Once recognized, it fuses with stateless distills on either side without changing observable order.

Related: #539, #540, #546.

yegor256:
Adds a recognition rule (216), a fold rule (353), and a fallback rule (730) for `Stream.distinct()` on object streams. The fold rewrites the call into `Stream.mapMulti(BiConsumer)` capturing a freshly allocated `java/util/HashSet`; the synthetic wrapper forwards an element to the downstream consumer only when `HashSet.add` returns true. The shape mirrors the takeWhile (#539) and dropWhile (#540) pipelines, with an object reference standing in for the boolean[1] guard those use.

Distinct is the canonical deduplication operation and shows up in essentially every introductory stream tutorial. Until now the call was opaque to the optimizer and acted as a barrier for adjacent fusions. The recognition + lowering infrastructure here gives later state-carrying-fusion work a pragma to plug into and, in the meantime, replaces the opaque invokeinterface with an explicit mapMulti that the runtime can JIT freely.

The streams-all-ops fixture exercises the new path together with the rest of the stream operators and its opcode counts move accordingly (+1 invokespecial, +1 invokevirtual, +1 invokedynamic, +1 return). A dedicated streams-distinct.yml pins the dedup semantics ([1,2,2,3,1,3,4].distinct().count() == 4), and phino unit tests cover 216 and 730 directly.

Closes #547

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/216-recognize-distinct.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/216-recognize-distinct.phr
@@ -1,0 +1,39 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# Recognizes a stream `.distinct()` call — the bare `invokeinterface
+# java/util/stream/Stream.distinct:()Ljava/util/stream/Stream;` opcode —
+# and abstracts it into a single Φ.hone.distinct block. Unlike skip,
+# takeWhile and dropWhile, distinct takes no extra argument and no
+# lambda, so the recognition pattern is just one opcode. The downstream
+# rule 353 folds Φ.hone.distinct into Stream.mapMulti with a captured
+# java/util/HashSet seen-set; 730 is the unrecognise fallback for
+# pragmas not consumed by 353.
+name: recognize-distinct
+pattern: |
+  ⟦
+    𝐵-before,
+    𝜏-invokeinterface ↦ ⟦
+      φ ↦ Φ.jeo.opcode.invokeinterface,
+      𝜏1 ↦ "java/util/stream/Stream",
+      𝜏2 ↦ "distinct",
+      𝜏3 ↦ "()Ljava/util/stream/Stream;",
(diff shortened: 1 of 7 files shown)


0crat:
@yegor256 **Great work on your contribution!** 🎉 You've earned **+4 points** (+16 base, -4 for 366 lines of code, -8 for missing code review). Your total score is now [+457](https://www.eolang.org/zerocracy/objectionary-vitals.html)! For future contributions, consider keeping changes under 100 lines and requesting code reviews to maximize your points - [here's why](https://www.yegor256.com/2018/03/06/speed-vs-quality.html). Check your [Zerocracy account](https://www.zerocracy.com/) for updates.

Issue #546, opened 2026-05-18
Optimize .skip(long) — count-based prefix discard


yegor256:
## The operation

`Stream.skip(long n)` (Java 8+). Discards the first `n` elements of the stream, then emits everything that follows. Mirrors `.limit` on the tail side. Signature shape:

```java
Stream<T> skip(long n);
```

The primitive specializations (`IntStream.skip`, `LongStream.skip`, `DoubleStream.skip`) have the same shape.

## Current behavior

No recognition rule. The `Φ.hone.lambda` with `stream.method = \"skip\"` flows through unoptimized as raw `invokedynamic` + `invokeinterface Stream.skip`. Adjacent transforms can't fuse across it.

## Why it matters

`.skip` is the count-based counterpart of `.dropWhile` (#540) and one of the most common ways to slice a stream — pagination (`stream.skip(page*size).limit(size)`), header trimming (`Files.lines(p).skip(1)`), and offset-based windowing. A chain like `.map(...).skip(k).map(...)` would collapse into a single mapMulti if a recognition+distill story existed for `.skip`, but today it is three stages.

## Sketch

Same shape as `.dropWhile` (#540) but with a counter instead of a predicate. Recognize `.skip` into `Φ.hone.skip` at the `2xx` stage, then fold into a state-carrying distill variant at `3xx`. The state is a single `long` counter that decrements until zero; once zero the operation is a pass-through and fuses with any stateless distill that follows. Pairs naturally with a future `.limit` recognizer (`.skip(k).limit(n)` is the canonical pagination idiom).

Related: #539, #540.

0crat:
@yegor256 Thanks for closing this issue! 🎉 According to our bonus policy, you've earned **+12 points**: +8 as the standard basis reward, plus +4 for your quick resolution in under 12 hours. Your running score is [+389](https://www.eolang.org/zerocracy/objectionary-vitals.html) - great work maintaining efficiency and helping the team! Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for the updated balance.

PR #563, opened 2026-05-18
#546: Fold Stream.skip(1L) into mapMulti with long[1] counter

yegor256:
## Summary

Continues #546 from [#560](https://github.com/objectionary/hone-maven-plugin/pull/560) by closing the remaining gap in the state-carrying fold for object Stream: `.skip(1L)`.

`353-skip-to-mapMulti.phr`'s `when:` clause excludes string-typed `count="0"` and `count="1"` — the two forms produced by `210` and `212` when javac picked `lconst_0` / `lconst_1` for the constant. `count="0"` is correctly handled by `311 → nop` (real optimization). `count="1"` was round-tripping through `710` back to `lconst_1 + invokeinterface Stream.skip(J)` and bypassing the state-carrying fold entirely.

`354-skip-one-to-mapMulti.phr` plugs that hole. Same mapMulti shape as `353` — `long[1]` counter array, captured `BiConsumer`, synthetic wrapper method with the decrement-or-forward logic — only the counter-init opcode differs (`lconst_1` instead of `ldc Φ.jeo.long(N)`). Wrapper body is byte-identical to `353`'s.

After `354` fires, `701` lowers the new `Φ.hone.lambda` to `invokedynamic + invokeinterface(mapMulti)` exactly as it does for `353`'s output.

## Test plan

- [x] `bash src/test/phino/run.sh` — 38 phino tests pass (37 existing + 1 new).
- [x] `yamllint` clean on all new files.
- [x] Manually verified that `354` + `710` together fire `354` first when given a `Φ.hone.skip(count="1")` block (skip becomes mapMulti, not raw bytecode).
- [ ] CI runs the Docker-backed end-to-end pack `streams-skipped-one-fused.yml` — `Stream.of(1..5).map(n -> n+1).skip(1L).mapToInt(...).sum()` should print `The result is 18` (3 + 4 + 5 + 6 = 18, proving the first element was discarded as a prefix).

## What's still open from #546

- Primitive-stream folds for `Φ.hone.primitive-skip(N)` — IntStream/LongStream/DoubleStream each need their own mapMulti variant (`mapMultiToInt` / `mapMultiToLong` / `mapMultiToDouble`).
- Cross-stage fusion across the counter-init bytecode (so `.map().skip(N)` actually collapses into one mapMulti via `512`).
- Negative-count safety (silently forwards instead of throwing `IllegalArgumentException`).
- `.skip(K).limit(N)` pagination fusion — `.limit` recognition just landed in master via #548; the cross-op fusion rule is the remaining piece.

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/354-skip-one-to-mapMulti.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/354-skip-one-to-mapMulti.phr
@@ -1,0 +1,193 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# 354-skip-one-to-mapMulti
+#
+# Sibling of 353-skip-to-mapMulti.phr for the `.skip(1L)` shape.
+# Recognises Φ.hone.skip with count="1" (the string-typed form
+# produced by 212 when javac emitted `lconst_1` for the count) and
+# lowers it into the same state-carrying mapMulti BiConsumer the
+# generic ldc case uses, with `lconst_1` in the counter-init instead
+# of `ldc Φ.jeo.long(N)`. Wrapper body is identical to 353's.
+#
+# See 353 for the full design notes (counter cell semantics, known
+# adjacency limitation of the counter-init bytecode w.r.t. 512, and
+# the negative-count gap — none of which apply to count="1" since 1
+# is a positive constant).
+name: skip-one-to-mapMulti
+pattern: |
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 **Great work on your contribution!** 🎉 You've earned **+4 points** (+16 base, -4 for high hits-of-code at 284 >= 100, -8 for missing code review). [Keep the momentum going](https://www.yegor256.com/2018/03/06/speed-vs-quality.html)! Your total score is now [+429](https://www.eolang.org/zerocracy/objectionary-vitals.html) - check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #566, opened 2026-05-18
#546: Fold primitive .skip(N) into mapMulti

yegor256:
Extends the 353/354 state-carrying mapMulti fold to the primitive
streams. Each primitive `.skip(N)` is lowered into the same shape
as the object Stream variant — a `long[1]` counter array captured
by a synthetic primitive `MapMultiConsumer`, with a new
private-static wrapper method per call site that decrements the
counter while it is positive and forwards the value otherwise.

Two rules carry the lowering: 355 for the `ldc Φ.jeo.long(N)`
form, and 356 for the `lconst_1` form that 224 produces when
javac picked the one-byte opcode for `1L`. Both dispatch across
`IntStream` / `LongStream` / `DoubleStream` through a single
`sed` substitution on `stream-class`, producing the right
lambda interface (`IntMapMultiConsumer` / `LongMapMultiConsumer`
/ `DoubleMapMultiConsumer`), wrapper signature, wrapper bytecode
(`iload` / `lload` / `dload`, consumer at `aload 2` for `int` and
`aload 3` for `long` / `double`, `max-locals` 3 or 4) and the
matching `mapMulti` invokeinterface descriptor.

There are phino unit tests covering all three primitive variants
of 355 plus the lconst_1 form of 356, and a Docker integration
test (`streams-skipped-primitive-fused`) that exercises
`.map().skip(2).sum()` on `IntStream` and `.skip(1).sum()` on
`LongStream` / `DoubleStream` end-to-end. The same adjacency and
negative-count caveats noted on 353 carry over here.

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/355-primitive-skip-to-mapMulti.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/355-primitive-skip-to-mapMulti.phr
@@ -1,0 +1,317 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# 355-primitive-skip-to-mapMulti
+#
+# Primitive-stream counterpart of 353. Lowers Φ.hone.primitive-skip
+# (produced by 226 for the ldc form on IntStream / LongStream /
+# DoubleStream) into:
+#   * Raw bytecode that allocates a long[1] counter array on the stack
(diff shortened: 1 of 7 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (910 >= 100); -8 for way too many hits-of-code (910 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+497](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #567, opened 2026-05-18
#546: Collapse adjacent .skip(K).skip(L) into .skip(K+L)

yegor256:
Two new rules collapse adjacent `.skip()` calls into one before the
mapMulti folds turn each into a state-carrying mapMulti. 312 handles
the object `Stream` variant; 314 handles the three primitive streams
(both ops must agree on `stream-class`, so cross-class collapses
never fire). The merged op carries the sum of the two counts, so
`.skip(K).skip(L)` ends up as a single counter array and a single
synthetic wrapper instead of two.

Both rules gate on `Φ.number` counts. The string-typed `count="0"` /
`count="1"` forms produced by 210 / 212 / 222 / 224 get nop'd or
mapMulti-folded on their own pass, and a mixed-typed pair (one ldc,
one lconst) is rare enough that an extra normalisation step is not
worth the rule complexity.

Phino unit tests cover the object Stream case with counts 4 + 3 and
the IntStream case with 5 + 2; the chosen counts produce
unambiguous sums (7) and stream-class-specific output so any
accidental cross-class fusion or off-by-one would surface
immediately.

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/312-collapse-adjacent-skips.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/312-collapse-adjacent-skips.phr
@@ -19,6 +19,17 @@
 # require an extra normalisation step and is not worth the rule
 # complexity given how rarely adjacent `.skip()` calls appear in real
 # code.
+#
+# Java validates each .skip(long) argument independently and throws
+# on a negative literal, while the rest of the skip pipeline (353,
+# the per-count-form folds) silently treats a negative count as
+# zero. This collapse preserves the pre-existing limitation: a pair
+# such as .skip(-4).skip(5) no longer surfaces an exception once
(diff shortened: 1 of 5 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this: **+16** as a basis, **+14.2** for your 284 hits-of-code (284 × 0.05), **-4** for exceeding 100 hits-of-code, **-8** for lack of code review, and **-14.2** to stay within our 24-point maximum. While your code shows solid effort, consider breaking larger contributions into smaller, reviewable chunks to maximize your rewards according to our [policy](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) - code review is essential for earning full points! Your running score is [+501](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

Issue #536, opened 2026-05-18
Recognize method references in 111-invokedynamic-to-lambda


yegor256:
## The gap

`111-invokedynamic-to-lambda.phr` only fires when the invokedynamic target method matches `lambda\$.+`:

```yaml
when:
  and:
    - matches:
        - '\(\).+'
        - 𝑒-interface
    - matches:
        - 'lambda\$.+'
        - 𝑒-target-method
```

This skips **every method reference** form. Streams like:

```java
.mapToInt(Integer::intValue)
.filter(Objects::nonNull)
.peek(System.out::println)
.map(MyClass::transform)
```

…stay as raw `invokedynamic` + `invokeinterface` opcodes in the body and never get promoted to `Φ.hone.lambda`. As a result, none of the 2xx recognition rules (`201-lambda-to-filter`, `202-lambda-to-map`, `205-lambda-to-unbox`, `207-lambda-to-flatMap`, etc.) ever see them, and they propagate untouched through the entire pipeline.

This is currently the largest blind spot in the optimizer — `Integer::intValue` is the canonical idiom for `.mapToInt(...)` and is present in every realistic numeric stream pipeline. The `streams-flatmapped`, `streams-mapped`, and `streams-all-ops` fixtures all hit this.

## Why it's load-bearing

Concrete example from `streams-flatmapped.yml`:

```java
list.stream()
  .map(n -> n * 2)
  .flatMap(n -> Stream.of(n, n + 1))
  .map(n -> n + 10)
  .mapToInt(Integer::intValue)   // ← never optimized
  .sum();
```

After the current pipeline, `main` has:

```
invokeinterface List.stream
invokeinterface Stream.mapMulti     (= the collapsed .map.flatMap.map)
invokeinterface Stream.mapToInt     (← still here, untouched)
invokeinterface IntStream.sum
```

A working 513-style rule (PR #535) for the `.mapToInt(explicit lambda)` form already exists. It just can't fire because the upstream `Φ.hone.unbox` is never produced.

## What it would unlock

- `513-merge-mapMulti-unbox` would fire end-to-end, collapsing `mapMulti.mapToInt` into a single `Stream.mapMultiToInt` in `main`.
- Future rules for `.filter(predicate-ref)`, `.peek(consumer-ref)`, etc. become possible.

## The trade-off

[...]

PR #568, opened 2026-05-19
Pin heterogeneous mapMulti + mapToInt merge for #542

### LINKED ISSUE #542: Heterogeneous merge: Stream.mapMulti + Stream.mapToInt → IntStream.mapMultiToInt
## The operation

Direct continuation of [#536](https://github.com/objectionary/hone-maven-plugin/issues/536) and PR [#535](https://github.com/objectionary/hone-maven-plugin/pull/535). A real-world numeric pipeline ends:

```java
.map(...).flatMap(...).map(...).mapToInt(Integer::intValue).sum()
```

After the existing rules:
- `512-merge-mapMulti` collapses the three `Stream<Integer>` ops into one `Stream.mapMulti(merged)`.
- `Stream.mapToInt(...)` remains as a separate stage because `512` only fuses operations with the same `stream-class`.

## What's missing

A rule that fuses adjacent `Stream.mapMulti` + `Stream.mapToInt` into a single `Stream.mapMultiToInt`. Output type is `IntStream` (not `Stream`), so the two pragmas have different `stream-class` fields and `512` won't touch them.

`513-merge-mapMulti-unbox.phr` (in PR #535) handles the explicit-lambda case (`.mapToInt(n -> n.intValue())`), but not the method-reference idiom (`.mapToInt(Integer::intValue)`) which needs [#536](https://github.com/objectionary/hone-maven-plugin/issues/536) first.

Parallel rules for `.mapToLong → mapMultiToLong` and `.mapToDouble → mapMultiToDouble` follow the same shape.

## Why it matters

This is the **only remaining barrier** preventing a `.map.flatMap.map.mapToInt.sum()` chain from collapsing to a single `IntStream.mapMultiToInt(...).sum()` in `main`. Once present, the optimized bytecode has only two stream-related invokeinterface calls in `main` instead of four.

## Sketch

Pattern: caller body has adjacent `Φ.hone.mapMulti(stream-class=Stream)` followed by `Φ.hone.unbox` (or raw invokedynamic+invokeinterface for method-ref case, depending on resolution of #536). Result: single `Φ.hone.lambda(stream.method=mapMultiToInt, stream.class=Stream)` plus two synthesized methods (merged + wrap-to-primitive). 701 lowers in the usual way.

The wrap method calls the unbox target with explicit invokevirtual — this is the bytecode-growth trade-off discussed in #536.

yegor256:
Adds a phino-level integration test that pins the canonical end-state for issue #542. The test feeds an adjacent `Φ.hone.mapMulti(stream-class=Stream)` followed by the raw `invokedynamic + invokeinterface` pair that `.mapToInt(Integer::intValue)` lowers to, runs the chain 112 → 141 → 205 → 513, and asserts that the body collapses into a single `Φ.hone.lambda(stream.method=mapMultiToInt)` plus the two synthesised `merged` and wrap-to-primitive methods.

The underlying rule chain that makes this work was completed when PR #562 (issue #536) added `112-methodref-to-lambda` — until then `.mapToInt(Integer::intValue)` stayed as raw bytecode that 205 and 513 never saw. The fixtures `streams-flatmapped.yml`, `streams-mapped.yml`, and `streams-all-ops.yml` already exercise this end to end, but each of them spins up Docker and Maven; a regression in any of the four rules currently surfaces only after the slow optimise fixtures run. This test pins the same end-state at the phino layer so it fails fast.

Closes #542

### DIFF
--- a/src/test/phino/pipeline-mapMulti-mapToInt-methodref.yml
+++ b/src/test/phino/pipeline-mapMulti-mapToInt-methodref.yml
@@ -1,0 +1,137 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# Pins the heterogeneous merge from issue #542. After the streams
+# pipeline reaches the 4xx/5xx phases, a .map().flatMap().map() chain
+# folds into a single Φ.hone.mapMulti(stream-class=Stream). The
+# downstream .mapToInt(Integer::intValue) is the method-reference
+# idiom that 111 never sees, so 112-methodref-to-lambda promotes it
+# to Φ.hone.lambda first, 141 supplies the missing opcode, and 205
+# recognises the .mapToInt shape as Φ.hone.unbox. Once the unbox is
+# adjacent to the upstream Φ.hone.mapMulti, 513 collapses both into a
+# single Φ.hone.lambda whose stream.method is mapMultiToInt, plus the
+# two synthesized merged + wrap-to-primitive methods that carry the
+# fused work. This is the canonical end-state that the optimised
+# bytecode of every `.map…mapToInt(Integer::intValue).sum()` chain
+# needs to reach in `main`, and the test asserts on each of the three
+# load-bearing shapes (fused lambda, merged method, wrap method).
+rules:
+  - src/main/resources/org/eolang/hone/rules/streams/112-methodref-to-lambda.phr
+  - src/main/resources/org/eolang/hone/rules/streams/141-set-opcode-in-lambda.phr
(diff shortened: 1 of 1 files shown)


0crat:
@yegor256 Hey there! 🎉 Nice work on your contribution – you've snagged **+4 points** this time! Here's the breakdown: **+16** as your base reward, **-4** because you hit 137 lines (anything over 100 gets dinged), and **-8** since there wasn't a code review. Next time, try to get those reviews in to avoid that penalty – they're worth it! Your running score is sitting pretty at **[+497](https://www.eolang.org/zerocracy/objectionary-vitals.html)**, so [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) those contributions coming and don't forget to peek at your **[Zerocracy account](https://www.zerocracy.com/)** too!

PR #569, opened 2026-05-19
Split opcodes into before+after in YAML test packs

### LINKED ISSUE #564: before + after in the YAML tests of the stream pipeline
We have a `.yml` tests in the `src/test/resources/org/eolang/hone/optimize/` directory. Let's improve their YAML format a bit. Let's rename `opcodes` to `after`.

Let's also introduce `before`, which should also list opcodes, but the counts should be calculated before the optimization.

The new format will make the optimization effect more visible.

yegor256:
Every YAML pack under src/test/resources/org/eolang/hone/optimize/
used to carry a single `opcodes` block that listed how many of each
JVM opcode were expected after the rewrite ran. This change splits
that block in two: `before` (counts in the bytecode that javac just
emitted) and `after` (counts in the bytecode after phino is done).
OptimizeMojoTest reads target/classes-before-hone/ for `before` and
target/classes/ for `after`. A zero value now means the opcode must
be absent at that stage.

The motivation is the request in #564: with the two blocks side by
side the optimization effect is immediately visible in the fixture
(for example, streams-mapped jumps from invokedynamic=5 down to
invokedynamic=2, and streams-take-while shows the boolean[1] guard
appearing as newarray=0 -> newarray=1).

Closes #564

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -178,8 +178,11 @@ The output is the rewritten expression on stdout.
 5. **Add or update an end-to-end YAML** in
    `src/test/resources/org/eolang/hone/optimize/`. These specify the
    Java source to compile, the expected `log` lines, and the expected
-   `opcodes` count in the optimized class. Updating the opcodes counts
-   is normal when a rule changes how a stream pipeline is lowered.
+   opcode counts in the class — `before` for the bytecode produced by
+   `javac` and `after` for the bytecode produced by the optimizer.
+   A count of `0` asserts the opcode is absent at that stage. Updating
+   the `after` counts is normal when a rule changes how a stream
+   pipeline is lowered.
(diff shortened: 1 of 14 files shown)


0crat:
@yegor256 Hey there! Nice work on the contribution! 🎉 You've earned **+4 points** for this one: started with **+16** as the base reward, then **-4** for going over 100 hits-of-code (you hit 171), and **-8** for skipping the code review process. Next time, try to get some reviews in there - our policy rewards contributions that go through proper review, and you'll avoid that 8-point penalty! Your running score is now [+501](https://www.eolang.org/zerocracy/objectionary-vitals.html), so [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) those contributions flowing, and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates!

Issue #570, opened 2026-05-19
not all non-terminals are fused


yegor256:
Look at the `streams-full-non-terminal.yml`. We still have 19 invokedynamic in the end. However, we should have 1.

All non-terminal Stream operations should be fused into a single `mapMulti()`. For some reason it's not happening now.

yegor256:
@rultor release, tag is `0.27.0`

0crat:
@yegor256 Thanks for closing this issue! You've earned **+5 points** for this according to our work policy: **+8** base points, **-10.5** for 21 days of delay (capped at -20), then adjusted to **+5** to meet the minimum **4 points** threshold. Your running score is [**+924**](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too! 🎯

PR #573, opened 2026-05-19
docs(claude): #570 chart which stream operations bypass the distill path

yegor256:
## Summary

Spell out in [CLAUDE.md](CLAUDE.md) which Stream operators currently
fold into `Φ.hone.distill` and which short-circuit straight to
`Φ.hone.mapMulti`, then outline the two payload additions the
distill pragma would need before `401-fuse` could collapse every
sort-bounded segment of [streams-full-non-terminal.yml](src/test/resources/org/eolang/hone/optimize/streams-full-non-terminal.yml) into a single
mapMulti.

Driven by [#570](https://github.com/objectionary/hone-maven-plugin/issues/570). The note records *why* the test
pins `after: invokedynamic: 19` instead of `1` and what a future PR
would have to change to move that number down:

- `filter / map / peek / type / transform / dup` already fold into
  distill via 30x rules.
- `flatMap / mapMulti / distinct / take-while / drop-while / skip`
  bypass distill via 208 / 215 / 351 / 352 / 353 / 354 / 355 / 356
  and each synthesises its own private `BiConsumer` wrapper.
- The fix needs `Φ.hone.distill` to carry **captures** (state types
  for the BiConsumer instance) and an **emit-shape** marker so
  401-fuse can compose CPS-style bodies through continuation
  passing rather than the current stack concatenation.

I also tried a `cancel-identity-unbox-box` rule that erases the
`.mapToInt(Integer::intValue).boxed()` round-trip; locally it took
the pin 21 → 20, but under the test's `target 11` javac the freed
upstream `mapMulti` participated in two extra 512 merges and the
pin moved 19 → 20 instead. Reverted — until distill grows captures,
surface-level fixes can't push the number down.

## Test plan

- [ ] CI runs the existing `OptimizeMojoTest` and the pin stays at 19
- [ ] Reviewers sanity-check the bypass-operator list against the
      streams/ rules
- [ ] Decide whether the captures / emit-shape design is the right
      direction for the follow-up PR

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -108,6 +108,61 @@ If you insert a rule with a prefix that lies between two phases (say,
 a prefix that reflects which invariant your rule preserves on its
 *output* — that determines what later rules see.
 
+### Not every operation lives on the distill path (issue #570)
+
+The 3xx description above is the *intent* — in practice only the
+stateless point-wise operators are folded into `Φ.hone.distill`:
+
+```text
+filter   → 304 / 305 → distill
+map      → 306 → distill
+peek     → 307 → distill
+type     → 303 → distill
+transform→ 302 → distill
+dup      → 301 → distill
+```
+
+Everything else short-circuits straight to `Φ.hone.mapMulti` because
+it either carries state or emits a variable number of downstream
+elements per upstream item:
+
+```text
+flatMap            → 208 (stateless, fan-out via Stream.forEach)
+mapMulti (verbatim)→ 215 (stateless)
+distinct           → 353 (state = HashSet seen-set)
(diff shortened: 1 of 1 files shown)


0crat:
@yegor256 Hey! Nice work on the contribution! 🎉 You've earned **+8 points** for this one: **+16** as a basis, but **-8** for the lack of code review (ouch!). Next time, getting some reviews could really boost your score - our policy rewards quality contributions that go through proper review. Your running score is [+519](https://www.eolang.org/zerocracy/objectionary-vitals.html), so [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too!

PR #586, opened 2026-05-19
Warm up Maven cache before measuring the "before" time

yegor256:
The `hone-it.sh` script that produces the coverage table now runs an
unmeasured `mvn test` against each downstream repository before it
starts the stopwatch. The exit code of that warm-up run is ignored.

Until now the "before" column timed `mvn clean test` on a fresh
checkout, so the very first run had to download the project's whole
dependency tree from Maven Central. That made the "before" number
systematically larger than the "after" number, which then ran on a
warm local repository. After this change both columns measure
compile-plus-test cost on a populated cache and are comparable.

No change to the surrounding pipeline or the CSV schema; the next
coverage run on `master` should already show the narrower gap.

### DIFF
--- a/.github/hone-it.sh
+++ b/.github/hone-it.sh
@@ -71,6 +71,8 @@ budget=300
 flags=(-ntp -B -q --batch-mode -Dlicense.skip -Drat.skip -Dspotbugs.skip -Dcheckstyle.skip -Dpmd.skip -Denforcer.skip)
 
 row="${repo};${sha}"
+echo "warming up Maven dependency cache for ${repo}"
+timeout "${budget}" mvn "${flags[@]}" -f "${dir}" test || true
 start=$(date +%s)
 rc=0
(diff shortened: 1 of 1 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this: **+16** as a basis per our [work policy](https://www.yegor256.com/2014/11/14/hits-of-code.html), **-8** for the lack of code review (required by policy), and **-4** for contributing only 2 hits-of-code (policy requires minimum 40). To maximize your bonus next time, ensure your contributions get reviewed and aim for at least 40 hits-of-code while [keeping quality high](https://www.yegor256.com/2018/03/06/speed-vs-quality.html). Your running score is [+569](https://www.eolang.org/zerocracy/objectionary-vitals.html) - check your [Zerocracy account](https://www.zerocracy.com/) for details!

PR #589, opened 2026-05-19
fix(rules): #570 keep streams-full-non-terminal yaml pack assembling

yegor256:
## Summary

`OptimizeMojoTest.optimizesAsSpecifiedInYamlPack` was red on two packs after 803057e migrated `distinct` from a direct mapMulti-emitting rule to a cps `Φ.hone.distill` flowing through 491 + 502:

- `streams-stateful-ops.yml` — `phino` refused to build the rewritten expression with `Duplicated attribute 'k0-new'` whenever a method contained two `.distinct()` calls. The wrapper-method synthesis in 502 spliced the three statically-named HashSet init opcodes from `353-distinct-to-distill`'s `captures.init` straight into the caller body, so two firings injected the same three names twice into the same formation.
- `streams-full-non-terminal.yml` — `jeo:assemble` blew up with `Unknown opcode name: distill`. `401b-fuse-auto-cps` and `401c-fuse-cps-auto` matched bridge types via metavariables, so `.map(n -> n + 0)`'s Integer-typed auto distill fused into the Object-typed cps distill from `distinct`. 502 hardcodes the wrapper item slot and the LambdaMetafactory bridge as Object, so the mixed-type fused distill survived past 502 and reached `jeo:assemble`.

## Fix

- **502** now captures the three init opcodes by binding name plus value (`𝜏-c0-init-1..3` / `𝑒-c0-init-1..3`) and re-emits them in the caller body with fresh `random-tau` names per firing. The cps distill schema phino sees is unchanged, so no other rule has to move.
- **401b** / **401c** now hardcode `bridge-input` / `bridge-output` / `accepted` / `returned` to `Ljava/lang/Object;` on both sides. Typed auto distills go through 501 on their own; the cps side still reaches 502 alone. Lifting the gate needs 502 to emit a CHECKCAST around the auto body and pick the matching wrapper-item type — tracked under issue #570.
- **streams-full-non-terminal.yml** previously pinned `invokedynamic = 19`, assuming the two `.map(...).distinct()` segments fused into one mapMulti each. With the new gate the two `.distinct()` calls emit their own mapMulti, so the count is `21` today. Pin updated and the comment header explains why.

## Regression test

`src/test/phino/pipeline-double-distinct-to-lambda.yml` chains two adjacent `Φ.hone.distinct` pragmas through 353 + 491 + 502 and asserts the rewritten caller body contains two `Φ.hone.lambda` calls plus the HashSet init opcodes — the exact shape that used to crash phino on `Duplicated attribute 'k0-new'`.

## Test plan

- [x] `bash src/test/phino/run.sh` — 58/58 pass (was 57/57 before the new regression case).
- [x] `mvn -B -Dtest='OptimizeMojoTest#optimizesAsSpecifiedInYamlPack' -Pdeep test` — 9/9 packs pass.

The three unrelated `OptimizeMojoTest` errors I observed (`optimizesSimpleApp`, `generatesStatisticsWithDocker`, `optimizesJustOneLargeJnaClass`) are sandbox-only: the in-Docker `wget` of `apache-maven-${MAVEN_VERSION}-bin.tar.gz` fails TLS verification against the local egress-inspection CA. No code touched by this PR affects them.

https://claude.ai/code/session_01K1Wayj5cpNafhpMUuEaXQq

---
_Generated by [Claude Code](https://claude.ai/code/session_01K1Wayj5cpNafhpMUuEaXQq)_

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -35,6 +35,9 @@ jobs:
           key: ubuntu-jdk-21-maven-${{ hashFiles('**/pom.xml') }}
           restore-keys: ubuntu-jdk-21-maven-
       - run: |
+          sudo apt-get update
+          sudo apt-get install -y cloc
+      - run: |
           sudo wget -q -O /usr/bin/phino http://phino.objectionary.com/releases/ubuntu-24.04/phino-latest
           sudo chmod a+x /usr/bin/phino
       - run: .github/smoke.sh apache/commons-collections 8631ca30ef6222cbafbbca87ca5a5d6123354263
--- a/src/main/resources/org/eolang/hone/rules/streams/401b-fuse-auto-cps.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/401b-fuse-auto-cps.phr
@@ -8,6 +8,16 @@
 # The bodies concatenate verbatim: the auto body leaves the transformed item
 # on top of stack, where the cps body picks it up. The auto side has no
 # captures by convention, so merged captures are exactly the cps side's.
+#
+# Gated to Object-typed distills on both sides. 502-cps-one-cap-to-lambda
+# hardcodes the bridge as (Object, Consumer)V and the wrapper item slot as
+# Object, so a fused distill with a non-Object bridge-input would survive
+# past 502 and leak into jeo:assemble as an unknown Φ.hone.distill opcode.
+# A typed auto distill (e.g. Integer→Integer from .map(n -> n)) is left
+# unfused, so 501 lowers it to its own mapMulti and 502 lowers the cps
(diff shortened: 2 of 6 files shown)


0crat:
@yegor256 Thank you for your contribution! You've earned **+4 points** for this submission: **+16** as a basis, **-4** for exceeding 100 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (158 total), and **-8** for the lack of code review. Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+573](https://www.eolang.org/zerocracy/objectionary-vitals.html) - don't forget to check your [Zerocracy account](https://www.zerocracy.com/) as well.

PR #601, opened 2026-05-20
ci(coverage): stop mvn from eating the repo-loop input via stdin

yegor256:
Run 26145676127 of the `coverage` workflow processed only 9 of the 15 repos listed in `.github/coverage.csv` and then exited cleanly, well under the 30-minute job timeout, with no error in the log. The loop in `.github/coverage.sh` reads the repo list from a process substitution into the shell's stdin, and during each iteration that stdin is inherited by `hone-it.sh` and the many `mvn` invocations it triggers. Maven (or one of its plugins) silently consumed lines from that input, so by the time control returned to the loop the remaining six repos had been read away and `read` saw EOF.

The fix moves the loop's read to file descriptor 3 and feeds `hone-it.sh` from `/dev/null`. No child process can then reach the loop's input, regardless of what mvn or any of its plugins decide to do with stdin. The second loop in the same file only calls `gh api` and rendered all 15 rows correctly in the buggy run, so it is left untouched.

A small bash reproducer with `cat > /dev/null` inside the loop body confirms both the failure and the fix.

### DIFF
--- a/.github/coverage.sh
+++ b/.github/coverage.sh
@@ -16,11 +16,11 @@ echo "hone-maven-plugin installed into local Maven repository"
 
 printf 'repo,sha,build_before,time_before,classes_total,classes_modified,build_after,time_after,loc\n' > "${csv}"
 
-while IFS=',' read -r repo sha; do
+while IFS=',' read -r -u 3 repo sha; do
   test -n "${repo}" || continue
   printf '\n=== %s @ %s ===\n' "${repo}" "${sha}"
-  "${root}/.github/hone-it.sh" "${repo}" "${sha}" "${csv}" || true
-done < <(tail -n +2 "${repos}")
+  "${root}/.github/hone-it.sh" "${repo}" "${sha}" "${csv}" </dev/null || true
+done 3< <(tail -n +2 "${repos}")
 
 echo ""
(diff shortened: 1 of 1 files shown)


0crat:
@yegor256 Hey! Just a heads up - the branch name "`coverage-stdin-leak`" doesn't follow our [naming convention](https://www.yegor256.com/2014/04/15/github-guidelines.html), so that's **-6 points** as per our policy. Next time, just use the ticket number - in this case "`600`" would be perfect! Your running score is [+652](https://www.eolang.org/zerocracy/objectionary-vitals.html), so you're still doing great overall. Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates! 🚀

0crat:
@yegor256 Great work on your contribution! 🎉 You've earned **+4 points** this time (+16 base, -8 for missing code review, -4 for 6 hits-of-code). Your running total is now [+656](https://www.eolang.org/zerocracy/objectionary-vitals.html) - fantastic progress! Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the quality contributions coming and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

Issue #606, opened 2026-05-20
VerifyError: 502b doesn't bump caller max-stack after splicing capture init


yegor256:
Applying `hone-maven-plugin` to the `ronmamo/reflections` repo
(sha `f5514b125c4f4b58e92beb0979a40ddce48d5be1`) leaves
`org.reflections.util.ConfigurationBuilder.addClassLoaders(ClassLoader[])`
with bytecode that the JVM verifier rejects at class load:

```
java.lang.VerifyError: Operand stack overflow
Exception Details:
  Location:
    org/reflections/util/ConfigurationBuilder.addClassLoaders(
        [Ljava/lang/ClassLoader;)Lorg/reflections/util/ConfigurationBuilder; @29: dup
  Reason:
    Exceeded max stack size.
  Current Frame:
    bci: @29
    stack: { 'org/reflections/util/ConfigurationBuilder',
             'java/util/stream/Stream', uninitialized 26 }
```

The verifier walks two cells of upstream Stream state plus the freshly
allocated (still-uninitialized) `HashSet`, then `dup` tries to push a
fourth cell — but the method's `max_stack` attribute is too low to
accept that fourth cell. The original Java is the standard
`distinct` pipeline:

```java
public ConfigurationBuilder addClassLoaders(ClassLoader... classLoaders) {
  this.classLoaders = this.classLoaders == null ? classLoaders :
    Stream.concat(Arrays.stream(this.classLoaders), Arrays.stream(classLoaders))
          .distinct().toArray(ClassLoader[]::new);
  return this;
}
```

The root cause is in
[`src/main/resources/org/eolang/hone/rules/streams/502b-cps-one-cap-to-lambda.phr`](src/main/resources/org/eolang/hone/rules/streams/502b-cps-one-cap-to-lambda.phr).
The rule splices the capture's init triple — for `distinct` it is
`new HashSet`, `dup`, `invokespecial HashSet.<init>` — into the
caller body just before the synthesised lambda. That triple grows
the operand stack by +2 cells at its peak (two `uninitialized`
references between `new` and `invokespecial`), and the net height
afterwards is +1 (the initialised capture value). But the rule
preserves the caller method's `Φ.jeo.maxs` attribute verbatim
through the `𝐵-caller-head` binding group — the only `max-stack`
the rule rewrites is the one on the *new* private wrapper it
synthesises (`max-stack ↦ 6`), not the one on the caller. The
[...]

PR #607, opened 2026-05-20
Replace 5 failing projects in the coverage table

yegor256:
Five projects in `.github/coverage.csv` were showing up as failing in the
README's coverage table — three of them never had a chance because they
don't even build with `mvn` (`mockito` and `RxJava` are Gradle-only, and
`redis/jedis` needs a running Redis server before its tests can do
anything), and the other two (`ronmamo/reflections` and `raphw/byte-buddy`)
fail with a JVM `VerifyError` after hone touches their bytecode.

This PR drops those five and adds five Maven projects that pass the same
`hone-it.sh` flow cleanly, before- and after-hone, on my machine:
`apache/commons-lang`, `commons-collections`, `commons-pool`,
`commons-configuration`, and `commons-validator`. The README table is
updated to match (the times are from the local runs and will get rewritten
the next time `coverage.yml` fires).

I filed #606 separately for the underlying `VerifyError` that bit
`reflections` and `byte-buddy` — `502b-cps-one-cap-to-lambda.phr` splices
the capture-init triple (`new HashSet` / `dup` / `invokespecial`) into
the caller body but doesn't bump that method's `max_stack`, so the
verifier rejects the class as soon as the spliced bytecode runs over the
declared maximum. That's a real hone bug and would be worth fixing
before re-adding similar projects to the table.


0crat:
@yegor256 Hey! I noticed your branch name "`coverage-replace-failing-projects`" doesn't follow our [naming convention](https://www.yegor256.com/2014/04/15/github-guidelines.html). According to our policy, this means **-6** points for you. Next time, just use the ticket number as the branch name - in this case, "`606`" would've been perfect! Your running score is [+640](https://www.eolang.org/zerocracy/objectionary-vitals.html), so you're still doing great overall. Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates! 🚀

0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this: +16 as a basis; +1 for 20 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html); -8 for the lack of code review; -4 for contributing less than 40 hits-of-code; -1 for being under the minimum threshold. Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+664](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #609, opened 2026-05-21
Route every non-terminal stream operator through distill (#570)

yegor256:
Folds every non-terminal Stream operator onto the `Φ.hone.distill` path and lets the 4xx fuse pass collapse adjacent distills before 5xx emits a single `Φ.hone.mapMulti` per remaining one. `take-while`, `drop-while`, the `skip` family (including primitive variants), `flatMap`, raw `mapMulti`, and the `flatMapTo*` / `mapMultiTo*` primitive bridges now produce cps distills with explicit `Φ.hone.emit` markers; `491`/`502` were generalized over capture count so a stateful operator can carry its own state through the wrapper. The fuse pass gained `401b-fuse-auto-cps`, `401c-fuse-cps-auto` (migrated to phino's new `splice` where-function so multi-emit cps bodies fuse on a single firing), `412`/`413` for distill+transition+distill, and a lifted Object-only gate on 401b/401c so typed auto distills can join adjacent cps neighbours. `502a` was widened to derive its SAM class, lambda signature, item-load opcode, and consumer index from `bridge-input`, mirroring the 502b dispatch table.

The motivation is that the old pipeline only fused stateless operators and left every stateful one as its own `invokeinterface` dispatch, which capped how much of a Stream chain could collapse into a single `mapMulti`. Routing everything through distill — and gating fusion only on the genuinely non-fusable operators (`sorted`, `limit`, the flatMap-family cps whose emit shape is `Stream.forEach`-driven) — gives the optimizer a uniform substrate to reason about and pins the `streams-full-non-terminal` fixture to the residual 21 invokedynamic dispatches that are intrinsically unfusable. The `CLAUDE.md` updates record which operators stay non-fusable and why, so future contributors do not try to chase them through the distill path. Phino was bumped to 0.0.0.69 for the `splice` primitive 401c depends on.

A follow-up plan splits the remaining 21 dispatches into three independent phases (flatMap multi-emit fusion, mapMulti fusion behind a phino feature, and a sorted/limit driver-method) — none of that lands here.


0crat:
@yegor256 Thanks for your contribution! You've earned **+4 points** for this submission: +16 as a basis; -4 for exceeding 100 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (8303 >= 100); -8 for exceeding 400 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (8303 >= 400); -8 for the lack of code review; +8 to ensure minimum reward threshold. Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the contributions coming! Your running score is [+668](https://www.eolang.org/zerocracy/objectionary-vitals.html) - don't forget to check your [Zerocracy account](https://www.zerocracy.com/) as well.

PR #619, opened 2026-05-27
Fold .distinct() into the stream-fusion pipeline

### LINKED ISSUE #616: distinct to distill
Let's find out how to map `.distinct()` Stream API methods to distill.

yegor256:
This routes `.distinct()` through the same distill machinery that already handles `.map()` and `.filter()`. Until now distinct survived as a raw `Stream.distinct()` call and acted as a barrier that stopped the map/filter segments on either side from fusing. Now it folds into a `Φ.hone.distill` that carries its `java/util/HashSet` as a `state` bytecode block, fuses with its neighbours through the existing 401 pass, and lowers to a single `Stream.mapMulti` backed by the captured set. The set is built once, captured into the wrapper as an `Object[]`, and consulted per element.

The interesting wrinkle is keeping fusion as plain concatenation: the distill body is index-agnostic and opens with a `Φ.hone.fetch` marker that only gets resolved into the real array load at emit time, so two adjacent distills still fuse without any opcode renumbering. A distinct that never finds a neighbour to fuse with gains nothing from mapMulti, so 441 reverts it back to a plain `invokeinterface Stream.distinct()` — tagged with a marker that stops 216 from re-recognising it, which avoids a fold/revert loop. The first commit realigns CLAUDE.md (it was documenting the reverted #570 design), the second adds the empty `state` scaffolding as a no-op, and the third is the feature.

v1 handles one state variable per fused wrapper and object streams only; the frame type is read off `bridge-input`. Those limits are documented in CLAUDE.md with notes on how a size-driven fixpoint fill generalises them later. Verified with `mvn -Pdeep test` (the new `streams-distinct` fixture plus 216/307/441 phino packs).

Closes #616


0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this submission: +16 as a basis; -4 for exceeding 100 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (1109 total); -8 for exceeding 400 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html); -8 for the lack of code review; +8 to ensure minimum threshold. Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+745](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #631, opened 2026-05-29
Fold .peek() into a fusable distill pragma

yegor256:
A `.peek()` call used to stay a native lambda all the way through the pipeline and never folded, so it broke fusion between the operators around it. This teaches the rewriter to recognise peek's `Consumer` (new rule 207) and fold it into the same auto-style distill that map and filter use (new rule 309). Peek's body is a `dup` followed by the consumer invocation: it keeps a copy of the element, the consumer swallows the duplicate and returns void, and the original is forwarded unchanged — so the body stays one-in-one-out and 401-fuse concatenates it blindly. A `.map().peek().map()` chain now collapses into a single `mapMulti`.

It matters because peek sitting between two maps previously forced two separate dispatches; now the whole run fuses into one, and peek stops being a fusion barrier on the stateless point-wise path.

Two things for the reviewer. Capturing-lambda and method-reference peeks (`x -> sb.append(x)`, `System.out::println`) stay native by design — the same limitation already noted for map, since 111 only lifts `lambda$…` targets. And because the fused result is a non-SIZED `mapMulti`, a `.peek().count()` will now run the peek side effect where the JDK's SIZED `count()` may elide it; peek's javadoc leaves that unspecified and the existing `streams-fusion.yml` fixture already relies on peek folding under `count()`. Covered by new phino unit packs for 207/309 and an end-to-end `streams-peek.yml` that pins the 3→1 invokedynamic collapse.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -132,8 +132,11 @@ body                 the opcode formation spliced between item-load and emit
 
 The body is one-in-one-out (auto-style): phino feeds it a single item
 and it produces a single item, so `401-fuse` can concatenate any two
-adjacent bodies blindly. Pointwise operators (map, filter, …) carry an
-empty `state`; the two stateful operators today are `distinct` (whose
+adjacent bodies blindly. Pointwise operators (map, filter, peek, …) carry
+an empty `state`; `peek` is the one whose body opens with a `dup` — it runs
+its Consumer for the side effect and forwards the element unchanged, so the
+body keeps a copy before the consumer invocation swallows the duplicate and
+returns void (see 309). The two stateful operators today are `distinct` (whose
 `state` builds a `java/util/HashSet` and whose body consults it) and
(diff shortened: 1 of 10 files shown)


0crat:
@yegor256 Thanks for your contribution! You've earned **+4 points** according to our [work policy](https://www.yegor256.com/2014/11/14/hits-of-code.html): **+16** as a basis, **-4** for exceeding 100 hits-of-code (355 total), and **-8** for missing code review. To maximize your bonus next time, consider breaking larger contributions into smaller chunks and ensuring proper code review 🚀 Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+888](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #634, opened 2026-05-30
Fuse stream method references by desugaring them to lambdas

yegor256:
This teaches the optimizer to fuse a stream operator whose argument is a method reference, not just a lambda. Forms like `Stream.of(...).map(n -> n + 1).mapToInt(Integer::intValue).sum()` or `.peek(Integer::intValue)` used to stay as native calls; only the explicit lambda (`n -> n.intValue()`) fused, even though the method-reference form is what most people actually write.

The reason is that rule 111 only lifts an `invokedynamic` whose target is a synthetic `lambda$…` method, and a method reference points its target handle straight at the referenced method. So instead of touching 111, a set of rules running before it desugar the reference into the lambda the compiler could have emitted: they synthesise a private-static `lambda$hone$N` wrapper whose body is the forwarding the `LambdaMetafactory` would have generated, append it to the class, and repoint the `invokedynamic` at it. After that 111 lifts it like any lambda and the existing recognise/fold/fuse pipeline collapses the whole chain into a single `mapMulti`. Trying to admit method refs in 111 directly was the obvious shortcut, but it breaks the `boxed()`-roundtrip and stateful paths at run time — the wrapper's invokestatic-to-a-real-method shape is what keeps the downstream rules (which assume `lambda$…` are static methods) correct.

Three flavours are covered: 103 for unbound-instance refs (`Integer::intValue`), 104 for static refs (`Integer::valueOf`, `X::keep`), and 105 for an unbound-instance ref feeding `.peek(...)`, where the Consumer SAM returns void so the wrapper drops the reference's return value (`invoke …; pop; return`). Conversion never pessimises code: an unbox ref that does not fuse is reverted to its native call by 711/712, and the peek rule keeps a predecessor gate (it fires only after a pointwise `map`/`filter`/`peek`) because a folded peek that never fuses cannot be reverted. Still left native, as follow-ups the same wrapper mechanism can grow into: boxing/widening adaptation (`String::length`), multi-argument refs (`Integer::sum`), constructor refs, interface targets, and captured-receiver refs (`self::decorate`).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -95,7 +95,8 @@ pipeline stages:
 
 ```text
 1xx  prep         strip line-numbers, desugar method refs to lambdas
-                  (103 unbound, 104 static), lower invokedynamic to Φ.hone.lambda
+                  (103 unbound, 104 static, 105 peek/return-drop), lower
+                  invokedynamic to Φ.hone.lambda
 2xx  recognise    lambda + invokeinterface → Φ.hone.{filter,map,unbox,box}
 3xx  fold         every pragma → uniform Φ.hone.distill
 4xx  fuse         adjacent distills → one combined distill  ← the actual win
@@ -210,6 +211,8 @@ that gap before `111`:
 
 - `103-methodref-to-lambda` — **unbound-instance** refs (handle 5).
 - `104-static-methodref-to-lambda` — **static** refs (handle 6).
+- `105-methodref-peek-to-lambda` — **unbound-instance** refs feeding a
+  `.peek(...)` (handle 5, **return-dropping**).
 
 Each synthesises a `private static lambda$hone$N` wrapper whose body is the
 forwarding the `LambdaMetafactory` would have generated (`<load arg>;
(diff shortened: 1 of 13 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this based on our [work policy](https://www.yegor256.com/2014/11/14/hits-of-code.html): starting with **+16** as a basis, **-4** for exceeding 100 hits-of-code (1400 >= 100), **-8** for exceeding 400 hits-of-code (1400 >= 400), **-8** for missing code review, then adjusted to **+4** minimum per policy. While this contribution was substantial with 1400 hits-of-code, consider breaking larger changes into smaller, reviewable pieces to maximize your points and maintain code quality. Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+881](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #638, opened 2026-06-02
Fuse capturing-closure stream lambdas

### LINKED ISSUE #636: we don't fuse with closures
This pipeline won't fuse:

```
final int x = 42;
Stream.of(1, 2, 3).map(i -> i + x).sum();
```

yegor256:
Stream pipelines that used a capturing lambda — `map(i -> i + x)`, `filter(n -> n > t)`, or `map(i -> i + x + y + z)` over effectively-final locals — never fused. javac compiles each captured value as an extra argument on the lambda factory, so the `invokedynamic` descriptor is `(I)L…Function;` (or `(III)…` for three captures) rather than the zero-capture `()L…Function;`, and rule 111's zero-capture guard skipped it. The operators were invisible to the whole 2xx→7xx pipeline.

The fix treats a capture as a pointwise operator that owns one slot in the shared state List that `distinct`/`skip` already thread per-element state through, so capturing operators reuse the existing fuse and emit machinery (401/502/512/521/503/504/601) untouched. They ride new pragma names (`c-map`, `cp-filter`) so no existing recogniser, fold, or revert is affected. New rules 112/113 lift the capturing indy, 314/316 fold it into a stateful distill, and 443/444 revert a lone unfused operator back to its native call so a single capturing op is never pessimised.

The second commit extends the map side to any number of int captures: 112's guard is relaxed to `(I+)`, a new rule 114 peels the run of pushes into the state List one at a time (right-to-left, so they land in source order), and 316 parks the incoming item in its own dead local slot and reloads it after the captures, so one and many captures share a single body shape.

Scope is object streams with int captures — map handles any number, filter one. Reference captures, type-transforming maps, category-2 captures, a multi-capture filter, and `this`-field captures stay native and are noted as follow-ups in the rule headers and CLAUDE.md. Verified with `mvn -Pdeep test` (the full OptimizeMojoTest pack-and-fixture suite is green) and by running the optimised bytecode under `-Xverify:all`.

Closes #636

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -96,7 +96,8 @@ pipeline stages:
 ```text
 1xx  prep         strip line-numbers, desugar method refs to lambdas
                   (103 unbound, 104 static, 105 peek/return-drop), lower
-                  invokedynamic to Φ.hone.lambda
+                  invokedynamic to Φ.hone.lambda; lift a single-int-capture
+                  map/filter indy straight to Φ.hone.{c-map,cp-filter} (112, 113)
 2xx  recognise    lambda + invokeinterface → Φ.hone.{filter,map,unbox,box}
 3xx  fold         every pragma → uniform Φ.hone.distill
 4xx  fuse         adjacent distills → one combined distill  ← the actual win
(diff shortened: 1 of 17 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this work. Here's the breakdown according to our [policy](https://www.yegor256.com/2014/11/14/hits-of-code.html): **+16** base points, **-4** for exceeding 100 hits-of-code (1446), **-8** for exceeding 400 hits-of-code, **-8** for missing code review, then adjusted to our minimum threshold of **+4 points**. While large contributions show dedication, consider breaking them into smaller, reviewable chunks for better quality and higher scores - this approach aligns with our [speed vs quality principles](https://www.yegor256.com/2018/03/06/speed-vs-quality.html). Your running score is [+893](https://www.eolang.org/zerocracy/objectionary-vitals.html); check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #650, opened 2026-06-03
#649 narrow Object item before boxed()/mapToX tail of a distinct/skip pipeline

### LINKED ISSUE #649: VerifyError when a distinct()/skip()-led pipeline has a boxed()/mapToX tail (uncast Object item)
A stream pipeline that begins with `distinct()` or `skip()` before any typed `map`/`filter`, then carries a `boxed()` or trailing `mapToX` tail, optimizes to a class that fails JVM verification. `distinct`/`skip` erase the element type, so the fused stateful distill carries `bridge-input ↦ Ljava/lang/Object;`. The synthesized `distill_N(List, Object, XConsumer)` wrapper loads the item with a fixed `aload 1` (`src/main/resources/org/eolang/hone/rules/streams/5xx/512-distill-state-lambda-to-method.phr:51`) and frames it as `bridge-input` (`src/main/resources/org/eolang/hone/rules/streams/5xx/503-distill-state-frame.phr:73-76`). The boxing body's first opcode then invokes the desugared method-ref wrapper — for `Integer::doubleValue` that target is `(Ljava/lang/Integer;)D` — but the item on the stack is still `Object` and is never narrowed.

Because of that missing cast the verifier rejects the wrapper with `Bad type on operand stack: Type 'java/lang/Object' is not assignable to 'java/lang/Integer'`. A pipeline like `Stream.of(5,3,8,5,2,7).distinct().mapToDouble(Integer::doubleValue).boxed().mapToDouble(Double::doubleValue).sum()` runs before optimization but throws `VerifyError` after it. It stays latent because every test fixture leads with a typed `map`/`filter`, which fixes `bridge-input` to the wrapper; it is independent of the stateful-tail fusion work in #570 / #648, and CLAUDE.md already flags it as future work under the stateful "Limitations" and "boxed() round-trips" sections.

The smallest fix is to insert a `checkcast` to the expected wrapper right after the item load in `512` (and list that type in the `503` keep-frame) whenever `bridge-input` is `Ljava/lang/Object;` but the body's first target invoke expects a narrower reference type.

yegor256:
Fixes #649.

## Problem

A stream pipeline that *begins* with `distinct()`/`skip()` (before any typed `map`/`filter`) and then carries a `boxed()` round-trip or a trailing `mapToX` optimized to a class that failed JVM verification:

```
Bad type on operand stack: Type 'java/lang/Object' is not assignable to 'java/lang/Integer'
```

`distinct`/`skip` erase the element type, so the fused stateful distill carries `bridge-input ↦ Ljava/lang/Object;`. The synthesized `distill_N(List, Object, XConsumer)` wrapper loads the item with a fixed `aload 1` and frames it as Object, but the boxing body's first opcode invokes the desugared method-ref `lambda$` (e.g. for `Integer::doubleValue`, an `(Ljava/lang/Integer;)D` handle) — and the item is never narrowed.

Repro that threw `VerifyError` after optimization:

```java
Stream.of(5, 3, 8, 5, 2, 7)
  .distinct()
  .mapToDouble(Integer::doubleValue)
  .boxed()
  .mapToDouble(Double::doubleValue)
  .sum();
```

## Fix

New rule `505-distill-state-item-cast.phr`. The stateful guard (307/308) keeps the item at a keep-label whose stackmap frame `503` resolves into a `Φ.jeo.frame`; the very next opcode is that first typed `invokestatic`. `505` anchors on the `Φ.jeo.frame → invokestatic` adjacency and splices a `checkcast <wrapper>` between them, narrowing the item exactly where the boxing tail first needs it. The cast runs **after** the frame, which still rightly declares Object, so `503` is unchanged.

It fires only when:
- `bridge-input` is `Ljava/lang/Object;` (the distinct/skip element-erasure signature), and
- the following `invokestatic` takes a **single non-Object reference** argument (so a symmetric distinct/skip and an Object-accepting body are left alone, and a captured `lambda$` — which takes the List plus the item — is skipped).

It runs after `503` (frame-item → frame) and before `512` (wraps the distill-lambda into a method), applied at fixpoint — a checkcast it already inserted now sits between frame and invoke, so the adjacency does not re-match.

### Limitation (documented, follow-up)

Only the first typed op directly after the (last) guard is reached — exactly the boxed()/mapToX-after-distinct/skip shape #649 reports. A `filter`/`peek` wedged between the guard and the boxing tail opens its body with a `dup`, breaking the adjacency; narrowing through that interleaving remains future work.

## Tests

- `src/test/phino/505-distill-state-item-cast.yml` — single-rule unit pack: feeds a distill-lambda with an Object bridge-input and a guard keep-label frame followed by an `(Ljava/lang/Integer;)D` invoke, asserts a `checkcast java/lang/Integer` lands between the frame and the invoke.
- `src/test/resources/org/eolang/hone/optimize/streams/stateful-object-tail.yml` — end-to-end fixture for the issue's exact repro: builds successfully, prints `The result is 25`, and collapses two method-ref `invokedynamic`s to one `Stream.mapMultiToDouble`.
- `CLAUDE.md` updated: the "boxed() round-trips" and stateful "Limitations" sections now document the fix instead of flagging it as future work.

> [!NOTE]
> The e2e/unit packs are `@Tag("deep")` and require the `phino` binary, which is not available in this environment — they were not run locally. The `before`/`after` `invokedynamic` counts and runtime assertion should be confirmed by CI (`mvn -Pdeep test`).

---
_Generated by [Claude Code](https://claude.ai/code/session_017XqayByhvKMPj7MvSBSNNu)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -264,16 +264,22 @@ the trailing mapToX folds into the stateful distill too and the whole stateful
 pipeline collapses to ONE `Stream.mapMultiToX` (see `stateful-boxed-tail.yml`
 and `full-non-terminal.yml`, now 9 → 1) — closing #570 for the stateful path.
 
-One stateful boxed() shape is still unsupported and is a separate puzzle: a
-pipeline that *begins* with `distinct()`/`skip()` (before any typed
+A pipeline that *begins* with `distinct()`/`skip()` (before any typed
 `map`/`filter`) has `bridge-input ↦ Ljava/lang/Object;`, because distinct/skip
 erase the element type. A boxing tail whose desugared `lambda$` expects a
 specific wrapper (e.g. `Integer::doubleValue`, an `(Ljava/lang/Integer;)D`
-handle) then VerifyErrors — the object item is not assignable to `Integer`
-without a `checkcast`. This is pre-existing (it predates and is independent of
-the trailing-mapToX fusion) and is dodged by every fixture by leading with a
-typed operator; lifting it (insert a `checkcast` when `bridge-input` is
-`Object` and the body's first invoke wants a narrower type) is future work.
+handle) used to VerifyError — the object item is not assignable to `Integer`
+without a `checkcast`. `505-distill-state-item-cast` (issue #649) closes this:
+after `503` resolves the guard's keep-label `frame-item` into a `Φ.jeo.frame`,
+the very next opcode is that first typed `invokestatic`, so `505` splices a
+`checkcast <wrapper>` between the frame and the invoke whenever `bridge-input`
+is `Object` and the invoke takes a single non-`Object` reference argument. The
+cast runs *after* the frame, which still rightly declares `Object`, so `503` is
+unchanged. It reaches the first typed op directly after the (last) guard — the
+boxed()/mapToX-after-distinct/skip shape; a `filter`/`peek` wedged between the
+guard and the boxing tail opens its body with a `dup`, breaking the
+`frame → invokestatic` adjacency, so narrowing through that interleaving is a
+follow-up puzzle (see `stateful-object-tail.yml`).
 
 A method reference (`Integer::intValue`, `String::toUpperCase`, `X::keep`, …)
 is **desugared into a lambda** up front so the rest of the pipeline picks it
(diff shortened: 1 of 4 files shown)


0crat:
@yegor256 **Great contribution! 🎉** You've earned **+4 points** (+16 base, -4 for 223 hits-of-code ≥ 100, -8 for missing code review). [Keep the momentum going](https://www.yegor256.com/2018/03/06/speed-vs-quality.html)! Your total score is now [+911](https://www.eolang.org/zerocracy/objectionary-vitals.html) – check your [Zerocracy account](https://www.zerocracy.com/) for updates.

Issue #663, opened 2026-06-03
112-capturing-invokedynamic-to-map.phr:88-95: Generalise...


0pdd:
The puzzle `661-aba99e45` from #661 has to be resolved: 

https://github.com/objectionary/hone-maven-plugin/blob/b610151a2a08c60febcc742d2cca22a49c3a874f/src/main/resources/org/eolang/hone/rules/streams/1xx/112-capturing-invokedynamic-to-map.phr#L88-L95

The puzzle was created by @yegor256 on 03-Jun-26. 

Estimate: 45 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
The puzzle `661-aba99e45` has disappeared from the source code, that's why I closed this issue.

0pdd:
@0pdd the puzzle [#665](https://github.com/objectionary/hone-maven-plugin/issues/665) is still not solved.

0pdd:
@0pdd the puzzle [#668](https://github.com/objectionary/hone-maven-plugin/issues/668) is still not solved; solved: [#665](https://github.com/objectionary/hone-maven-plugin/issues/665).

0pdd:
@0pdd all 2 puzzles are solved here: [#665](https://github.com/objectionary/hone-maven-plugin/issues/665), [#668](https://github.com/objectionary/hone-maven-plugin/issues/668).

PR #666, opened 2026-06-03
#665 Revert lone three-capture map/filter to native invokedynamic

### LINKED ISSUE #665: 112-capturing-invokedynamic-to-map.phr:99-108: Generalise...
The puzzle `663-bc75076a` from #663 has to be resolved: 

https://github.com/objectionary/hone-maven-plugin/blob/81819f925d74f4cbd887975cd791cbf0960c451c/src/main/resources/org/eolang/hone/rules/streams/1xx/112-capturing-invokedynamic-to-map.phr#L99-L108

The puzzle was created by @yegor256 on 03-Jun-26. 

Estimate: 45 minutes,  role: DEV. 

yegor256:
Resolves #665 — the `@todo #663` puzzle in `112-capturing-invokedynamic-to-map.phr:99-108`.

## What

Generalises the lone-capture revert one step further along the established ladder: one capture (`443`/`444`) → two captures (`446`/`447`) → **three captures (`448`/`449`)**.

A lone capturing `map`/`filter` that never fuses with a neighbour is folded into a stateful distill (`316`/`314`) and would otherwise be emitted as a standalone `mapMulti` — strictly slower than the native `Stream.map`/`Stream.filter` the JVM already has. The new rules revert the three-int-capture case back to its original `invokedynamic` + `Stream.{map,filter}`.

- **`448-unfused-three-capturing-map-to-invokedynamic`** — the three-capture twin of `446`. Matches a CLOSED state/body of EXACTLY three boxing appends and three fetches, replays all three `iload` pushes and rebuilds the `(III)L…Function;` factory descriptor, marking the call `reverted ↦ Φ.true` so `112` cannot re-lift it.
- **`449-unfused-three-capturing-filter-to-invokedynamic`** — the filter twin (three-capture twin of `447`), same shape over `Stream.filter` / `Predicate`.

Because both patterns are closed on exactly three appends/fetches, a *fused* operator (two `park/reload/call` triples) cannot match and is left for `502` — the same never-touch-a-fused-distill guarantee `446`/`447` get, extended by one more capture.

## Tests

- Unit packs `src/test/phino/448-*.yml` and `449-*.yml`.
- End-to-end fixture `streams/closure-three-lone.yml`: lone `map(n -> n + p + q + r).reduce(...)` and `filter(n -> n > a && n < b && n != c).count()`; `invokedynamic` count unchanged **3 → 3** (proving neither is pessimised to a `mapMulti`), runtime result `190 2`.

I validated locally end-to-end (jeo 0.15.1 + phino 0.0.0.73): both lone three-capture operators revert to native (2 `reverted` markers, no `mapMulti`); removing `448`/`449` yields 2 `mapMulti`s instead; and the reverted output reassembles, verifies, and runs to `The result is 190 2`.

## Docs / puzzle

- Removed the resolved `@todo` text and re-filed the remaining deferred items (mixed-reference capture run; the lone **four-or-more**-capture revert) as `@todo #665`.
- Updated the headers of `112`, `443`, `444`, `446`, `447` and `CLAUDE.md` to reflect the three-capture revert.

Note: the lone four-or-more-capture revert remains a follow-up puzzle — the real fix there is to replace the one-rule-per-arity ladder with a single arity-agnostic revert that rebuilds the `(I^N)` descriptor for any N.

---
_Generated by [Claude Code](https://claude.ai/code/session_0147ERXMJTfYjuo9e6CYsV54)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -177,9 +177,9 @@ skip                             → 220 → 308 → state distill;
 capturing map (N prim OR 1 ref)  → 112 → 114 (int) / 116 (long) / 117 (double)
                                     / 115 (ref peeler) → 316 → c-map state
                                     distill; 443 reverts a lone single-int-capture
-                                    map, 446 a lone TWO-int-capture map and 448 a
-                                    lone THREE-int-capture map if
-                                    unfused (4+-capture, category-2 and lone-ref
+                                    map, 446 a lone TWO-int-capture map and
+                                    448 a lone THREE-int-capture map if unfused
+                                    (4+-capture, category-2 and lone-ref
                                     stay mapMulti)
 capturing filter (N prim)        → 113 → 118 (int) / 120 (long) / 122 (double)
                                     → 314 → c-filter state distill; 444
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -177,14 +177,16 @@ skip                             → 220 → 308 → state distill;
 capturing map (N prim OR 1 ref)  → 112 → 114 (int) / 116 (long) / 117 (double)
                                     / 115 (ref peeler) → 316 → c-map state
                                     distill; 443 reverts a lone single-int-capture
-                                    map and 446 a lone TWO-int-capture map if
-                                    unfused (3+-capture, category-2 and lone-ref
+                                    map, 446 a lone TWO-int-capture map and 448 a
+                                    lone THREE-int-capture map if
+                                    unfused (4+-capture, category-2 and lone-ref
                                     stay mapMulti)
 capturing filter (N prim)        → 113 → 118 (int) / 120 (long) / 122 (double)
                                     → 314 → c-filter state distill; 444
-                                    reverts a LONE single-int-capture filter and
-                                    447 a lone TWO-int-capture filter to
-                                    invokedynamic if unfused (3+-capture,
+                                    reverts a LONE single-int-capture filter,
+                                    447 a lone TWO-int-capture filter and 449 a
+                                    lone THREE-int-capture filter to
+                                    invokedynamic if unfused (4+-capture,
                                     category-2 stay mapMulti)
 ```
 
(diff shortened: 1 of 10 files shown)


yegor256:
**The arity-agnostic revert this PR defers is blocked by phino's matcher — so extending the ladder may not be worth it.**

I ran a spike against phino `0.0.0.73` (the pinned version) to check whether the "real fix" the description mentions — a single arity-agnostic revert that rebuilds the `(I^N)` descriptor for any N — is actually reachable. It is not, with phino as it stands, and the reason reframes whether `448`/`449` should land.

The lone-vs-fused discrimination is what forces the per-arity ladder. A *fused* capturing distill (two operators `join`ed by `401`) has a body of `park, fetch…, reload, call, park, fetch…, reload, call`; a *lone* one has `park, fetch…, reload, call`. To revert only the lone one, the pattern must reject the fused body. Tested directly with `phino rewrite`:

- A variadic, arity-agnostic pattern `⟦ park, 𝐵-fetches, reload, call, ρ↦∅ ⟧` **fires on the fused body** — `𝐵-fetches` backtracks and swallows the first operator's `reload, call, park`. A single arity-agnostic seed would therefore corrupt fused distills.
- An incremental "peel the last capture" pattern `⟦ park, 𝐵-earlier, fetch, reload, call, ρ↦∅ ⟧` **also fires on the fused body** — same backtracking, peels the wrong operator's capture.
- Only **enumerating the exact fetch count** (what `443`/`446`/`448` do) pins the `reload` position so a fused body cannot match.

So the one-rule-per-arity ladder isn't a stylistic shortcut — it's the only shape phino's matcher currently allows. The forward peeler `114` is arity-agnostic only because it runs *before* `401-fuse`, against a non-repeating anchor (the previous operator's `invokeinterface`); fusion `join`s the bodies and erases that anchor, so the revert has nothing to peel against. Forward and revert are not symmetric.

The clean alternative — stamp `unfused ↦ Φ.true` on the distill at fold time (`316`/`314`) and key the revert on the marker, ignoring body shape — is also blocked: enumerated formation patterns reject an extra binding (a `⟦φ,a⟧` pattern won't match `⟦φ,a,b⟧`; it needs a `𝐵-rest`). The marker would silently stop `401`/`41x`/`42x`/`43x`/`44x`/`45x`/`50x`/`51x` from matching the distill unless each grows a catch-all group — ~15 rules, the opposite of this PR's "no edits to existing rules" virtue.

**Suggestion:**

- The real unblock is a **phino feature** — a repetition/ellipsis matcher that binds a homogeneous group `(fetch unbox)*` and its count (or a count/negation guard). I'm filing that upstream (link in a follow-up).
- `448`/`449` add ~390 rule lines + ~190 test lines to spare a *lone three-capture* map/filter, which is rare and only marginally slower as a `mapMulti`. I'd hold the ladder at two and leave lone-3+ as a `mapMulti` until phino can express the counted match, rather than climbing one more rung.

If you'd rather land it as a stopgap, that's a reasonable call — but then the puzzle in `112`'s header should point at the phino issue (counted-repetition matcher), not at "the four-or-more-capture revert", since that is just the next rung.

Issue #516, opened 2026-05-15
OptimizeMojoTest.java:43-50: enable 'grep-in' tests. The...


0pdd:
The puzzle `440-94f758fa` from #440 has to be resolved: 

https://github.com/objectionary/hone-maven-plugin/blob/688a9de4e4de8878d697b8bcc72d68a6a6a29f44/src/test/java/org/eolang/hone/OptimizeMojoTest.java#L43-L50

The puzzle was created by @yegor256 on 15-May-26. 

Estimate: 90 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

PR #672, opened 2026-06-04
#671 Make default grep-in ERE-compatible

### LINKED ISSUE #671: Default `grep-in` uses PCRE lookaround that `grep -E` cannot parse, so every class is skipped
The default `DEFAULT_GREP_IN` at `src/main/java/org/eolang/hone/OptimizeMojo.java:85` is `(?<!-)(66-69-6C-74-65-72|6D-61-70)(?!-)`, a PCRE pattern that relies on lookbehind and lookahead. The shell pre-filter that consumes it, `src/main/resources/org/eolang/hone/scaffolding/rewrite.sh:79`, applies it with `grep -qE` (POSIX ERE), which has no lookaround. GNU grep prints `warning: ? at start of expression` and reports no match, so the `if ! grep -qE …` guard takes the "No grep-in match … skipping" branch for every `.xmir`. With the default configuration hone therefore skips every class and optimizes nothing. The latest coverage run confirms it: all 15 repositories show `Edits` = 0, including `apache/commons-lang`, whose `RuntimeEnvironment.readFile` has a textbook `filter`/`map`/`map` pipeline.

The lookaround was added in #644 (fixing #449, so `map`/`filter` bytes do not match inside longer tokens such as `mapped/X` or `filtered`). The pattern is correct under Java's regex engine — which is exactly what the unit tests at `src/test/java/org/eolang/hone/OptimizeMojoTest.java:1289` onward exercise via `Pattern.compile(DEFAULT_GREP_IN)`, so they stay green. But the runtime consumer is `grep -E`, not `java.util.regex`, and the two dialects diverge precisely on lookaround; the runtime silently skips everything while CI passes. The regression landed on 2026-06-03, a few hours before the first coverage run that recorded all-zero edits.

The smallest fix is to make the default ERE-compatible by anchoring on the XMIR tag boundaries instead of lookaround, e.g. `>(66-69-6C-74-65-72|6D-61-70)<` — it matches a standalone `map`/`filter` token and still rejects `mapToInt`, `mapped/X`, and `filtered`, satisfying #449. Alternatively, flip `rewrite.sh:79` from `grep -qE` to `grep -qP` to keep the PCRE pattern. Either way the grep-in unit tests should validate the pattern with the actual consumer (`grep -E`/`-P`) rather than `java.util.regex`, so this dialect mismatch cannot recur (related: #516).

yegor256:
Fixes #671.

## Problem

`DEFAULT_GREP_IN` was `(?<!-)(66-69-6C-74-65-72|6D-61-70)(?!-)`, using PCRE negative lookbehind/lookahead. But `rewrite.sh:79` consumes it with `grep -qE` (POSIX ERE), which cannot parse lookaround:

```
grep: warning: ? at start of expression
```

`grep` then matched nothing and exited non-zero, so `! grep -qE ...` was always true — **every class was copied through unchanged and optimization was skipped entirely**.

## Fix

Anchor the hex-byte alternatives between the XML tag boundaries (`>` … `<`) that delimit a complete PHI bytes literal inside an `.xmir`:

```
>(66-69-6C-74-65-72|6D-61-70)<
```

This stays within POSIX ERE (`grep -E` parses it cleanly) while still matching only a standalone `map`/`filter` token and rejecting `mapToInt`, `mapped/X`, and `filtered` — satisfying #449. `grep -qE` is kept, so no `-P`/PCRE portability dependency is introduced.

## Tests

The four grep-in unit tests previously validated the pattern with `java.util.regex`, which masked the dialect mismatch. They now run the pattern through its **real consumer** (`grep -E`) so this cannot recur (related: #516). Added one more case asserting `mapToInt` is rejected.

```
Tests run: 5, Failures: 0, Errors: 0, Skipped: 0
```

https://claude.ai/code/session_01RsBj3QHkaqkCPwX9cmCMqd

---
_Generated by [Claude Code](https://claude.ai/code/session_01RsBj3QHkaqkCPwX9cmCMqd)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -566,10 +566,10 @@ captures over Integer):
   marginally slower than native, and rarer than a lone one- or two-capture
   operator. Reverting it is **not** a matter of writing the next ladder rung: an
   arity-agnostic lone-vs-fused revert is not expressible with phino 0.0.0.73,
-  because a variadic `𝐵-` group backtracks over every split (`Matcher.hs:86`), so
-  one pattern matches a fused two-operator body as readily as a lone one — only
-  enumerating the EXACT capture count pins the `reload` position that tells them
-  apart. The real unblock is a phino counted-repetition matcher (filed as
+  because a variadic `𝐵-` group backtracks over every split (`Matcher.hs:86`),
+  so one pattern matches a fused two-operator body as readily as a lone one —
+  only enumerating the EXACT capture count pins the `reload` position that tells
+  them apart. The real unblock is a phino counted-repetition matcher (filed as
   `objectionary/phino#747`); until it lands, a lone three-or-more-capture
   operator stays a `mapMulti`.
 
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Hey! Nice work on that contribution – you've earned **+8 points** this time: **+16** as a baseline, but **-8** for skipping code review (which our policy requires to avoid that deduction). Getting a review next time would help you max out those points! Your running score is [+913](https://www.eolang.org/zerocracy/objectionary-vitals.html) – [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too! 🚀

PR #673, opened 2026-06-04
Match maxs by literal m1/m2 to restore capturing-closure fusion

yegor256:
The capturing-closure lift rules `112`/`113` matched the enclosing method's `maxs` block with two wildcard-name/wildcard-value bindings (`𝜏-max-stack ↦ 𝑒-max-stack, 𝜏-max-locals ↦ 𝑒-max-locals`). phino 0.0.0.73 can no longer disambiguate that shape and returns no match — so both rules silently stopped firing and every capturing map/filter stayed a native invokedynamic. This was invisible while the default grep-in was broken; once #671 re-enabled optimization, the `closure-*` deep packs and `closures.yml` all regressed. jeo names those bindings `m1` and `m2`, so this matches them by literal name in both the pattern and the result, and rewrites the one unit-pack assertion that used the same broken idiom.

The same red run surfaced two unrelated test-fixture problems that had never been validated (the deep job had been red for ~16 commits behind the grep-in bug). `optimizesAsSpecifiedInYamlPack` was the only optimize test that did not pin a grep-in, so `stateful-object-tail` — whose pipeline has no plain `map`/`filter` token — was skipped instead of optimized; the test now reads a per-pack `grep-in:` (defaulting to the production default) and that pack opts into `.*`. And `closures.yml` expected 28 invokedynamics where the optimizer correctly produces 29: `loneMulti`/`multiFil` have no neighbour to fuse with, so the lone-capture reverts put them back to native rather than folding their trailing `mapToInt` — the never-pessimise choice. Full `OptimizeMojoTest` under `-Pdeep` now passes 92/92.

Two follow-ups for the reviewer: the phino 0.0.0.73 behaviour change looks like an upstream regression worth a separate issue, and folding the trailing `mapToInt` of a lone capturing operator (to reach 27 instead of 29) would need the lone-capture reverts to run after 451 — left as-is here.

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/1xx/112-capturing-invokedynamic-to-map.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/1xx/112-capturing-invokedynamic-to-map.phr
@@ -44,9 +44,11 @@
 # peaks one slot higher than the int append. jeo never recomputes maxs, so this
 # rule BUMPS the enclosing method's max-stack by 2 (a safe over-allocation) whenever
 # the factory descriptor carries a J or D primitive capture; an int-only or a
-# reference capture leaves max-stack untouched. The two maxs values ride generated
-# binding names (jeo emits e.g. "v295 ↦ 5"), so they are matched positionally with
-# 𝜏-max-stack / 𝜏-max-locals (max-stack first, then max-locals).
+# reference capture leaves max-stack untouched. jeo names the two maxs bindings
+# "m1" (max-stack) and "m2" (max-locals), so they are matched by those literal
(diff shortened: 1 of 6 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned **+8 points** for this: **+16** as a basis; **-8** for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming and consider requesting code reviews for future contributions to maximize your points. Your running score is [+921](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #683, opened 2026-06-08
Fuse adjacent Stream.mapMulti operations

### LINKED ISSUE #678: we don't fuse adjacent `mapMulti` operations, but we can
Currently, we don't fuse adjacent `mapMulti` operations.

However, we can do this. For example:

```
s
  .mapMulti((x, c) -> c.accept(x + 1)) // "a" Consumer
  .mapMulti((x, c) -> c.accept(x + 2)) // "b" Consumer
```

May be fused into:

```
s.mapMulti((x, c) -> a.accept(x, y -> b.accept(y, c)));
```

yegor256:
A user-written `s.mapMulti(a).mapMulti(b)` currently stays as two separate stream stages — we fuse `map`/`filter`/etc. but never `mapMulti`. This adds `streams/4xx/461-fuse-mapMulti.phr`, which folds an adjacent pair into one. `111` already lifts each `mapMulti` invokedynamic into a `Φ.hone.lambda` pragma carrying `stream.method=mapMulti`, so the rule just swaps the pair for a single pragma backed by a synthesised `BiConsumer` that drives the first lambda with a wrapper consumer feeding the second — `(x,c) -> a(x, y -> b(y,c))`. It folds a chain of any length left-to-right and fires only on user lambdas (`lambda$…`) and its own products (`fused_…`), so pipeline-generated mapMultis are left alone.

Unlike `map`/`filter`, `mapMulti` is deliberately not folded into a `distill`: a distill is single-emit while `mapMulti` is one-to-many, so keeping it as a first-class pragma is what preserves the one-to-many semantics. The grep-in default also gains `mapMulti`'s hex bytes, otherwise a class that uses only `mapMulti()` gets filtered out before any rule runs.

I checked this end-to-end (disassemble → optimize → assemble → run) for chains of two, three and four: each collapses to a single `Stream.mapMulti`, the composition order is correct, the one-to-many emit is preserved, and the bytecode verifies and runs. Two things to flag: the `closure-reference` deep fixture fails for me, but it fails identically on a clean `master` (a pre-existing flake in the multi-reference capturing-map fusion, and this rule is provably inert on it), and the two Docker-based tests time out here for want of Docker. Primitive `mapMultiToInt/Long/Double`, capturing lambdas, and inlining away the per-element wrapper are left for follow-ups.

Closes #678

### DIFF
--- a/src/main/java/org/eolang/hone/OptimizeMojo.java
+++ b/src/main/java/org/eolang/hone/OptimizeMojo.java
@@ -87,9 +87,16 @@ public final class OptimizeMojo extends AbstractMojo {
      * {@code grep} reject every class and skip optimization entirely
      * (see #671).</p>
      *
+     * <p>The third alternative <tt>6D-61-70-4D-75-6C-74-69</tt> is the
+     * <tt>"mapMulti"</tt> method name; because <tt>"map"</tt> is anchored it
+     * does not cover <tt>"mapMulti"</tt> on its own, so a class that uses only
+     * {@code mapMulti()} (the input that {@code 461-fuse-mapMulti} fuses, see
+     * issue #678) would otherwise be skipped entirely.</p>
+     *
      * @since 0.19.0
      */
-    static final String DEFAULT_GREP_IN = ">(66-69-6C-74-65-72|6D-61-70)<";
+    static final String DEFAULT_GREP_IN =
+        ">(66-69-6C-74-65-72|6D-61-70|6D-61-70-4D-75-6C-74-69)<";
 
     /**
      * Splitter for comma-separated values (with optional whitespace).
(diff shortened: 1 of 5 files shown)


0crat:
@yegor256 Great work on your contribution! 🎉 You've earned **+4 points** this time (+16 base, -12 for code size optimizations needed, -8 for missing review, +8 bonus). Your running total is now **+887** points! For future contributions, consider breaking larger changes into smaller chunks and ensuring code review coverage - this helps maintain our quality standards while maximizing your points. [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the contributions flowing and check your [Zerocracy account](https://www.zerocracy.com/) for updates. You can track progress on the [leaderboard](https://www.eolang.org/zerocracy/objectionary-vitals.html) too!

PR #684, opened 2026-06-08
Fuse flatMap by converting it to mapMulti

### LINKED ISSUE #679: we don't fuse `flatMap`, but we can
Currently, we don't fuse `flatMap()` operations.

However, we can. Right after #678.

We should convert `flatMap` to `mapMulti`. They will be fused later with other `mapMulti`.

yegor256:
This teaches the stream pipeline to fuse `Stream.flatMap`. A flatMap and a mapMulti are two ways to spell the same one-to-many fan-out, so the new rule `455-flatMap-to-mapMulti` rewrites the `Φ.hone.lambda` that `111` lifts a flatMap into the user-mapMulti shape, backed by a synthesised BiConsumer `flatadapt(x, c)` that runs the original mapper and drains its stream into the consumer — `flatMap(f)` becomes `mapMulti((x, c) -> f(x).forEach(c))`. `461` then folds that into any adjacent mapMultis (its guard now also accepts `flatadapt_` targets), and a lone flatMap lowers to one native `Stream.mapMulti` via `701`.

Why: until now a flatMap sitting next to a mapMulti (or another flatMap) stayed a separate pipeline stage. Converting it to mapMulti lets the existing `461` fusion collapse the whole run into one pass. Scope is reference `Stream.flatMap` only; the primitive `flatMapToInt/Long/Double` variants and capturing flatMaps are left for a follow-up. New tests cover it: a `455` single-rule pack and an end-to-end `flatmap-fusion` fixture (flatMap + two mapMultis fuse into one, with an order- and value-sensitive reduce confirming behaviour is preserved); `all-ops` after-counts are updated since its `.flatMap().mapMulti()` now fuses.

Heads-up for the reviewer: the phino pin is bumped to `0.0.0.75`, which is stricter about commas and rejected two comma-less bindings `461` had — those are fixed here. Separately, the `optimize/streams/fusion C` closure fixtures flake in the deep suite on `0.0.0.75` (a pre-existing class-name collision, unrelated to this change — it reproduces with these rules removed and passes in isolation); I'll file that separately.

Closes #679


0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this work. Here's the breakdown: **+16** as a basis, **-4** for exceeding 100 hits-of-code (297 total), and **-8** for the lack of code review. According to our [policy](https://www.yegor256.com/2018/03/06/speed-vs-quality.html), getting your code reviewed helps maximize your bonus - consider requesting reviews for future contributions to avoid this deduction. Your running score is [+891](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming!

PR #690, opened 2026-06-09
Fail loudly on phino errors; distinct class names for closure fixtures (#685)

### LINKED ISSUE #685: Deep suite flakes on fusion.C closure fixtures: maps left unfused (invokedynamic 3 not 2)
`OptimizeMojoTest#optimizesAsSpecifiedInYamlPack` (the opcode assertion at `src/test/java/org/eolang/hone/OptimizeMojoTest.java:298`) intermittently fails on the closure-fusion fixtures whose class is `fusion.C` — `closure-reference.yml` and its siblings `closure-long`, `closure-long-filter`, `closure-multi-lone`, `closure-reference-filter`, `closure-reference-lone`, `closure-string`. The capturing-map fusion does not apply, so the optimized class keeps `invokedynamic: 3` — the unoptimized javac count — instead of the expected `2`.

It is order/load dependent. In the full deep suite it failed at the same parameter index `[6]` (`fusion/C`) in three consecutive runs, yet the same fixture compiled and optimized on its own passes 26/26 (`invokedynamic: 2`, correct runtime result `60`). It is pre-existing and not caused by the #679 flatMap work in #684: a control build with those flatMap rules removed fails `[6]` with the identical opcode map. It just became reliably reproducible after bumping the phino pin to `0.0.0.75` in #684, which looks like timing/ordering pressure rather than a phino correctness change, since `0.0.0.75` optimizes the fixture correctly in isolation. The unoptimized count is the tell: `rewrite.sh` silently `cp`s a file through unoptimized on a phino timeout or skip (`rewrite.sh:80` for grep-in, `rewrite.sh:145` for timeout), so a swallowed phino error or a missed fusion surfaces only as a wrong opcode count instead of a loud failure.

Smallest useful fixes: make `rewrite.sh` fail loudly (non-zero exit) instead of copying the file through unoptimized, so the real phino error shows up in the test log; and give each closure fixture a distinct class name rather than reusing `fusion.C` across eight fixtures, so the suite can attribute which one regressed. Once the error is visible, de-flake the capturing-map fusion if it proves nondeterministic. This is separate from #684 and should not block it.

yegor256:
Fixes #685.

The deep suite intermittently failed on the `fusion.C` closure-fusion
fixtures with `invokedynamic: 3` (the unoptimized javac count) instead of
the expected `2`. As the issue diagnoses, the root visibility problem is in
`rewrite.sh`: a swallowed phino error or a missed fusion surfaced only as a
wrong opcode count instead of a loud failure, and eight fixtures sharing the
class `fusion.C` made it impossible to attribute which one regressed.

This PR implements the two "smallest useful fixes" from the issue:

### 1. `rewrite.sh` fails loudly instead of copying through unoptimized

The timeout handler used a fragile `sec -eq 0` heuristic: if the failed
rewrite took less than a second it was a "failure" (`exit 1`), otherwise it
was treated as a "timeout" and the source was silently `cp`-ed through
unoptimized. Any phino crash taking longer than a second was therefore
misclassified as a timeout and hidden.

Now the handler captures the actual exit code and distinguishes a genuine
`timeout(1)` kill (exit `124`/`137`, still copied through for production
resilience) from a real rewrite failure (any other non-zero exit), which
aborts with the phino exit code and an explicit message on stderr. The real
phino error now shows up in the test log.

### 2. Distinct class names for the closure fixtures

Renamed the colliding fixtures so each `target/classes/.../<Class>.class`
failure message points at exactly one YAML pack:

| Fixture | Old class | New class |
| --- | --- | --- |
| `closure-long.yml` | `fusion.C` | `fusion.CLong` |
| `closure-long-filter.yml` | `fusion.C` | `fusion.CLongFilter` |
| `closure-multi-lone.yml` | `fusion.C` | `fusion.CMultiLone` |
| `closure-reference.yml` | `fusion.C` | `fusion.CReference` |
| `closure-reference-lone.yml` | `fusion.C` | `fusion.CReferenceLone` |
| `closure-reference-filter.yml` | `fusion.CF` | `fusion.CReferenceFilter` |

(`closure-string.yml`'s `fusion.T` and `closures.yml`'s `closures.X` were
already distinct and left unchanged.)

Per the issue, de-flaking the capturing-map fusion itself (should it prove
nondeterministic once the error is visible) is left as a follow-up. This is
separate from #684 and does not block it.

Verified locally: `shellcheck`, `bashate -i E006,E003`, and `yamllint` all
clean on the changed files.

https://claude.ai/code/session_016x7Ln3TcBuLD9adpwtgDWL

---
_Generated by [Claude Code](https://claude.ai/code/session_016x7Ln3TcBuLD9adpwtgDWL)_

### DIFF
--- a/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
@@ -135,14 +135,17 @@ function rewrite_with_timeout {
   xi=${4}
   xo=${5}
   start=$(date '+%s.%N')
-  if ! timeout "${HONE_TIMEOUT}" "${0}" rewrite "$@"; then
+  code=0
+  timeout "${HONE_TIMEOUT}" "${0}" rewrite "$@" || code=$?
+  if [ "${code}" -ne 0 ]; then
     sec=$(perl -E "say int($(date '+%s.%N') - ${start})")
-    if [ "${sec}" -eq 0 ]; then
-      echo "Failure in ${idx} $(basename "${xi}") ($(du -sh "${xi}" | cut -f1))"
-      exit 1
+    if [ "${code}" -eq 124 ] || [ "${code}" -eq 137 ]; then
+      echo "Timeout in ${idx} $(basename "${xi}") ($(du -sh "${xi}" | cut -f1)) after ${sec} seconds"
+      cp "${xi}" "${xo}"
+    else
+      echo "Failure (exit code ${code}) in ${idx} $(basename "${xi}") ($(du -sh "${xi}" | cut -f1)) after ${sec} seconds; refusing to copy it through unoptimized" >&2
+      exit "${code}"
     fi
-    echo "Timeout in ${idx} $(basename "${xi}") ($(du -sh "${xi}" | cut -f1)) after ${sec} seconds"
-    cp "${xi}" "${xo}"
   fi
 }
 
--- a/src/test/resources/org/eolang/hone/optimize/streams/closure-long-filter.yml
+++ b/src/test/resources/org/eolang/hone/optimize/streams/closure-long-filter.yml
@@ -30,7 +30,7 @@ java: |
   package fusion;
   import java.util.*;
   import java.util.stream.*;
-  public class C {
+  public class CLongFilter {
     static int run(long x) {
       return Stream.of(1, 2, 3, 4, 5)
         .filter(n -> n > x)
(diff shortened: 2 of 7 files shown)


0crat:
@yegor256 Hey! Nice work on the contribution – you've earned **+4 points** this time around! Here's the breakdown: **+16** as the base reward, but **-8** for missing code review (that really hurts the score!) and **-4** for the **27** [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) being under our **40** minimum. Getting some reviews next time could boost your points significantly, and aiming for that sweet spot between **40-100** hits-of-code maximizes your rewards. [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming though – your running score is [+895](https://www.eolang.org/objectionary-vitals.html) and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too! 🚀

PR #693, opened 2026-06-09
Fuse stream pipelines in instance methods, not only static ones

### LINKED ISSUE #689: we only fuse in static methods
Currently, a stream pipeline is not fused if it's inside a non-static method, for example:

```
int sum() {
  return Stream.of(1, 2, 3).map(x -> x + 1).map(x -> x + 2).sum();
}
```

However, this one is fused:

```
static int sum() {
  return Stream.of(1, 2, 3).map(x -> x + 1).map(x -> x + 2).sum();
}
```

yegor256:
The six front-door `1xx` rules that lift a stream pipeline out of its
enclosing method (103, 104, 105, 111, 112, 113) matched the method's
modifiers with a hard-coded `static -> Phi.true`. This swaps that literal
for a metavariable that captures and reproduces the real flag, mirroring
what the `4xx` rules already do, so the pipeline lifts whether the
enclosing method is static or not.

A static and an instance method that hold the same pipeline compile to
byte-identical opcode bodies — only the `static` flag, `max-locals`, and
the lambda names differ. Because the front door insisted on a static
enclosing method, an instance pipeline was never lifted and every
downstream phase had nothing to fuse, so the same code that collapsed to
one `mapMultiToInt` in a static method stayed unfused in an instance one.

Verification was awkward locally: the Docker end-to-end test
`OptimizeMojoTest#optimizesAsSpecifiedInYamlPack` can't run on my machine
(a pre-existing `@Mktmp`-vs-parameterized resolver clash errors out all
fixtures before any body runs), so I checked the new `instance-fusion`
fixture by running the installed plugin on the host-phino path and
diffing the bytecode — before 3/4, after 1/3 invokedynamic/invokeinterface,
output 15. The host-phino pack suite (45 tests, including two new instance
packs) passes here; the Docker fixtures should run in CI as usual.

Closes #689

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/1xx/103-methodref-to-lambda.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/1xx/103-methodref-to-lambda.phr
@@ -53,7 +53,7 @@ pattern: |
       access ↦ 𝑒-method-access,
       modifiers ↦ ⟦
         𝐵-modifiers-before,
-        static ↦ Φ.true,
+        static ↦ 𝑒-static,
         𝐵-modifiers-after
       ⟧,
       name ↦ 𝑒-function,
@@ -130,7 +130,7 @@ result: |
       access ↦ 𝑒-method-access,
       modifiers ↦ ⟦
         𝐵-modifiers-before,
-        static ↦ Φ.true,
+        static ↦ 𝑒-static,
         𝐵-modifiers-after
       ⟧,
       name ↦ 𝑒-function,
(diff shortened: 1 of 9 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this: +16 as a basis, +8.8 for 176 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (capped at +16), -4 for exceeding 100 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html), -8 for the lack of code review, final score capped at minimum 4 points. Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+883](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #697, opened 2026-06-09
Add test coverage for asDoubleStream/asLongStream widening

### LINKED ISSUE #696: No test coverage for asDoubleStream()/asLongStream() widening in the stream-fusion pipeline
The `streams/*` rule pipeline has no fixture exercising the primitive widening conversions `IntStream.asDoubleStream()`, `IntStream.asLongStream()`, and `LongStream.asDoubleStream()`. `src/test/resources/org/eolang/hone/optimize/streams/sources.yml` covers `DoubleStream.of` / `LongStream.of` / `IntStream.range` as pipeline *sources*, but no fixture covers these *conversions*, and no `.phr` under `src/main/resources/org/eolang/hone/rules/streams/` references them.

These conversions change the stream element type (int→double, int→long, long→double), and they cannot collapse into a single fused `mapMulti`: `IntStream`/`LongStream` expose only the self-typed `mapMulti` (e.g. `IntStream.mapMulti` returns `IntStream`), while only the reference `Stream<T>` has `mapMultiToInt/Long/Double`. The pipeline already does the right thing — it fuses each run of same-typed operations on either side of the conversion into one `mapMulti` and leaves the conversion between them. For example `IntStream.of(...).filter(...).map(...).asDoubleStream().map(...).filter(...).sum()` lowers to `IntStream.mapMulti · asDoubleStream · DoubleStream.mapMulti` (invokedynamic 6→4, identical result). But nothing guards that behaviour: a future change to `streams/4xx` could silently break same-side fusion across the boundary, or attempt an invalid cross-boundary fusion and corrupt the bytecode, with no test to catch it.

Add an end-to-end fixture under `src/test/resources/org/eolang/hone/optimize/streams/` (alongside `mapto-primitive.yml`) running a pipeline that mixes `asDoubleStream()` and `asLongStream()` with multi-operation runs on each side, asserting the program output and the `before`/`after` opcode counts so the same-side fusion and the preserved conversion are both locked in. A one-line note in `README.md` that primitive widening conversions are intentionally not cross-fused would keep this from being re-investigated.

yegor256:
This adds an end-to-end optimize fixture for the primitive widening conversions `IntStream.asDoubleStream()` and `asLongStream()`, which had no coverage in the `streams/*` suite. `sources.yml` exercised `DoubleStream.of` / `LongStream.of` as pipeline *sources*, but nothing touched the widening calls themselves.

These conversions change the element type (int→double, int→long), and they can't fold into a single fused `mapMulti`: `IntStream` and `LongStream` only expose the self-typed `mapMulti`, and only the reference `Stream<T>` carries the typed `mapMultiToInt/Long/Double` variants that the `mapToX` and `boxed()` fusions lean on. So the conversion call has to stay put. What the pipeline can still do — and what this fixture pins down — is fuse each run of same-typed operations on either side of the conversion into one `mapMulti`. The fixture mixes both conversions with two-operation runs per side and asserts the program output plus the before/after opcode counts (`invokedynamic` 6→4), so a future `4xx` change can't silently break same-side fusion or attempt an invalid cross-boundary fusion unnoticed.

One heads-up: I couldn't run the deep Docker suite locally — every pack errors out on a competing JUnit/Mktmp `Path` resolver, unrelated to this change — so I validated the fixture by replaying the plugin pipeline by hand (same pinned phino/jeo, same flags) and confirmed the optimized class roundtrips and prints the same result. CI should exercise it normally.

Closes #696

### DIFF
--- a/src/test/resources/org/eolang/hone/optimize/streams/as-stream.yml
+++ b/src/test/resources/org/eolang/hone/optimize/streams/as-stream.yml
@@ -1,0 +1,55 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# The primitive widening conversions IntStream.asDoubleStream(),
+# IntStream.asLongStream() and LongStream.asDoubleStream() change the
+# element type of the stream (int -> double, int -> long, long -> double).
+# Such a conversion can never collapse into a single fused mapMulti: the
+# primitive streams expose only the self-typed mapMulti (IntStream.mapMulti
+# returns IntStream, DoubleStream.mapMulti returns DoubleStream), and only
+# the reference Stream<T> carries the typed mapMultiToInt / mapMultiToLong /
+# mapMultiToDouble variants that the mapToX / boxed() / non-terminal fusions
+# rely on. So a primitive-to-primitive transition has no single-mapMulti
+# expression in the JDK and the conversion call must survive untouched.
+#
+# What the pipeline can do, and what this pack locks in, is fuse every run
+# of same-typed operations on each side of the conversion independently. The
+# first pipeline holds two operations on the int side (filter + map) and two
+# on the double side (map + filter); each run collapses into one mapMulti,
+# while asDoubleStream() stays between them. The second pipeline pairs one
+# int-side operation (filter) with one long-side operation (map) across an
+# asLongStream(). Across both pipelines the optimizer drops invokedynamic
+# 6 -> 4 (the two two-operation runs on the first pipeline each fuse 2 -> 1),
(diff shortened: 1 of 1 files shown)


0crat:
@yegor256 Hey! Nice work on that contribution - you've picked up **+8 points** this time! Here's the breakdown: **+16** as your base reward, but **-8** got deducted since there wasn't a code review. Getting your code reviewed is definitely worth it since it helps avoid those point deductions and keeps the quality high. Your running score is sitting pretty at [+881](https://www.eolang.org/zerocracy/objectionary-vitals.html) - [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the contributions flowing and don't forget to peek at your [Zerocracy account](https://www.zerocracy.com/) when you get a chance! 🚀

PR #701, opened 2026-06-09
fix(streams): #699 rename 422 tail transform to avoid duplicate i-transform

### LINKED ISSUE #699: Rule `422-transform-distill-tail-to-object` appends a duplicate `i-transform`, aborting the rewrite
The stream-fusion rule at `src/main/resources/org/eolang/hone/rules/streams/4xx/422-transform-distill-tail-to-object.phr:30-41` can produce a φ-term that binds `i-transform` twice. When that happens, `phino rewrite` aborts with `Couldn't build expression, Duplicated attribute 'i-transform' found in i-transform, i2, i3, i4, i5, i6, i-transform, ρ`, the rewrite of the whole class fails, and the bytecode is shipped unoptimized.

The `result` block rebuilds the body as `body ↦ ⟦ 𝐵-body, i-transform ↦ ⟦…⟧, ρ ↦ ∅ ⟧` (lines 30-40). The meta-variable `𝐵-body` carries over whatever bindings the matched body already had, so when the matched body already binds `i-transform` — for example after `421-transform-distill-head-to-object` has run on the same distill, or on a second application of `422` itself — the appended `i-transform` is a duplicate. The `when` clause (lines 42-50) only constrains the `returned` and `bridge-output` types; nothing prevents the rule from firing on a body that is already transformed.

Reproduced with hone-maven-plugin 0.27.0, phino 0.0.77, jeo 0.15.2, `smallSteps=true`, on the `Main` class of `objectionary/sabj25`: a pipeline that boxes `long` to `Long` (`(J)Ljava/lang/Long;`) at the tail of a fused chain triggers it; the rewrite reaches `401-fuse` and then dies when `422` fires. Because the class is rewritten as one unit, the failure leaves every method unoptimized, not just the offending one.

The smallest fix is to add a `not: matches` guard to the `when` block so `422` does not fire when `𝐵-body` already binds `i-transform` (or have the result replace that binding instead of appending it), so applying the rule can never yield a duplicate attribute.

yegor256:
Fixes #699.

## Problem

`422-transform-distill-tail-to-object` appended a body binding named `i-transform` — the **same** name `421-transform-distill-head-to-object` uses for its head unbox. A distill that accepts an object, works on the primitive internally, and must return an object (e.g. boxing `long`→`Long`, `(J)Ljava/lang/Long;`, at the tail of a fused chain) legitimately needs **both** rules:

- `421` inserts an `invokevirtual longValue` at the head (unbox the incoming `Long`),
- `422` appends an `invokestatic valueOf` at the tail (box the outgoing `long`).

Because both used the name `i-transform`, firing `421` then `422` on the same body produced a duplicate attribute and `phino rewrite` aborted with:

```
Couldn't build expression, Duplicated attribute 'i-transform' found in i-transform, i2, i3, i4, i5, i6, i-transform, ρ
```

Since the class is rewritten as one unit, the crash left **every** method in the class unoptimized.

## Fix

Name `422`'s tail binding `o-transform` instead of `i-transform`. Body opcodes are positional labels inside a `Φ.jeo.seq.of` (existing rules already use non-numeric names like `i-pop-item`, `i-return`), so the rename is purely a label change: the head transform (`i-transform`, before `𝐵-body`) and the tail transform (`o-transform`, after `𝐵-body`) keep their order and coexist without colliding.

I chose this over the guard suggested in the issue because the repro *needs* the tail box: guarding `422` off when `i-transform` is already present would drop that transform and leave the distill half-transformed (object bridge-output, primitive `returned`). Renaming keeps both optimizations and still guarantees the rewrite can never yield a duplicate attribute. (`422` already cannot fire twice on the same distill — after firing, `returned` becomes the object type and fails the `when` guard.)

## Test

Added `src/test/phino/422-transform-distill-tail-to-object.yml`, a single-rule pack that applies `421` and `422` together to a `Long`→`long`→`Long` distill. It asserts the rewrite succeeds (the regression: previously it aborted) and that both the head `invokevirtual` and tail `invokestatic` survive under distinct names. Run with `mvn -Pdeep test`.

https://claude.ai/code/session_01NFQSU9fVooSEn2LBxjAm4Y

---
_Generated by [Claude Code](https://claude.ai/code/session_01NFQSU9fVooSEn2LBxjAm4Y)_

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/4xx/422-transform-distill-tail-to-object.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/4xx/422-transform-distill-tail-to-object.phr
@@ -29,7 +29,7 @@ result: |
     state ↦ 𝑒-state,
     body ↦ ⟦
       𝐵-body,
-      i-transform ↦ ⟦
+      o-transform ↦ ⟦
         φ ↦ Φ.jeo.opcode.invokestatic,
         i2 ↦ 𝑒-transform-class,
         i3 ↦ "valueOf",
--- a/src/test/phino/422-transform-distill-tail-to-object.yml
+++ b/src/test/phino/422-transform-distill-tail-to-object.yml
@@ -1,0 +1,45 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# A distill that accepts an object (Ljava/lang/Long;), works on the primitive
+# (J) internally, and must hand an object (Ljava/lang/Long;) back needs BOTH
+# transforms: 421 unboxes the incoming Long at the head (invokevirtual
+# longValue) and 422 boxes the outgoing long at the tail (invokestatic
+# valueOf). Before #699 both rules appended a binding called `i-transform`, so
+# firing 421 and then 422 on the same body produced a duplicate `i-transform`
+# attribute and `phino rewrite` aborted the whole class. 422 now names its tail
+# binding `o-transform`, so the two transforms coexist and the rewrite
+# succeeds.
+rules:
+  - src/main/resources/org/eolang/hone/rules/streams/4xx/421-transform-distill-head-to-object.phr
+  - src/main/resources/org/eolang/hone/rules/streams/4xx/422-transform-distill-tail-to-object.phr
+input: |
+  {⟦
+    foo ↦ ⟦
+      φ ↦ Φ.hone.distill,
+      class ↦ "Foo",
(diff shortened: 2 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned **+8 points** for this submission: **+16** as the base reward, but **-8** for the lack of code review as required by our team policy. Please ensure future contributions go through proper code review to maximize your points. [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+913](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

Issue #351, opened 2025-09-30
`mustOptimize` to pass quietly if can't optimize


yegor256:
Let's introduce a new parameter `mustOptimize` (false by default). If it's set to true, the `OptimizeMojo` must fail with an error if Docker is not available and `phino` is not available either.

If `mustOptimize` is set to false, we should quietly proceed (with a log message) if we can't optimize `.class` files, leaving them intact.

We should also pass quietly in the `PullMojo` and `RmiMojo` if this parameter is set to `false`.

yegor256:
@volodya-lombrozo maybe you can help?

volodya-lombrozo:
@yegor256 sure. Could you assign it to me, please?

yegor256:
@volodya-lombrozo any update here?

PR #702, opened 2026-06-09
fix(scaffolding): #700 require output XMIR before skipping rewrite

### LINKED ISSUE #700: Hone reports BUILD SUCCESS and ships an unoptimized class when a phino rewrite fails (`rewrite.sh`)
When phino fails to rewrite a class, a `-Phone` build can still finish BUILD SUCCESS and emit the original, unoptimized `.class`. The intended hard-fail path does exist (`src/main/resources/org/eolang/hone/scaffolding/rewrite.sh:146`: "refusing to copy it through unoptimized", followed by `exit`), but two other paths in the same script let a class that was selected for rewriting pass through as a "success" without ever being rewritten.

First, the up-to-date guard at `rewrite.sh:74-76` returns as soon as the optimized phi `${pho}` is newer than the source phi `${phi}`, which is *before* the `phino rewrite --output=xmir … > "${xo}"` step at `rewrite.sh:122`. It never checks that the downstream unphi XMIR `${xo}` (the file `jeo:assemble` consumes) was actually produced, so when a half-finished `phi-optimized` from an earlier failed run is present, the rewrite is "skipped" and assemble finds nothing. Second, the timeout branch at `rewrite.sh:142-144` copies the un-rewritten input through (`cp "${xi}" "${xo}"`) and returns 0. In both cases the build is green and the bytecode is unchanged.

Reproduced on `objectionary/sabj25` `Main` with hone 0.27.0, phino 0.0.77, jeo 0.15.2: phino dies on the rule-422 duplicate-attribute bug (#699), yet `mvn -Phone … package` reports BUILD SUCCESS, the resulting `Main.class` is byte-for-byte identical to the un-honed input, `target/hone-statistics.csv` contains only its header row, and `jeo:assemble` logs "Total 0 files were assembled" — all at exit code 0 with nothing at WARN/ERROR. A consumer measuring before-vs-after hone sees no difference and no signal that optimization silently did not happen. (This is distinct from #351, which concerns Docker/phino being *unavailable*; here phino runs and errors.)

Smallest fix: make the `rewrite.sh:74` guard also require that `${xo}` exists and is newer than `${pho}` before skipping (so a half-finished prior run is redone rather than accepted), and treat a phino non-zero exit as a hard build failure instead of leaving the original class in place; the quiet pass-through can stay behind the `mustOptimize=false` parameter proposed in #351.

yegor256:
Fixes #700.

## Problem

When a `phino` rewrite fails, a `-Phone` build could still finish `BUILD SUCCESS` and emit the original, unoptimized `.class`.

The up-to-date guard in `rewrite.sh` (line 74) returned as soon as the optimized phi `${pho}` was newer than the source phi `${phi}` — *before* the `phino rewrite --output=xmir … > "${xo}"` step at line 122. It never checked that the downstream output XMIR `${xo}` (the file `jeo:assemble` consumes) was actually produced. So when a half-finished `phi-optimized` from an earlier failed run was present, the rewrite was "skipped", `jeo:assemble` found nothing (`Total 0 files were assembled`), `hone-statistics.csv` contained only its header row, and the build went green with the bytecode unchanged.

## Fix

The guard now additionally requires that `${xo}` exists and is newer than `${pho}` before skipping:

```bash
if [ -f "${pho}" ] && [ "${pho}" -nt "${phi}" ] && [ -f "${xo}" ] && [ "${xo}" -nt "${pho}" ]; then
```

A half-finished prior run (missing or stale output XMIR) is now redone rather than accepted. On the redo, a genuine `phino` non-zero exit propagates through the existing hard-fail path (`rewrite.sh:146-147`, "refusing to copy it through unoptimized"), so the build fails loudly instead of silently shipping an unoptimized class.

## Scope notes

- The timeout pass-through (`rewrite.sh:142-144`) is left untouched: it is documented intended behavior (`OptimizeMojo.timeout`: "the file remains untouched … doesn't lead to runtime failure"), and per the issue the quiet pass-through should stay behind the future `mustOptimize=false` parameter proposed in #351.
- The existing `optimizesTwice` test is unaffected: on a clean prior run, `${xo}` is written after `${pho}`, so the new condition holds and the skip still occurs. The grep-in skip path also always produces `${xo}`, so any *complete* prior run satisfies the new condition.

https://claude.ai/code/session_014PBPVDcHZpK4ULqyJkr2dM

---
_Generated by [Claude Code](https://claude.ai/code/session_014PBPVDcHZpK4ULqyJkr2dM)_

### DIFF
--- a/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
@@ -71,8 +71,8 @@ function rewrite {
   mkdir -p "$(dirname "${phi}")"
   mkdir -p "$(dirname "${pho}")"
   mkdir -p "$(dirname "${xo}")"
-  if [ -f "${pho}" ] && [ "${pho}" -nt "${phi}" ]; then
-    echo "Target $(basename "${pho}") is newer than source $(basename "${phi}"); skipping transformation for ${idx}"
+  if [ -f "${pho}" ] && [ "${pho}" -nt "${phi}" ] && [ -f "${xo}" ] && [ "${xo}" -nt "${pho}" ]; then
+    echo "Target $(basename "${pho}") is newer than source $(basename "${phi}") and output $(basename "${xo}") is present; skipping transformation for ${idx}"
     return
   fi
   verbose "Next ${idx} XMIR is ${xi} ($(du -sh "${xi}" | cut -f1))"


0crat:
@yegor256 Great work on your contribution! 🎉 You've earned **+4 points** (+16 base, -8 for missing code review, -4 for low hits-of-code at 4). Your momentum is building with a running score of [+917](https://www.eolang.org/zerocracy/objectionary-vitals.html)! [Keep the quality contributions flowing](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #707, opened 2026-06-10
fix(streams): #704 keep boxed-primitive filter in the boxed domain

### LINKED ISSUE #704: VerifyError: filter().boxed() fuses to a distill whose keep-frame says 'long' while the stack holds a 'Long'

```
```

yegor256:
Fixes #704.

## Problem

A primitive `filter(...).boxed()` chain whose `boxed()` is consumed by a reference-stream operation (e.g. `collect`, `forEach`, `count`) stays in the boxed domain:

- `203`→`211`→`232` turn the filter into a `Φ.hone.p-filter` whose `bridge-input`/`bridge-output` are the wrapper (`Ljava/lang/Long;`) while `accepted`/`returned` keep the **primitive** predicate shape `J`.
- Because the trailing `boxed()` keeps the boxing alive, `411` never unwinds the filter distill back to a primitive one, so the distill genuinely handles a one-slot `Ljava/lang/Long;` reference on the stack.
- But `304` emitted the keep-frame from `returned` (→ `long`, two slots) and `422` appended a `Long.valueOf(J)` as if re-boxing a primitive. The kept value is already a `Long`, so the optimized class failed JVM verification:

```
```

## Fix

The keep value of a filter is its `bridge-input`, whose final shape is only known **after** `411` decides whether the box/unbox round-trip could be unwound. So:

- **`304`** now defers the keep-frame to a `Φ.hone.frame-item` marker (the non-capturing twin of what `314` already does).
- **`506`** (new, stateless twin of `503`/`504`) resolves that marker from the **final** `bridge-input` — the wrapper when the filter stayed boxed (`java/lang/Long`), the primitive when `411` unwound it (`long`).
- **`412`** (new) realigns a boxed filter distill's `returned` to its wrapper `bridge-output` (a filter preserves its element), so `422` sees a type-preserving distill and adds no redundant `valueOf`. `accepted` stays primitive so `421` still unboxes for the `(J)Z` predicate. It is scoped to filter distills via the `Φ.hone.frame-item` marker, leaving boxed-primitive **map** distills (which do compute a fresh primitive and need `422`) untouched.

The resulting distill body matches exactly what the issue says it should be: a `java/lang/Long` keep-frame and a survivor passed straight to `Consumer.accept(Object)` with no stray `valueOf`.

## Tests

- `src/test/resources/org/eolang/hone/optimize/streams/long-filter-boxed.yml` — end-to-end reproduction (`LongStream.filter(...).boxed().collect(...)`), now verifies and runs (`The result is 18`).
- `src/test/phino/412-distill-filter-returned-to-object.yml` and `src/test/phino/506-distill-filter-frame.yml` — single-rule packs for the two new rules.

## Validation (JDK 21, host phino 0.0.77, no Docker)

- All 48 single-rule phino packs pass (46 existing + 2 new).
- All non-Docker end-to-end `OptimizeMojoTest` cases pass.
- Direct pipeline checks: `LongStream.filter().boxed().collect()`, `...forEach()` now verify; non-boxed `Int/Long/Double` `filter().map().sum()`, object `filter`/`filter().map()` still verify and produce correct results.

## Known limitation

A *chained* boxed-primitive filter (`filter().boxed().filter()`) hits a separate, deeper inter-filter box/unbox issue (rules `221`/`252`/`282`) and is not fixed here — it was already broken before this change. This PR resolves the single-`filter().boxed()` case reported in #704.

https://claude.ai/code/session_01Hg1txmiF5dhvSPKjj3Q41F

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hg1txmiF5dhvSPKjj3Q41F)_

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/3xx/304-primitive-filter-to-distill.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/3xx/304-primitive-filter-to-distill.phr
@@ -54,17 +54,7 @@ result: |
         φ ↦ Φ.jeo.label,
         i1 ↦ 𝑒-label
       ⟧,
-      i6 ↦ ⟦
-        φ ↦ Φ.jeo.frame,
-        type ↦ 4,
-        nlocal ↦ 0,
-        locals ↦ ⟦ φ ↦ Φ.jeo.seq.of0 ⟧,
-        nstack ↦ 1,
-        stack ↦ ⟦
-          φ ↦ Φ.jeo.seq.of1,
-          item ↦ 𝑒-stack-item
-        ⟧
-      ⟧
+      i6 ↦ ⟦ φ ↦ Φ.hone.frame-item ⟧
     ⟧
   ⟧
 where:
@@ -74,19 +64,6 @@ where:
   - meta: 𝜏-opcode
     function: tau
     args: [𝑒-opcode]
-  - meta: 𝑒-stack-item
-    function: sed
-    args:
-      - 𝑒-returned
-      - '"s/^B$/byte/g"'
-      - '"s/^C$/char/g"'
-      - '"s/^D$/double/g"'
-      - '"s/^F$/float/g"'
-      - '"s/^I$/integer/g"'
-      - '"s/^J$/long/g"'
-      - '"s/^S$/short/g"'
-      - '"s/^Z$/boolean/g"'
-      - '"s/^L(.+);$/$1/g"'
   - meta: 𝑒-distill
     function: sed
     args:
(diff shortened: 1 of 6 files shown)


0crat:
@yegor256 Hey there! 🎉 Nice work on your contribution! You've earned **+4 points** for this one: started with +16 as the baseline, then -4 for having 296 hits-of-code (anything over 100 gets a deduction), and -8 for skipping the code review. I know the points took a hit this time, but remember that getting those reviews really helps boost your score - plus keeping contributions under 100 hits-of-code works in your favor too! [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming though! Your running score is [+921](https://www.eolang.org/zerocracy/objectionary-vitals.html), and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) as well.

PR #709, opened 2026-06-10
Count genuine stream optimizations in coverage "Edits" column

### LINKED ISSUE #708: count_modified in hone-it.sh counts stream-agnostic edits, so coverage "Edits" shows NN > MM
The coverage table in `README.md` renders an "Edits" column as `modified/streams` (`.github/coverage.sh:43-44`). The numerator comes from `count_modified` in `.github/hone-it.sh:78-89`, which tallies rows in `hone-statistics.csv` where the `Changed` field is greater than zero. Many rows show the numerator larger than the denominator — `junit4 11/0`, `vavr 169/24`, `commons-collections 101/12` — which is impossible if the numerator means "stream classes optimised."

The reason is that `Changed` is just the line diff between the input and output `.phi` of phino's rewrite pass (`rewrite.sh:118-121`), and that pass is not stream-specific. Rule `streams/1xx/151-remove-line-numbers.phr` strips the `Φ.jeo.line-number` nodes from every debug-compiled class, and the grep-in pre-filter (`OptimizeMojo.java:98-99`) admits any class with a method named `map`/`filter`/`mapMulti` (such as `Optional.map`), not only `java.util.stream` users. So a class with no stream bytecode gets `Changed > 0` and is counted as "modified," while `count_streams` (`.github/hone-it.sh:57-62`) greps the bytecode for `java/util/stream/…Stream` and correctly reports zero. junit4 (pre-Java-8, zero streams) showing `11/0` is the proof.

The fix is to make `count_modified` count genuine optimisations instead of `.phi` churn: compare the stream-op count in `target/classes-before-hone` (the original javac bytecode saved at `entry.sh:71`) against `target/classes` (optimised), using the same regex as `count_streams`, and count only classes where it dropped. The jeo round-trip preserves invokes, so this is immune to round-trip churn, and it guarantees the numerator is a subset of the stream-class population (`NN ≤ MM`). Suggested replacement:

```bash
count_optimized() {
  local base=$1
  local total=0 after before a b
  while IFS= read -r -d '' after; do
    before=${after/\/target\/classes\//\/target\/classes-before-hone\/}
    test -f "${before}" || continue
    a=$(javap -c -p "${after}"  2>/dev/null | grep -cE 'java/util/stream/(Int|Long|Double)?Stream' || true)
    b=$(javap -c -p "${before}" 2>/dev/null | grep -cE 'java/util/stream/(Int|Long|Double)?Stream' || true)
    test "${a}" -lt "${b}" && total=$(( total + 1 ))
  done < <(find "${base}" -type f -path '*/target/classes/*.class' -print0)
  echo "${total}"
}
```

yegor256:
Fixes #708.

## Problem

The coverage table's "Edits" column (`modified/streams`) was rendering numerators larger than denominators — `junit4 11/0`, `vavr 169/24`, `commons-collections 101/12` — which is impossible if the numerator means "stream classes optimised".

`count_modified` tallied rows in `hone-statistics.csv` where `Changed > 0`. But `Changed` is just the line diff between the input and output `.phi` of phino's rewrite pass, which is **not** stream-specific:

- `streams/1xx/151-remove-line-numbers.phr` strips `Φ.jeo.line-number` nodes from every debug-compiled class.
- The grep-in pre-filter (`OptimizeMojo.java`) admits any class with a `map`/`filter`/`mapMulti` method (such as `Optional.map`), not only `java.util.stream` users.

So classes with no stream bytecode get `Changed > 0` and are counted, while `count_streams` correctly reports zero (junit4, pre-Java-8, showing `11/0` is the proof).

## Fix

Replace `count_modified` with `count_optimized`, which compares the `java.util.stream` invoke count in `target/classes-before-hone` (the original javac bytecode) against `target/classes` (optimised), using the same regex as `count_streams`, and counts only classes where it dropped.

The jeo round-trip preserves invokes, so this is immune to round-trip churn, and it guarantees the numerator is a subset of the stream-class population (`NN ≤ MM`).

---
_Generated by [Claude Code](https://claude.ai/code/session_017TL48XTFuMUT5QWLWdjRgU)_

### DIFF
--- a/.github/hone-it.sh
+++ b/.github/hone-it.sh
@@ -75,17 +75,17 @@ apply_hone() {
   done < <(find "${base}" -type d -path '*/target/classes' -print0)
 }
 
-count_modified() {
+count_optimized() {
   local base=$1
-  local -a stats=()
-  while IFS= read -r -d '' f; do
-    stats+=("${f}")
-  done < <(find "${base}" -type f -path '*/target/hone-statistics.csv' -print0)
-  if [ "${#stats[@]}" -eq 0 ]; then
-    echo 0
-    return
-  fi
-  awk -F',' 'FNR == 1 { next } $(NF-1) + 0 > 0 { modified += 1 } END { print modified + 0 }' "${stats[@]}"
+  local total=0 after before a b
+  while IFS= read -r -d '' after; do
+    before=${after/\/target\/classes\//\/target\/classes-before-hone\/}
+    test -f "${before}" || continue
+    a=$(javap -c -p "${after}"  2>/dev/null | grep -cE 'java/util/stream/(Int|Long|Double)?Stream' || true)
+    b=$(javap -c -p "${before}" 2>/dev/null | grep -cE 'java/util/stream/(Int|Long|Double)?Stream' || true)
+    test "${a}" -lt "${b}" && total=$(( total + 1 ))
+  done < <(find "${base}" -type f -path '*/target/classes/*.class' -print0)
+  echo "${total}"
 }
 
 build_outcome() {
@@ -133,7 +133,7 @@ total=$(count_classes "${dir}")
 hstart=$(date +%s)
 apply_hone "${dir}"
 hone_seconds=$(( $(date +%s) - hstart ))
-count=$(count_modified "${dir}")
+count=$(count_optimized "${dir}")
 row="${row},${total},${count},${hone_seconds}"
 echo "warming up the test runner for honed ${repo}"
 timeout "${budget}" mvn "${flags[@]}" -f "${dir}" initialize surefire:test || true


0crat:
@yegor256 Thanks for your contribution! You've earned **+4 points** for this submission: **+16** as the base reward, **-8** for the lack of code review, and **-4** for contributing fewer than 40 hits-of-code (22 total). Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming and consider getting your code reviewed next time. Your running score is [+929](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #712, opened 2026-06-10
Convert flatMapToInt/Long/Double to mapMultiToX (#711)

### LINKED ISSUE #711: flatMapToInt/ToLong/ToDouble are not converted to mapMultiToX, though 455 does it for flatMap
Rule \`src/main/resources/org/eolang/hone/rules/streams/4xx/455-flatMap-to-mapMulti.phr\` converts a user-written \`Stream.flatMap\` into \`Stream.mapMulti\`, but its pattern (lines 58-62) is gated to \`stream.method = "flatMap"\` returning the reference \`java/util/stream/Stream\`, and its header (lines 27-31) explicitly leaves \`flatMapToInt\`, \`flatMapToLong\`, and \`flatMapToDouble\` "for a follow-up". Today those three are lifted by 111 into the \`Φ.hone.lambda\` pragma, match nothing, and are reverted untouched by 701. They are also absent from \`DEFAULT_GREP_IN\` in \`src/main/java/org/eolang/hone/OptimizeMojo.java:97-109\` — and since each alternative is anchored as the entire element content (see #449, #705), \`flatMap\` does not cover \`flatMapToInt\`, so a class using only the typed variants is skipped before any rule runs.

This is wrong to leave on the table because the JDK provides exact counterparts on the reference stream — \`Stream.mapMultiToInt(BiConsumer<T, IntConsumer>)\`, \`mapMultiToLong\`, \`mapMultiToDouble\` — and the very adapter shape 455 already synthesises works unchanged: \`(x, c) -> mapper(x).forEach(c)\`, since \`IntStream.forEach\` takes precisely the \`IntConsumer\` that \`mapMultiToInt\` hands down (same for long/double). The conversion removes the per-element substream spliterator plumbing exactly as it does for reference \`flatMap\`. Unlike the \`asDoubleStream\` case in #696, no JDK method is missing here, and 501/502 already emit \`mapMultiToInt/ToLong/ToDouble\` dispatches, so the jeo roundtrip of these calls is proven.

Suggested fix: add three sibling rules next to 455 (say 456/457/458) that differ only in \`stream.method\` and \`stream.signature\` (result \`IntStream\`/\`LongStream\`/\`DoubleStream\`), the adapter's second parameter type (\`Ljava/util/function/IntConsumer;\` etc.), and the \`forEach\` invokeinterface owner and descriptor; add the three method names to \`DEFAULT_GREP_IN\`. The \`after\` counts in \`src/test/resources/org/eolang/hone/optimize/streams/all-ops.yml\` will shift accordingly; a dedicated e2e fixture and \`src/test/phino/\` packs should pin the new rules. The same caveats 455 already accepts apply (null-skip and auto-close of the substream are not reproduced). The primitive-side \`IntStream.flatMap\` (an \`IntFunction\` SAM lowering to \`IntStream.mapMulti\`) is a separate possible follow-up, out of scope here.

yegor256:
Fixes #711.

`455-flatMap-to-mapMulti` converts a reference `Stream.flatMap` into a `Stream.mapMulti`, but explicitly left the primitive variants `flatMapToInt`/`flatMapToLong`/`flatMapToDouble` "for a follow-up". Those three were lifted by `111` into the `Φ.hone.lambda` pragma, matched nothing, and were reverted untouched by `701`. They were also absent from `DEFAULT_GREP_IN`, so a class using only the typed variants was skipped before any rule ran.

### What this does

Adds three sibling rules next to `455` — `456`/`457`/`458` — that convert a user-written `Stream.flatMapToInt`/`ToLong`/`ToDouble` into `Stream.mapMultiToInt`/`ToLong`/`ToDouble`. The adapter shape `455` already synthesises works unchanged: `(x, c) -> mapper(x).forEach(c)`, since `IntStream.forEach` takes precisely the `IntConsumer` that `mapMultiToInt` hands down (same for long/double). Each rule differs from `455` only in:

- `stream.method`/`stream.signature` in the pattern (`flatMapToX`, result `IntStream`/`LongStream`/`DoubleStream`) and result (`mapMultiToX`);
- the adapter's second parameter type (`Ljava/util/function/IntConsumer;` etc.);
- the `forEach` `invokeinterface` owner and descriptor.

`flatMapToInt`/`flatMapToLong`/`flatMapToDouble` are added to `DEFAULT_GREP_IN`. Since `501`/`502` already emit `mapMultiToInt`/`ToLong`/`ToDouble` dispatches, the jeo roundtrip of these calls is proven. The same caveats `455` accepts apply (null-skip and auto-close of the substream are not reproduced). The primitive-side `IntStream.flatMap` is left as a separate follow-up, out of scope here.

### Tests

- `src/test/phino/456|457|458-*.yml` — single-rule packs pinning each conversion.
- `src/test/resources/org/eolang/hone/optimize/streams/flatmaptox.yml` — dedicated e2e fixture exercising all three typed variants with order- and value-sensitive reduces.
- `all-ops.yml` `after` counts shifted (`invokestatic` 27→30, `return` 14→17): one synthesised `flatadapt` per typed variant adds one `invokestatic` (mapper) and one `return`.

### Verification

Ran the real pipeline (host `phino` 0.0.77 + `jeo` 0.15.1) against both the `all-ops` and the new `flatmaptox` sources: `BUILD SUCCESS`, the optimized classes reproduce the baseline output byte-for-byte (`54` and `-2017196980 5869572705746538085 1839030`), and the opcode counts match the pinned values. The deep phino-pack tests and the full non-deep suite pass.

https://claude.ai/code/session_012JbGWRfZsWSz9avvuFDhKi

---
_Generated by [Claude Code](https://claude.ai/code/session_012JbGWRfZsWSz9avvuFDhKi)_

### DIFF
--- a/src/main/java/org/eolang/hone/OptimizeMojo.java
+++ b/src/main/java/org/eolang/hone/OptimizeMojo.java
@@ -105,6 +105,9 @@ public final class OptimizeMojo extends AbstractMojo {
         "distinct",
         "skip",
         "flatMap",
+        "flatMapToInt",
+        "flatMapToLong",
+        "flatMapToDouble",
         "mapMulti"
     ).toString();
 
--- a/src/main/resources/org/eolang/hone/rules/streams/4xx/456-flatMapToInt-to-mapMultiToInt.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/4xx/456-flatMapToInt-to-mapMultiToInt.phr
@@ -1,0 +1,153 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# The IntStream sibling of 455-flatMap-to-mapMulti. Converts a user-written
+# Stream.flatMapToInt into an equivalent Stream.mapMultiToInt. flatMapToInt takes
+# a Function<T, IntStream> and flattens the returned IntStreams; the same fan-out
+# is expressed by mapMultiToInt((x, c) -> mapper(x).forEach(c)), where c is the
+# downstream IntConsumer. 111 lifts the flatMapToInt invokedynamic into a
+# Φ.hone.lambda carrying stream.method = "flatMapToInt" and a handle-6 target to a
+# non-capturing lambda$ body of signature (L<elem>;)Ljava/util/stream/IntStream;.
+# This rule rewrites that pragma into the shape 111 produces for a user
+# mapMultiToInt — a Φ.hone.lambda with interface BiConsumer, stream.method =
+# "mapMultiToInt", and a handle-6 target — whose target is a freshly synthesised
+# BiConsumer `flatadapt`:
+#
+#   flatadapt(x, c):  invokestatic lambda$(x) -> IntStream ; mapper applied to item
+#                     invokeinterface IntStream.forEach(c) ; drains it into c
+#
+# The mapper body is kept untouched. IntStream.forEach takes precisely the
+# IntConsumer that mapMultiToInt hands down, so the same adapter shape 455
+# synthesises works unchanged here. 501/502 already emit mapMultiToInt dispatches,
+# so the jeo roundtrip of these calls is proven.
+#
+# Scope: reference Stream.flatMapToInt only (mapper returns
+# java/util/stream/IntStream). Like 455, this assumes the mapper returns a
+# non-null, non-resource stream — the null-skip and auto-close of flatMapToInt are
+# not reproduced.
+name: flatMapToInt-to-mapMultiToInt
+pattern: |
+  ⟦
+    φ ↦ Φ.jeo.class,
+    𝐵-class-head,
+    𝜏-method ↦ ⟦
+      φ ↦ Φ.jeo.method,
+      𝐵-method-head,
+      body ↦ ⟦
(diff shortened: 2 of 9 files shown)


0crat:
@yegor256 Hey! Just a heads up - the branch name `claude/gifted-bell-dadoda` doesn't follow our [naming convention](https://www.yegor256.com/2014/04/15/github-guidelines.html), so per our policy that means **-6 points**. For next time, just use the ticket number - something like `711` would be perfect here. Your score is still looking good at [+903](https://www.eolang.org/zerocracy/objectionary-vitals.html) though! Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) when you get a chance.

0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this work. Here's how it breaks down according to our [policy](https://www.yegor256.com/2014/11/14/hits-of-code.html): **+16** base points, **-4** for exceeding 100 hits-of-code (684 total), **-8** for exceeding 400 hits-of-code, and **-8** for missing code review, with a **+8** adjustment to ensure minimum reward. While larger contributions show effort, our policy encourages focused, reviewable changes that maintain code quality. Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+933](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #714, opened 2026-06-10
Fix #713: keep-frame for a head-unboxing-map filter must be the primitive survivor

### LINKED ISSUE #704: VerifyError: filter().boxed() fuses to a distill whose keep-frame says 'long' while the stack holds a 'Long'

### LINKED ISSUE #713: VerifyError: mapToLong(...).filter(...) over Stream<Long> fuses to a distill whose keep-frame says 'Long' while the stack holds a primitive 'long'
A boxed-stream pipeline led by an unboxing map and containing a filter — for example `list.stream().mapToLong(Long::longValue).filter(n -> n % 2L == 0L).map(n -> n + 1L).sum()` from the `collection` benchmark of `objectionary/sabj25` — optimizes to a class that fails JVM verification with `java.lang.VerifyError: Inconsistent stackmap frames at branch target 12`. The fusion emits a synthetic `private static void distill_N(java.lang.Long, java.util.function.LongConsumer)` whose body is `aload_0; invokestatic lambda$hone$N(Ljava/lang/Long;)J; dup2; invokestatic (J)Z; ifne L; return; L: invokestatic (J)J; Long.longValue()J; aload_1; dup_x2; pop; LongConsumer.accept(J)V`, with a single `same_locals_1_stack_item` frame at the keep-label declaring the stack as `[ class java/lang/Long ]`. The verifier rejects the whole class, so none of the optimized methods can be loaded. Observed with hone-maven-plugin 0.27.2, JEO 0.15.2, PHINO 0.0.77, JDK 26.0.1 (Zulu); the same shape without the trailing map (`set.stream().mapToLong(Long::longValue).filter(n -> n % 3L == 0L).sum()`) fails identically.

The keep-frame type comes from `src/main/resources/org/eolang/hone/rules/streams/5xx/506-distill-filter-frame.phr:59-71`, which derives the frame item from the distill's `bridge-input` on the premise that a filter returns the item it was given. That premise breaks once an object-to-primitive map stage is fused into the distill ahead of the filter: the head `lambda$hone$N:(Ljava/lang/Long;)J` has already moved the element into the primitive domain, so the `dup2`-ed survivor at the keep-label is a two-slot `long` while the frame declares a one-slot `java/lang/Long`. Consistently, `src/main/resources/org/eolang/hone/rules/streams/4xx/412-distill-filter-returned-to-object.phr:69-76` promotes the distill's `returned` to the wrapper for any distill carrying the filter's `frame-item` marker, so 422 skips its `Long.valueOf(J)` re-box while the tail still carries the `Long.longValue()` bridge unbox for the `LongConsumer` — now applied to a primitive `long` that is not a `Long`. A pure-primitive pipeline (`distill` with `bridge-input` `J`, frame `long`) and a filterless boxed one (which gets the matched `valueOf`+`longValue` pair from 422) both verify fine. This is the mirror image of the closed #704 — there the keep-frame said `long` while the stack held a `Long` — and it is observed on 0.27.2, which already contains the #704 fix.

The smallest fix: derive the keep-label frame item in 506, and the `when` guard of 412, from the element's type at the filter itself — the parameter type of the filter's predicate target, e.g. the `J` of the `(J)Z` — instead of `bridge-input`/`bridge-output`, so a head-fused unboxing map keeps the distill in the primitive domain: the frame stays `long`, `returned` stays `J`, and 422 re-boxes exactly as it already does for filterless distills.

yegor256:
Fixes #713.

## Problem

A boxed-stream pipeline led by an unboxing map and containing a filter — e.g.
`list.stream().mapToLong(Long::longValue).filter(n -> n % 3L == 0L).sum()` —
optimized to a class that failed JVM verification:

```
  ... distill_N(Ljava/lang/Long;Ljava/util/function/LongConsumer;)V @8: ifne
  Stackmap stack: { 'java/lang/Long' }   actual stack: { long, long_2nd, ... }
```

When the head `mapToLong` fuses in **ahead** of the filter it permanently
unboxes the element, so the `dup` before the `(J)Z` predicate is a two-slot
`dup2` and the survivor kept at the filter's keep-label is a two-slot primitive
`long`. Two rules misread that survivor as the wrapper the synthesized method
receives:

* **506** derived the keep-frame from the distill's `bridge-input` (still the
  wrapper `Long`), declaring a one-slot `java/lang/Long` over a two-slot `long`.
* **412** promoted the distill's `returned` to that wrapper, so **422** skipped
  its `Long.valueOf(J)` re-box while **451** still appended a `Long.longValue()`
  bridge unbox — a `longValue` applied to a primitive `long`.

This is the mirror image of the closed #704 (there the keep-frame said `long`
while the stack held a `Long`), and reproduces on 0.27.2 which already contains
the #704 fix.

## Fix

* **412** now also requires `accepted` to be primitive before promoting
  `returned`. A fused head map leaves `accepted` a wrapper, so the #713 shape is
  left primitive and 422 boxes the survivor exactly as for a filterless distill
  — the `valueOf`/`longValue` pair then type-checks.
* **501 / 511** thread the distill's `returned` onto the `distill-lambda`.
* **506** becomes the specific rule for **primitive-predicate** filters
  (predicate signature `(<primitive>)Z` with a wrapper `bridge-input`): it
  derives the keep-frame from `returned`, which 412 keeps primitive for #713 and
  (still) promotes to the wrapper for the #704 boxed filter.
* **New 507** fallback derives the keep-frame from `bridge-input` for every
  other marker — including object-predicate filters whose survivor stays the
  wrapper even when trailing maps drive `returned` to a primitive
  (`boxed-roundtrip`).

The predicate parameter type is what cleanly separates the head-unboxing-map
shape (`(J)Z`, survivor primitive) from a `boxed-roundtrip` object-predicate
filter (`(Ljava/lang/Integer;)Z`, survivor stays the wrapper).

## Tests

* New end-to-end fixture `optimize/streams/maptolong-filter.yml` (the issue's
  reproduction; `invokedynamic` 2 → 1).
* New single-rule packs: `412-distill-filter-returned-kept-primitive`,
  `506-distill-filter-frame-unboxed`, `507-distill-filter-frame-fallback`;
  updated `506-distill-filter-frame`.

## Verification

Built phino `0.0.77` and ran the real phino + jeo roundtrip on the host:

* The #713 reproduction now runs (`The result is 9`) with no `VerifyError`.
* #704 (`long-filter-boxed`) still verifies (`The result is 18`).
* All 30 `optimize/streams/*.yml` fixtures pass the compile → optimize → run
  roundtrip with correct output and no verification errors.
* All 51 single-rule phino packs pass (`mvn -Pdeep test
  -Dtest=OptimizeMojoTest#appliesPhinoRulesAsSpecifiedInYamlPack`: 51 tests,
  0 failures).

https://claude.ai/code/session_01XT3mf7ksijA6xbNasyYQ2c

---
_Generated by [Claude Code](https://claude.ai/code/session_01XT3mf7ksijA6xbNasyYQ2c)_

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/4xx/412-distill-filter-returned-to-object.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/4xx/412-distill-filter-returned-to-object.phr
@@ -27,6 +27,23 @@
 # predicate; only `returned` is promoted to the bridge-output wrapper so 422
 # sees a type-preserving distill and adds nothing. Runs after 411 and before
 # 421/422.
+#
+# The promotion is correct only when the survivor the filter keeps is itself
+# the wrapper — i.e. the element enters the filter boxed and is unboxed only
+# transiently for the predicate (a lone boxed filter, the #704 shape, whose
+# `accepted` is the primitive predicate shape J). Once an object-to-primitive
+# map fuses in AHEAD of the filter (issue #713,
+# e.g. `mapToLong(Long::longValue).filter(...)`), that head map has already
+# moved the element permanently into the primitive domain: 401 carries the
+# head map's wrapper `accepted` onto the fused distill, the dup before the
+# predicate is a two-slot `dup2`, and the kept survivor is the primitive, not
+# the wrapper. Promoting `returned` there would make 422 skip its valueOf
+# re-box while 451 still appends the bridge longValue — a Long.longValue()
+# applied to a primitive long that is not a Long, failing verification. So the
+# guard also requires `accepted` to be primitive: a fused head map leaves it a
+# wrapper, which keeps this rule (and the re-box skip) off the #713 shape and
+# lets 422 box the primitive survivor exactly as it does for a filterless
+# distill.
 name: distill-filter-returned-to-object
 pattern: |
   ⟦
@@ -72,3 +89,6 @@ when:
     - matches:
         - '^L.+;$'
         - 𝑒-bridge-output
+    - matches:
+        - '^[BCDFIJSZ]$'
+        - 𝑒-accepted
--- a/src/main/resources/org/eolang/hone/rules/streams/5xx/501-distill-to-mapMulti.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/5xx/501-distill-to-mapMulti.phr
@@ -76,6 +76,7 @@ result: |
       φ ↦ Φ.hone.distill-lambda,
       static ↦ 𝑒-static,
       bridge-input ↦ 𝑒-bridge-input,
+      returned ↦ 𝑒-returned,
       consumer ↦ 𝑒-consumer,
       target-method ↦ 𝑒-target-method,
       body ↦ 𝑒-body
--- a/src/main/resources/org/eolang/hone/rules/streams/5xx/506-distill-filter-frame.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/5xx/506-distill-filter-frame.phr
@@ -2,33 +2,57 @@
 # SPDX-License-Identifier: MIT
 ---
 # yamllint disable rule:line-length
-# Resolves the Φ.hone.frame-item marker that 304 leaves at a STATELESS
-# filter guard's keep-label, the non-capturing twin of how 503/504 resolve
-# the marker for stateful (captured) guards. A filter is type-transparent —
-# it returns the item it was given — so the keep-label's stackmap frame must
-# declare the item's real type. That type is the distill's bridge-input,
-# which is only fixed once 411 has decided whether the box/unbox round-trip
-# around the filter could be unwound back to a primitive: for an unwound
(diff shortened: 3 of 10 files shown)


0crat:
@yegor256 Thanks for your contribution! You've earned **+4 points** according to our [work policy](https://github.com/zerocracy/farm/blob/master/_posts/2018/2018-03-05-policy.md): **+16** base points, **-4** for exceeding 100 hits-of-code (359 total), and **-8** for missing code review. While the total contribution size shows good productivity, consider breaking larger changes into smaller, reviewable chunks to maximize your bonus potential. Your running score is [**+937**](https://www.eolang.org/zerocracy/objectionary-vitals.html) - [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the momentum going and don't forget to check your [Zerocracy account](https://www.zerocracy.com/)!

Issue #717, opened 2026-06-11
Fused skip() races and breaks encounter order on parallel streams


yegor256:
The fused `skip()` is unsound on parallel streams. Rule `src/main/resources/org/eolang/hone/rules/streams/3xx/308-skip-to-distill.phr` (around lines 52–78) carries the countdown state as a plain one-cell `long[]` (`newarray` 11, then `laload`/`lastore` per item), captured into the single fused `BiConsumer`. On a parallel stream, several ForkJoin workers read-decrement-write that cell concurrently, updates are lost, and the wrong number of elements is dropped. This is the same race class that #715 fixed for `distinct()` by switching to `ConcurrentHashMap.newKeySet()`.

For `skip()`, however, thread-safety alone cannot fix it: `Stream.skip(n)` on an ordered stream must drop the first `n` elements in encounter order, while any per-item counter — even an atomic one — drops the first `n` in arrival order, a nondeterministic subset. The existing `parallel-distinct.yml` regression pack does not cover this, since it tests `distinct()` only, whose contract is order-free.

The smallest sound fix is a guard rather than a stronger counter: do not lift `skip()` (rules 220/308) unless the pipeline provably stays sequential, e.g. the stream is created, never handed to `parallel()`, and terminated within the same method. A regression pack mirroring `parallel-distinct.yml`, with `LongStream.range(0, N).parallel().skip(K).sum()` checked against the closed-form answer, would demonstrate the miscompilation.

yegor256:
@rultor release, tag is `0.27.5`

PR #721, opened 2026-06-11
Fuse dropWhile() via a sticky-latch distill (#718)

### LINKED ISSUE #718: Fuse dropWhile() via a sticky-latch distill, guarded to sequential pipelines
`dropWhile()` is the only stateful sink-free operation with no lifting rule: nothing under `src/main/resources/org/eolang/hone/rules/streams/` recognizes it, so it flows through unoptimized and splits an otherwise fusible `map(...).dropWhile(...).map(...)` chain into three stages. #540 requested this and was closed without an implementation, back when the design treated `dropWhile` as unfusable because of parallel streams.

That objection is no stronger for `dropWhile` than for the already-fused `skip()`: both drop a prefix defined by encounter order, which per-item state cannot identify on a parallel stream (see #717). Under the sequential-only guard proposed there, a one-shot sticky latch implements `dropWhile` faithfully: while the latch is unset, evaluate the predicate and drop on true; on the first false, set the latch; afterwards pass every item through without evaluating the predicate again.

The concrete change is a recognition rule in `streams/2xx/` (mirroring skip's) plus a `dropWhile-to-distill` rule in `streams/3xx/` modeled on `308-skip-to-distill.phr`, carrying the latch as one more state cell in the list that rule 502 builds in front of the dispatch — applied only under the same provably-sequential guard as #717. This closes the gap between the implementation and the claim that all sink-free operations of the Stream API are fused.

yegor256:
Closes #718.

`dropWhile()` was the only stateful sink-free `Stream` operation with no lifting rule, so it flowed through unoptimized and split an otherwise fusible `map().dropWhile().map()` chain into three stages. This adds the recognition + fold pair the issue asks for, plus the supporting wiring.

## What's here

- **`streams/2xx/208-lambda-to-dropwhile.phr`** — recognises a `Predicate` lambda immediately consumed by `Stream.dropWhile` and abstracts it into a `Φ.hone.dropWhile` pragma. This mirrors `201-lambda-to-filter` line for line (dropWhile carries a predicate, so — unlike `distinct`/`skip` — the element type and target handle come straight off the lambda). A `static ↦ "true"` match and a `\)Z$` guard on the target signature keep the out-of-scope shapes (primitive streams, captured/non-static predicates, `Boolean`-returning method references) as plain bytecode, so no pragma is produced that the fold can't lower.
- **`streams/3xx/310-dropwhile-to-distill.phr`** — folds the pragma into a stateful `Φ.hone.distill` carrying a one-shot **`int[1]` sticky latch** (the `dropWhile` analogue of `skip`'s `long[1]` countdown). The latch rides the same captured-`List` state channel `307`/`308` use. The body is `304`'s filter guard wrapped in the latch:
  - while the latch is unset, evaluate the predicate and `return` (drop) on true;
  - on the first false, set the latch;
  - afterwards pass every item through **without re-evaluating the predicate** (the dropWhile contract — re-evaluating would wrongly drop a later element the predicate returns true for).

  The latched branch and the predicate-false branch both reach a **single keep-label** with `[item]` on the stack, so the existing `503` `frame-item` helper is enough — no new per-branch frame rule is needed.

### Supporting changes
- **`512`**: widen the state-wrapper `max-locals` `4 → 6` for the latch guard's two scratch locals (the latch array and the negated predicate result; both dead by the keep-label, so `503`'s four-local frame still holds; unused by `distinct`/`skip`).
- **`283`**: insert the throwaway `dup` before a `filter` that follows a `dropWhile`, exactly as already done after `distinct`/`skip` — otherwise such a filter would eat the only item.
- **Tests**: phino packs for `208` and `310`; an end-to-end optimize fixture (`optimize/streams/dropwhile.yml`).
- **README**: list `dropWhile` among the fused operations.

## Verification (phino 0.0.77 + jeo, run locally)
- the `208` and `310` phino packs **pass** under the test harness, and **all 56 packs stay green** (no regression);
- the optimize pipeline **roundtrips through jeo to verifying bytecode that executes correctly** — `The result is 4 4` — fusing each `map().dropWhile().map()` into one `Stream.mapMulti` (two `distill_*` methods, two `int[1]` latches, two captured `ArrayList`s, `invokedynamic` 6 → 2);
- `dropwhile.yml`'s `log` strings and `before`/`after` opcode counts are the **measured** values (the only fix from my first pass was `after.invokestatic` 24 → 26).

## Relationship to #717 (sequential guard)
The issue specifies this should apply "only under the same provably-sequential guard as #717". That guard does not exist yet — and `skip()` (the model here) is itself currently **unguarded** (the bug #717 reports). This PR deliberately matches `skip`'s current behaviour: when #717 lands its sequential-only guard, it should cover the fused `dropWhile` the same way (both drop an encounter-order prefix that per-item state can't identify on a parallel stream). Happy to add the guard here once its shape is decided, or leave it to #717.

## Scope (v1, mirroring skip's narrow v1)
Object `java/util/stream/Stream` only; static, `Z`-returning predicates; element type preserved across fused neighbours. A *lone* `dropWhile` is **not** reverted to a native `Stream.dropWhile` the way `442` reverts a lone `skip` (rebuilding the predicate lambda is a follow-up) — it simply lowers to its own `mapMulti`, which is correct, just not faster than native.

https://claude.ai/code/session_01XEXK4gnUC8MoTqkWuVkuXn

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,7 +292,7 @@ The [hone paper][hone-paper] discusses this property,
 ## Why Short-Circuiting Operations Are Not Fused
 
 The pipeline deliberately fuses only non-short-circuiting operations
-  (`filter`, `map`, `peek`, `distinct`, `skip`, `flatMap`,
+  (`filter`, `map`, `peek`, `distinct`, `skip`, `dropWhile`, `flatMap`,
   the primitive conversions, and `mapMulti` itself);
   the two short-circuiting intermediate operations of the JDK,
   `limit()` and `takeWhile()`, are left untouched.
--- a/src/main/resources/org/eolang/hone/rules/streams/2xx/208-lambda-to-dropwhile.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/2xx/208-lambda-to-dropwhile.phr
@@ -1,0 +1,103 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# Recognises a stream `.dropWhile(p)` call — a Φ.hone.lambda whose SAM is a
+# Predicate immediately consumed by `Stream.dropWhile` — and abstracts it into
+# a single Φ.hone.dropWhile pragma. The shape is exactly the one 201 matches for
+# `filter` (a Predicate-returning lambda followed by the stream call); only the
+# `stream.method` differs, so this rule mirrors 201 line for line and 310 folds
+# the result into a stateful distill the same way 308 folds skip.
+#
+# dropWhile is the last sink-free Stream operation with no lifting rule (#718).
+# Like filter it carries a predicate, so — unlike distinct/skip — recognition
+# reads the element type and the target handle straight off the lambda; the only
+# extra ingredient at 310 is a one-shot sticky latch, the dropWhile analogue of
+# skip's countdown counter.
+#
+# v1 scope (mirrors skip's narrow v1): only the object `java/util/stream/Stream`
+# is handled, the predicate must be `static` (the normalised lambda shape 121
+# leaves), and its target must return a primitive `Z`. A `static ↦ "true"` match
+# and a `\)Z$` guard on the target signature keep the three other shapes —
+# primitive streams, captured (non-static) predicates, and method references to
+# `Boolean`-returning methods (the o-filter case 305 handles for filter) — as
+# plain bytecode, so no Φ.hone.dropWhile pragma is ever produced that 310 cannot
+# fold back to assemblable code.
+name: lambda-to-dropwhile
+pattern: |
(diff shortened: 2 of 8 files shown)


0crat:
@yegor256 Hey! Nice work on that contribution! 🎯 You've earned **+4 points** for this one: started with **+16** as the base reward, then **-4** for hitting over 100 hits-of-code (you had 398), and **-8** for skipping the code review step. I know it might seem like a lot of deductions, but remember our policy rewards quality over quantity - keeping contributions under 100 hits-of-code and getting them reviewed really pays off! Your running score is [+1005](https://www.eolang.org/zerocracy/objectionary-vitals.html), so [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too!

Issue #715, opened 2026-06-10
Fused distinct() races on parallel streams: shared HashSet state yields wrong, non-deterministic sums


yegor256:
The distinct-to-distill rewrite turns `Stream.distinct()` into a fused `mapMulti`/`mapMultiToLong` whose lambda deduplicates through a plain `java.util.HashSet` carried as captured state: `216-recognize-distinct.phr` matches the bare `Stream.distinct()` opcode, and `src/main/resources/org/eolang/hone/rules/streams/3xx/307-distinct-to-distill.phr:46` builds the distill state around `new java/util/HashSet`. Nothing in the pattern checks whether the pipeline is parallel.

This is wrong for parallel streams: the fused `BiConsumer` is invoked concurrently from multiple ForkJoin worker threads, and `HashSet.add` is not thread-safe, so racing threads both see "unseen" and duplicates slip through. The original `distinct()` handles parallelism inside the JDK. We hit this with hone `1.0-SNAPSHOT` on the sabj25 `sources` benchmark, whose pipeline `list.parallelStream().unordered().map(n -> n % 200_000L).distinct().mapToLong(Long::longValue).sum()` fails its `verified()` check after optimization with "the sum 2350048338543 does not match the expected 2344733191776" in one JMH fork and a different wrong sum (2346109911864) in the next. A five-line jshell equivalent of the rewritten bytecode (a captured `HashSet` guarding `sink.accept` inside `mapMulti` on that parallel stream) returns 28591876565, 23641333029, 23961802202 on consecutive runs against the expected 19999900000.

The smallest fix is to guard the rewrite: make `216-recognize-distinct.phr` (or `307`) bail out when the receiver may be parallel — for instance, only match pipelines whose source is a provably sequential `stream()`/`LongStream.range`-style call. Alternatively, keep the rewrite but build the seen-set with `ConcurrentHashMap.newKeySet()` instead of `HashSet`, which stays correct under concurrent invocation.

yegor256:
I suggest we use `ConcurrentHashMap.newKeySet()`

yegor256:
@rultor release, tag is `0.27.3`

0crat:
[...]

PR #722, opened 2026-06-11
#719: guard stateful skip lifting against parallel pipelines

### LINKED ISSUE #719: Stateful lifting needs a provably-sequential pipeline guard
The rules that lift stateful operations apply regardless of how the pipeline executes: `308-skip-to-distill.phr` (and `220` that feeds it) fires whether or not the stream may run in parallel, and nothing in `src/main/resources/org/eolang/hone/rules/streams/` inspects the pipeline for `parallel()`, `parallelStream()`, or for the stream escaping the enclosing method. Both #717 (fused `skip()` is unsound on parallel streams) and #718 (fusing `dropWhile()`) need the same missing precondition, so it deserves its own tracking issue.

The soundness condition is that arrival order equals encounter order, which holds when the stream provably stays sequential: it is created in the method, every stage of the chain is visible there, no stage is `parallel()` or `parallelStream()`, and the stream reference is consumed by a terminal operation in the same method rather than returned or stored.

The smallest concrete change is an early recognition rule that stamps a pipeline-level marker pragma (in the spirit of the existing `2xx` recognition family) when the whole chain from source to terminal matches this shape, and a `when` clause on the stateful lifting rules (`220`/`308`, and the future `dropWhile` rules) requiring that marker. Stateless lifting stays unconditional.

yegor256:
Fixes #719.

## Problem

`220-recognize-skip` → `308-skip-to-distill` lift every `.skip(n)` into a stateful countdown distill that fuses into one `Stream.mapMulti`. That rewrite is sound only when **arrival order equals encounter order**, i.e. when the pipeline provably stays sequential. On a parallel stream several `ForkJoin` workers drive the single captured `long[1]` counter concurrently, so it both races and counts in *arrival* order rather than *encounter* order — the wrong elements are dropped and the result is non-deterministic (#717). This is the order-dependent analogue of the parallel `distinct` race fixed in #715, except no thread-safe data structure can recover encounter order once it's lost.

## Approach (note the deviation from the issue text)

The issue proposes stamping a positive "provably-sequential" marker and gating `220`/`308` on it. While implementing I hit a structural limitation: **phino patterns cannot test a binding group for the *absence* of a `parallel()` call**, and an optimistically-stamped marker that's retracted later races with its own cleanup rule in phino's fixed-point evaluation.

So this PR takes the **dual, positive route already proven by `442-unfused-skip-to-invokeinterface`**: once `220` has produced a `Φ.hone.skip`, if the same method body *also* contains a `parallel()`/`parallelStream()` call, the skip pragma is reverted back to a native `invokeinterface Stream.skip:(J)…` **before `308` (3xx) ever folds it**. The native `Stream.skip(n)` honours the ordered/parallel contract, so the result stays correct; only the fusion optimization is forfeited on that method. The reverted opcode carries `reverted ↦ Φ.true` (the same fifth-operand guard `442` uses) so `220` cannot re-recognise it and the cycle cannot spin.

Because binding groups are order-preserving, the two layouts need two rules:
- **`222-parallel-reverts-skip`** — parallel call *before* the skip (`source.parallel()…skip(n)`, `coll.parallelStream()…skip(n)`)
- **`223-parallel-reverts-skip-after`** — parallel call *after* the skip (`source.skip(n).parallel()`)

Detection is intentionally over-broad (any parallel call in the method reverts every skip in it); over-reverting only forfeits an optimization and stays sound. The same guard extends verbatim to a future `dropWhile` lifting (#718).

`308` needs no change: the skip is already reverted before it runs, so it never sees a parallel-method skip. I kept the change to recognition/revert rather than adding a `when` to `308` to minimise risk.

## Changes

- `streams/2xx/222-parallel-reverts-skip.phr`, `streams/2xx/223-parallel-reverts-skip-after.phr`
- single-rule packs `src/test/phino/222-*.yml`, `223-*.yml`
- end-to-end fixture `optimize/streams/parallel-skip.yml` (runtime-only assertions, like `parallel-distinct.yml`): a 1,000,000-element ordered parallel range with an identity `map` neighbour, `skip(10)`, summed — correct = `500000499945`; the buggy fused skip gave a wrong, varying sum.
- README section "Why `skip` Is Only Fused on Sequential Pipelines".

## Validation

Validated locally against the prebuilt **phino 0.0.77** binary (the same one CI installs):
- both single-rule packs (`222`/`223`) rewrite as expected and every `expected` pattern matches;
- the rules are **inert without a `parallel()` call** (a sequential `skip` keeps its `Φ.hone.skip` pragma untouched, so `skip.yml` is unaffected);
- the **full 92-rule `streams/*` set terminates** on a parallel-skip input, leaving the skip as a native `Stream.skip` with no re-recognition loop.

The end-to-end JVM roundtrip + runtime-result assertion still depends on the `@Tag("deep")` CI jobs (`deep` / `codecov`), which run the real jeo+phino pipeline.

https://claude.ai/code/session_016Wqrw3TdkSDfiVbkTYNJqG

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,26 @@ Instead, `limit` and `takeWhile` act as barriers:
   fuse into their own `mapMulti` calls independently,
   while the native short-circuiting call stays in place.
 
+## Why `skip` Is Only Fused on Sequential Pipelines
+
+`skip(n)` is fused (it lowers to a countdown counter inside the
+  `mapMulti` body, see `308-skip-to-distill`), but only when the pipeline
+  provably stays sequential.
+The fused counter is correct only if arrival order equals encounter order:
+  on a parallel stream several `ForkJoin` workers drive the one captured
+  counter concurrently, so it both races and counts in arrival order rather
+  than encounter order, dropping the wrong elements
+  (this is the order-dependent analogue of the parallel `distinct` race
+  fixed in #715, except that no thread-safe data structure can recover
+  encounter order once it is lost).
+The JDK's native `Stream.skip(n)` honours the ordered/parallel contract,
+  so when a method contains a `parallel()` or `parallelStream()` call the
+  rules `222-parallel-reverts-skip` and `223-parallel-reverts-skip-after`
+  revert the recognised `skip` back to the native call before `308` can fold
+  it (#719, #717).
+The same revert-on-parallel guard applies verbatim to a future `dropWhile`
+  lifting, which is order-dependent for the same reason.
+
 ## How to Use in Gradle
 
 You can use this plugin with [Gradle] too, but it requires
--- a/src/main/resources/org/eolang/hone/rules/streams/2xx/222-parallel-reverts-skip.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/2xx/222-parallel-reverts-skip.phr
@@ -83,9 +83,7 @@ when:
     - or:
         - eq: [𝜏-op, 'invokeinterface']
         - eq: [𝜏-op, 'invokevirtual']
-    - or:
-        - eq: [𝑒-pmethod, 'parallel']
-        - eq: [𝑒-pmethod, 'parallelStream']
+    - matches: ['^(parallel|parallelStream)$', 𝑒-pmethod]
 where:
   - meta: 𝜏-push-back
     function: random-tau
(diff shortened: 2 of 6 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this: **+16** as a basis; **+14** for your 280 hits-of-code (280 × 0.05); **-4** for exceeding 100 hits-of-code; **-8** for lacking code review; **-14** adjustment to stay within policy limits. While your code size was substantial, getting a proper review next time will significantly boost your score according to our [policy](https://www.yegor256.com/2018/03/06/speed-vs-quality.html). Your running score is [+1009](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too!

PR #723, opened 2026-06-11
#717: guard skip() against parallel-stream miscompilation

### LINKED ISSUE #717: Fused skip() races and breaks encounter order on parallel streams

yegor256:
Fixes #717.

## Problem

A fused `skip()` is unsound on a parallel stream. The `220 → 308 → 502` path lowers `skip()` into a single `Stream.mapMulti` dispatch whose countdown state is a one-cell `long[]` captured into the **one** `BiConsumer` that backs the fused dispatch. On a parallel stream several `ForkJoin` workers decrement that cell concurrently (lost updates), and — fatally — a per-item counter drops the first `n` elements in **arrival** order while `Stream.skip(n)` on an ordered stream must drop them in **encounter** order.

Unlike the #715 `distinct()` race, thread-safety alone cannot repair this: `skip()`'s contract is order-dependent, so no counter (atomic or not) reproduces it on a parallel pipeline.

## Fix

A guard, not a stronger counter (as suggested in the issue). Three new rules taint a `skip()` that shares its method with a `parallel()`/`parallelStream()` call by appending a `parallel-unsafe` operand to its `invokeinterface`. Rule `220`'s recognition pattern is closed (`ρ ↦ ∅`), so the extra operand stops it folding the skip into a distill; the skip stays a native `Stream.skip(n)` — correct on sequential and parallel streams alike — and acts as a fusion barrier like `limit`/`takeWhile`. jeo ignores the extra operand on assemble (the same trick `442` uses with `reverted`).

- **213** — `parallelStream()` source upstream of the skip.
- **214** — explicit `.parallel()` upstream of the skip.
- **215** — `.parallel()` downstream of the skip (it toggles the whole pipeline).
- **220** — comment noting the guard and why a fused skip must not lift.

## Tests

- Single-rule packs `213`/`214`/`215` under `src/test/phino/`.
- `parallel-skip.yml` e2e regression (mirroring `parallel-distinct.yml`) asserting the deterministic sum `750000500000` for `list.parallelStream().map(+1).skip(500_000).map(*2).mapToLong(...).sum()`. On `master` the fused skip makes this wrong/non-deterministic.

## Known limitation

The guard is a **same-method** analysis: a stream handed in as a parameter and made parallel by the caller is invisible to it, so this is a `v1` best-effort guard rather than a whole-program proof of sequentiality. Documented in `README.md` ("Why `skip` Is Not Fused on Parallel Streams").

⚠️ I could not run the `deep` (phino-backed) suite locally — phino is not installed in this environment and there is no Docker fallback — so the `.phr` rules are validated by construction (mirroring existing rules `219`/`220`/`442`/`711`) and rely on CI's `-Pdeep` run.

https://claude.ai/code/session_01JGesjGXwyQdJsLk36VRpLJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01JGesjGXwyQdJsLk36VRpLJ)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,9 +336,9 @@ On a parallel stream several `ForkJoin` workers drive that `BiConsumer`
 This is the same race class that `distinct` had (see [#715][715]),
   but for `skip` thread-safety alone cannot fix it:
   `Stream.skip(n)` on an ordered stream must drop the first `n` elements
-  in *encounter* order,
+  in _encounter_ order,
   while any per-item counter — even an atomic one —
-  drops the first `n` in *arrival* order,
+  drops the first `n` in _arrival_ order,
   a non-deterministic subset.
 The fix (see [#717][717]) is therefore a guard, not a stronger counter:
   rules `213`–`215` taint a `skip()` that shares its method
--- a/README.md
+++ b/README.md
@@ -324,6 +324,35 @@ Instead, `limit` and `takeWhile` act as barriers:
   fuse into their own `mapMulti` calls independently,
   while the native short-circuiting call stays in place.
 
+## Why `skip` Is Not Fused on Parallel Streams
+
+`skip` is fused only when the pipeline provably stays sequential.
+The fused `skip` carries its countdown as a single `long[]` cell,
+  decremented once per element by the one `BiConsumer`
+  that backs the fused `Stream.mapMulti` dispatch.
+On a parallel stream several `ForkJoin` workers drive that `BiConsumer`
+  at once, so the read-decrement-write of the counter races
+  and the wrong number of elements is dropped.
+This is the same race class that `distinct` had (see [#715][715]),
+  but for `skip` thread-safety alone cannot fix it:
+  `Stream.skip(n)` on an ordered stream must drop the first `n` elements
+  in *encounter* order,
+  while any per-item counter — even an atomic one —
+  drops the first `n` in *arrival* order,
+  a non-deterministic subset.
+The fix (see [#717][717]) is therefore a guard, not a stronger counter:
+  rules `213`–`215` taint a `skip()` that shares its method
+  with a `parallel()` or `parallelStream()` call,
+  and the recognition rule `220` then leaves it
+  as a native `Stream.skip(n)` —
+  correct on sequential and parallel streams alike —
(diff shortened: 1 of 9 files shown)


yegor256:
Closing as superseded by #722 (commit 72cda5b), which already fixes #717 on `master`.

Both PRs guard `skip()` against the parallel-stream miscompilation and reach the same end-state (a native `Stream.skip(n)` left as a fusion barrier on parallel pipelines). #722's approach is the one to keep:

- It reverts the recognized skip pragma back to native *after* `220` (rules `222`/`223`), using `matches` to cover both `parallel()` and `parallelStream()` in one rule.
- It was validated against the prebuilt phino 0.0.77.

This PR took the dual route — tainting the skip *before* `220` recognizes it (rules `213`/`214`/`215`) — but it's functionally redundant with `222`/`223` and was not validated against phino locally. Keeping both would leave two overlapping guards, so closing this one.

---

PR #729, opened 2026-06-11
#725: reproduce flatMap's null-skip and sequential() in the mapMulti adapter

### LINKED ISSUE #725: 455 flatMap→mapMulti adapter drops flatMap's null-skip and sequential()
The `flatadapt` adapter synthesized by `455-flatMap-to-mapMulti.phr` (body at `src/main/resources/org/eolang/hone/rules/streams/4xx/455-flatMap-to-mapMulti.phr:118-138`, and its primitive siblings `456`/`457`/`458`) lowers to `aload0; invokestatic lambda$(x)→Stream; aload1; invokeinterface Stream.forEach(Consumer); return`. The JDK's `ReferencePipeline.flatMap` instead does, per element, `try (Stream result = mapper.apply(u)) { if (result != null) result.sequential().forEach(downstream); }`. The fused adapter reproduces neither the null guard nor the `sequential()` call.

Two observable differences result. First, when the mapper returns `null` the JDK silently skips that element, but the adapter calls `forEach` on the value and throws `NullPointerException` — a behavioral/correctness change. Second, the JDK drains the returned stream via `result.sequential()`, so a parallel sub-stream returned by the mapper is forced sequential before its elements reach the downstream `Consumer`; the adapter calls `forEach` directly, so a parallel sub-stream races into a `Consumer` that may not be thread-safe, producing wrong results. This is noted as a known limitation in the rule header (`455:31`).

Minimal fix, no exception table needed: in the `flatadapt` body add a null guard around the drain (`dup; ifnonnull L; pop; return; L: …` — the branch/label/`dup`/`pop` synthesis already exists in `308-skip-to-distill.phr`) and insert one `invokeinterface java/util/stream/BaseStream.sequential ()Ljava/util/stream/BaseStream;` before the `forEach`. Apply the same edit to `456`/`457`/`458`; `max-stack` stays 2. (The third gap, auto-close of the returned stream, needs try-finally synthesis and is tracked separately.)

yegor256:
Fixes #725.

## Problem

The `flatadapt` BiConsumer synthesised by `455-flatMap-to-mapMulti.phr` and its primitive siblings `456`/`457`/`458` lowered a `flatMap` into, per element:

```
aload0; invokestatic lambda$(x)→Stream; aload1; invokeinterface Stream.forEach(Consumer); return
```

The JDK's `ReferencePipeline.flatMap` instead does `if (result != null) result.sequential().forEach(downstream)`. The fused adapter reproduced neither contract, so:

1. **null-skip** — when the mapper returned `null` the adapter called `forEach` on it and threw `NullPointerException` instead of silently skipping the element.
2. **sequential()** — a parallel sub-stream returned by the mapper raced straight into a downstream `Consumer` that may not be thread-safe, instead of being forced sequential first.

## Fix

In each `flatadapt` body:

- Wrap the drain in a null guard — `dup; ifnonnull L; pop; return` (the same branch/label/`dup`/`pop` synthesis used by `308-skip-to-distill.phr`).
- Insert `invokeinterface java/util/stream/BaseStream.sequential ()Ljava/util/stream/BaseStream;` before the `forEach`.
- Emit an explicit `Φ.jeo.frame` (F_SAME1) at the keep-label. jeo's assembler does not compute stackmap frames, so a synthesized branch target needs the frame written out (as `506-distill-filter-frame.phr` does) — without it the reassembled class fails verification with *"Expecting a stackmap frame at branch target"*.

`max-stack` stays 2. The third gap noted in #725 — auto-close of the returned stream — needs try-finally synthesis and is left for a separate issue.

## Tests

- Updated the four `src/test/phino/*.yml` unit packs to assert the `ifnonnull` guard, the `Φ.jeo.frame`, and the `BaseStream.sequential` drain.
- Updated the pinned opcode counts in `flatmaptox.yml` (`invokeinterface` 9→12, `return` 5→8) and `flatmap-fusion.yml` (`invokeinterface` 6→7) for the added `sequential`/branch opcodes.
- Added `flatmap-null-skip.yml`, an end-to-end regression whose mapper returns `null` for odd elements: it throws `NullPointerException` (never printing its result line) without the guard, and reproduces the JDK's `75886572` with it.

## Validation

All four phino packs rewrite and match. The full disassemble → phino-rewrite → jeo-assemble pipeline was run on each fixture: every optimized class **verifies, runs, and prints the JDK-identical result**, with opcode counts matching the updated expectations exactly.

https://claude.ai/code/session_01CSz8mMdns1g4KLMCKtfAec

---
_Generated by [Claude Code](https://claude.ai/code/session_01CSz8mMdns1g4KLMCKtfAec)_

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/4xx/455-flatMap-to-mapMulti.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/4xx/455-flatMap-to-mapMulti.phr
@@ -14,6 +14,8 @@
 # handle-6 target — whose target is a freshly synthesised BiConsumer `flatadapt`:
 #
 #   flatadapt(x, c):  invokestatic lambda$(x) -> Stream ; mapper applied to item
+#                     dup; ifnonnull L; pop; return    ; flatMap's null-skip
+#                     L: invokeinterface BaseStream.sequential() ; force sequential
 #                     invokeinterface Stream.forEach(c) ; drains it into c
 #
 # The original mapper body (lambda$…) is kept untouched — the rule only wires the
@@ -26,9 +28,11 @@
 #
 # Scope: reference Stream.flatMap only (mapper returns java/util/stream/Stream).
 # The primitive flatMapToInt/Long/Double variants carry different SAM and result
-# types and are left for a follow-up. Like the JDK's own mapMulti-via-flatMap
-# bridge, this assumes the mapper returns a non-null, non-resource stream — the
-# null-skip and auto-close of flatMap are not reproduced.
+# types and are left for a follow-up. The adapter reproduces flatMap's null-skip
+# (a null mapper result is dropped rather than dereferenced) and its
+# `result.sequential()` drain (a parallel sub-stream is forced sequential before
+# its elements reach the downstream Consumer); the auto-close of the returned
+# stream still needs try-finally synthesis and is tracked separately.
 name: flatMap-to-mapMulti
 pattern: |
   ⟦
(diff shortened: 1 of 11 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this: **+16** as a basis, **+13.65** for your 273 hits-of-code (273 × 0.05), **-4** for exceeding 100 hits-of-code, **-8** for exceeding 400 hits-of-code, and **-8** for the lack of code review, capped at our minimum of 4 points. While your code volume shows great effort, consider getting a code review next time to avoid the -8 penalty and potentially earn more points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1085](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

Issue #720, opened 2026-06-11
rewrite.sh: parallel --retries=0 means INFINITE retries, hanging on a failing job


yegor256:
The GNU Parallel invocation in `src/main/resources/org/eolang/hone/scaffolding/rewrite.sh:239` passes `--retries=0`. In GNU Parallel, `--retries=0` does not mean "no retries" — its manual states "n=0 means infinite". So when a single rewrite job fails deterministically (the `rewrite_with_timeout` path exits non-zero on a genuine phino error, by design since #685), Parallel retries that same job forever and the whole `optimize` goal hangs with no further output, because Parallel buffers each job stream until completion.

This is what hung the `coverage` CI job for the full 180-minute timeout on `stleary/JSON-java`: phino 0.0.77 cannot re-parse its own `NaN` double output (filed upstream as objectionary/phino#789), the job for `JSONObject` exits 1, and `--retries=0` loops it indefinitely. Before #700 the stale-`pho` up-to-date guard masked it (the first retry no-opped as "already done"); after #700 every retry honestly redoes the work and never converges. This is a distinct mechanism from the realpath/BSD hang in #478.

The fix is to drop `--retries=0` from the `parallel` call so the default (no retry) applies; the existing `--halt-on-error=now,fail=1 --halt=now,fail=1` then fails the build in seconds on the first genuine failure, which is the fail-loud behaviour #685 intended.

0crat:
@yegor256 Hey there! 🎉 Nice work closing that issue – you've earned **+8 points** for getting it done! Your running score is now [+1129](https://www.eolang.org/zerocracy/objectionary-vitals.html), which is looking pretty solid. The quick turnaround definitely paid off here, and remember that faster fixes mean better bonuses, so keep up that momentum! Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for all the details.

PR #730, opened 2026-06-11
Kill phino on per-file timeout instead of orphaning it (#727)

### LINKED ISSUE #727: rewrite.sh per-file timeout orphans phino instead of killing it (rewrite.sh:139)
`rewrite_with_timeout` at `src/main/resources/org/eolang/hone/scaffolding/rewrite.sh:139` runs `timeout "${HONE_TIMEOUT}" "${0}" rewrite "$@"`. GNU `timeout` sends SIGTERM only to its direct child — the bash re-invocation of the script that runs the `rewrite` function — not to the `phino` process that bash launches in the foreground. When the limit is reached, bash exits (timeout returns 124, the branch at `rewrite.sh:142-144` logs "Timeout in …" and copies the input through), but the `phino` process is reparented to init and keeps running.

Across an optimize run with many slow files these orphaned `phino` processes accumulate and saturate the CPU, so the remaining files' rewrites slow down and time out as well — a cascade. On a large repository the optimize then never converges, even though every file is nominally being "skipped". Observed while optimizing `apache/commons-io` (`cf5f9a2`) with `-Dhone.timeout=120`: a `phino` process sat at ~37% CPU for 20+ minutes after its file's timeout should have stopped it, and the optimize goal never finished. This is distinct from #700 (which concerns the green-build accounting after the copy-through) and from #720 (which is about `parallel --retries`); here the problem is that the process is never actually killed.

The smallest fix is to make the timeout terminate the whole process tree rather than just the bash wrapper: launch the worker in its own process group (e.g. via `setsid`) and use `timeout -s KILL --kill-after=<n> …` so `phino` is killed too, or have the `rewrite` subshell trap SIGTERM and kill its `phino` child before exiting.

yegor256:
Fixes #727.

## Problem

`rewrite_with_timeout` ran `timeout "${HONE_TIMEOUT}" "${0}" rewrite "$@"`. GNU `timeout` sends its signal only to its **direct child** — the bash re-invocation of the script that runs the `rewrite` function — not to the `phino` process that bash launches in the foreground. When the limit was reached, bash exited but `phino` was reparented to init and kept running. Across a run with many slow files these orphaned `phino` processes accumulated, saturated the CPU, and made the remaining files time out too, so the optimize never converged.

## Fix

`rewrite_with_timeout` now:

- runs the worker in its **own session** via `setsid --wait`, so it (and the `phino` it spawns) live in a dedicated process group;
- starts a watchdog that, once `HONE_TIMEOUT` elapses, terminates the **whole process group** (`kill -TERM/-KILL -<pgid>`) and walks the process tree (`kill_tree`, children before parents) as a fallback, escalating from `TERM` to `KILL` after a grace period (`HONE_KILL_GRACE`, default 10s) — so `phino` is actually killed rather than orphaned;
- uses a flag file to distinguish a genuine **timeout** (copy the input through, exit 0, as before) from a real **failure** (refuse to copy through, propagate the exit code, as before);
- fails fast with a clear message in the pre-flight checks when `setsid` or `pgrep` is missing.

This is distinct from #700 (green-build accounting after the copy-through) and #720 (`parallel --retries`); here the process is now actually killed.

## Verification

Exercised the real (edited) `rewrite_with_timeout` dispatch with a stubbed `rewrite`:

- **Timeout**: orphaned grandchild is killed, input is copied through, exit 0, logs `Timeout in …`.
- **Fast failure**: exit code is propagated, output is **not** copied through, logs `Failure (exit code …) …` — and returns promptly (the leader-discovery loop breaks as soon as the worker exits, so fast files keep their low latency).
- **Fast success**: exit 0, output produced, no added latency.

`shellcheck` (0.10.0) and `bashate -i E006,E003` both pass clean.

https://claude.ai/code/session_01MNLnBX3HdSZso7rUNpYJsd

---
_Generated by [Claude Code](https://claude.ai/code/session_01MNLnBX3HdSZso7rUNpYJsd)_

### DIFF
--- a/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
@@ -128,6 +128,20 @@ function rewrite {
   fi
 }
 
+# Kill a process together with its whole descendant tree, escalating from a
+# given signal towards the leader. Children are signalled before their parents
+# so that nothing is reparented to init and left running. Used so that a
+# timed-out rewrite does not orphan the phino process it spawned (see #727).
+function kill_tree {
+  local sig="${1}"
+  local root="${2}"
+  local child
+  for child in $(pgrep -P "${root}" 2>/dev/null || true); do
+    kill_tree "${sig}" "${child}"
+  done
+  kill "-${sig}" "${root}" 2>/dev/null || true
+}
+
 function rewrite_with_timeout {
   idx=${1}
   phi=${2}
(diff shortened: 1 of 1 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned **+8 points** for this: **+16** as a basis per our policy, then **-8** for the lack of code review (as specified in our work policy, we deduct 8 points when there's no review process). Getting your code reviewed next time will help you earn the full bonus! Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1081](https://www.eolang.org/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too 🚀

PR #734, opened 2026-06-11
#733: break the 603-unbox-to-lambda / 205-lambda-to-unbox rewrite loop

### LINKED ISSUE #733: 603-unbox-to-lambda and 205-lambda-to-unbox form an infinite rewrite loop, crashing `optimize`
The `unbox-to-lambda` rule at `src/main/resources/org/eolang/hone/rules/streams/6xx/603-unbox-to-lambda.phr:8` rewrites every `Φ.hone.unbox` node into a `Φ.hone.lambda`, while `lambda-to-unbox` at `src/main/resources/org/eolang/hone/rules/streams/2xx/205-lambda-to-unbox.phr:8` rewrites that very same `Φ.hone.lambda` shape straight back into a `Φ.hone.unbox`. The two rules are exact inverses, and since the plugin hands the whole `streams/*` selection to a single `phino rewrite` invocation that runs to fixpoint, they ping-pong forever as soon as an un-fused `unbox` survives into the 6xx phase. phino gives up with `On rewriting step '1' of rule 'unbox-to-lambda' we got the same expression as we got at one of the previous step, it seems rewriting is looping`, refuses to copy the file through unoptimized, and the entire `optimize` goal fails with a non-zero exit code.

The oscillation is not coincidental: 603 has no `when` guard and its `where` block computes `method`, `interface`, `stream-method`, `lambda-signature` and `stream-signature` to be exactly the fields that 205's `when` guard checks (`applyAsInt` + `mapToInt` + `(L...;)I`, plus the Long and Double variants — see `205-lambda-to-unbox.phr:40-62`). The binding sets of 603's `result` and 205's `pattern` coincide, so whatever 603 produces, 205 immediately matches and reverts. The latest `coverage.yml` run ([27353155150](https://github.com/objectionary/hone-maven-plugin/actions/runs/27353155150)) tripped this on four independent projects — `MultiBackgroundInitializer` (commons-lang), `Traversable` (vavr), `CellExtractor$1` (commons-collections) and `GenericKeyedObjectPool` (commons-pool) — each of which carries a primitive `mapToInt`/`mapToLong`/`mapToDouble` that reaches lowering as a bare `unbox` rather than being folded into a `distill`.

The fix is to stop 603's freshly lowered lambda from being re-recognized by 205. The smallest change is a marker breaker: tag 603's `Φ.hone.lambda` result (`603-unbox-to-lambda.phr:16-31`) with one extra binding such as `lowered ↦ Φ.true`, which — because phino matches formations by their exact binding set — keeps 205's `pattern` from matching it, while leaving the downstream 7xx consumers of `Φ.hone.lambda` untouched (or, if needed, have them ignore the marker). That single guard turns the inverse pair into a one-way lowering and breaks the loop.

yegor256:
Fixes #733.

## Problem

`6xx/603-unbox-to-lambda` and `2xx/205-lambda-to-unbox` are exact inverses:
603 rewrites every `Φ.hone.unbox` into a `Φ.hone.lambda`, and 205 rewrites that
very same lambda shape straight back into a `Φ.hone.unbox`. Because the whole
`streams/*` selection is handed to a single `phino rewrite` invocation that runs
to fixpoint, an un-fused `unbox` surviving into the 6xx phase made the two rules
ping-pong forever. phino gave up with *"it seems rewriting is looping"*, refused
to copy the file through, and the entire `optimize` goal failed.

## Fix

603 now tags its freshly lowered lambda with a trailing `lowered ↦ Φ.true`
marker. Since phino matches formations by their **exact binding set**, the extra
binding stops 205 (and every other 2xx lambda consumer) from recognizing it —
turning the inverse pair into a one-way lowering. The marker is kept **last** so
that 701/702's trailing binding-set wildcard captures it.

`7xx/701-static-lambda-to-invokedynamic` and `7xx/702-nonstatic-lambda-to-invokedynamic`
(the only consumers of `Φ.hone.lambda` in 7xx) gain a trailing `𝐵-lowered`
wildcard so they still lower the marked lambda to `invokedynamic`; the wildcard
stays optional, so lambdas from every other producer keep matching too. The
marker is consumed by 701/702 and never reaches `jeo:assemble`.

## Reproduction / test

Adds `src/test/phino/733-unbox-lambda-no-loop.yml`, a phino pack that hands a
bare `Φ.hone.unbox` to rules `205` + `603`. Before the fix `phino rewrite` exits
non-zero (the loop); after the fix it terminates and settles on the marked
lambda, asserted via `⟦ 𝐵-a, φ ↦ Φ.hone.lambda, 𝐵-b, lowered ↦ Φ.true, 𝐵-c ⟧`.

## Phino version bump (0.0.79 → 0.0.80)

The `smoke`/`mvn`/`codecov` workflows install `phino-latest`, which advanced to
0.0.80 while the repo still pinned 0.0.79; `hone-it.sh` then aborted on the
version mismatch and `smoke` went red (unrelated to the loop fix). The pin is
bumped to 0.0.80 to match the published latest.

## Validation

Run with the now-pinned phino **0.0.80**:
- all 59 `mvn -Pdeep` phino rule packs pass, including the new #733 pack;
- the `603 → 701/702 → invokedynamic` chain still works for both static and
  nonstatic lambdas, with the marker dropped;
- the deep `optimize` fixtures show no regression from this change — the only
  two fixture mismatches (`allops`, `stateful`) reproduce **identically on
  `master`** (same opcode counts, with and without this change, under both
  0.0.79 and 0.0.80), i.e. they are pre-existing and unrelated to #733.

https://claude.ai/code/session_011wSNkn4pJuohMnZzKF6uDy


0crat:
@yegor256 Hey! Nice work on that contribution! 🎉 You picked up **+8 points** this time: started with the standard **+16 base points** but took a **-8 hit** since there wasn't any code review. I know reviews can slow things down, but they're actually part of how we calculate your bonus - plus they help catch issues early! Your total score is sitting pretty at **[+1117](https://www.eolang.org/zerocracy/objectionary-vitals.html)** now. [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the contributions flowing, and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates!

PR #737, opened 2026-06-11
#736: keep flatMap native when a short-circuit is downstream

### LINKED ISSUE #736: flatMap->mapMulti drops downstream short-circuiting, hangs on infinite inner streams
Rule `455-flatMap-to-mapMulti.phr` rewrites a user `Stream.flatMap` into a `mapMulti` whose synthesised adapter drains every inner stream with an unconditional `result.sequential().forEach(c)` (`src/main/resources/org/eolang/hone/rules/streams/4xx/455-flatMap-to-mapMulti.phr:168-172`, the `call-foreach` opcode); the primitive siblings `456`/`457`/`458` do the same. The rule's header comment claims the adapter "reproduces flatMap's full per-element contract," but it drops one part of that contract: downstream short-circuiting.

The JDK's `ReferencePipeline.flatMap` forwards `cancellationRequested()` downstream and, once a short-circuiting stage has fired, switches from `forEach` to a `tryAdvance` loop that stops part-way through an inner stream. `mapMulti` has no such mechanism — its lambda is handed a plain `Consumer`, so `forEach` always runs to completion for every source element. With a finite inner stream this merely wastes work, but with an unbounded or large inner stream feeding a short-circuit it never stops: `Stream.of(1L).flatMap(x -> Stream.iterate(x, i -> i + 1)).limit(5)` yields five elements with native flatMap and hangs forever after fusion. Because no `sabj25` benchmark combines an unbounded inner stream with a downstream short-circuit, the benchmark suite does not expose this.

A `mapMulti` cannot honour cancellation, so the fix mirrors the existing `222-parallel-reverts-skip` guard rather than trying to patch the adapter: leave `flatMap` native whenever the segment can reach a short-circuiting operator. Add a revert/guard rule that declines `455`/`456`/`457`/`458` when a `limit`, `takeWhile`, `findFirst`, `findAny`, `anyMatch`, `allMatch`, or `noneMatch` call appears downstream in the same method body, so those flatMaps stay unfused.

yegor256:
Fixes #736.

## Problem

`455-flatMap-to-mapMulti` (and its primitive siblings `456`/`457`/`458`) rewrite a user `Stream.flatMap` into a `mapMulti` whose synthesised adapter drains every inner stream with an unconditional `result.sequential().forEach(c)`. The adapter reproduces flatMap's per-element fan-out, null-skip and auto-close — but it drops **downstream short-circuiting**.

The JDK's `ReferencePipeline.flatMap` forwards `cancellationRequested()` and, once a short-circuiting stage has fired, stops part-way through an inner stream. A `mapMulti` body is handed a plain `Consumer`, so `forEach` always runs to completion. With an unbounded inner stream feeding a short-circuit it never stops:

```java
Stream.of(1L).flatMap(x -> Stream.iterate(x, i -> i + 1)).limit(5)
```

yields five elements with native flatMap but **hangs forever** after fusion. No `sabj25` benchmark combines an unbounded inner stream with a downstream short-circuit, so the suite never exposed it.

## Fix

A `mapMulti` cannot honour cancellation, so — exactly as `222-parallel-reverts-skip` keeps a skip native on a parallel pipeline rather than patching the adapter — the new **`454-flatMap-revert-on-short-circuit`** leaves the flatMap native whenever the same method body can reach a short-circuiting operator (`limit`, `takeWhile`, `findFirst`, `findAny`, `anyMatch`, `allMatch`, `noneMatch`) downstream of it.

Once `111` has lifted the flatMap into a `Φ.hone.lambda`, the guard reverts it straight back to the native `invokedynamic` + `invokeinterface` pair `701` would otherwise emit — before `455` (later in the 4xx pass) ever converts it. The reverted `invokeinterface` carries the `reverted ↦ Φ.true` fifth-operand marker (same trick `442`/`222` use) so `111`'s four-operand lift pattern cannot re-recognise it and the `111 → 454 → 701` cycle cannot spin. Because the lambda is gone, `455`/`456`/`457`/`458` simply never match — no change to those rules is needed.

Detection is intentionally over-broad (any matching short-circuit anywhere after the flatMap reverts it); over-reverting only forfeits the mapMulti fusion on that method, so it stays sound.

## Tests / docs

- `src/test/phino/454-flatMap-revert-on-short-circuit.yml` — single-rule pack asserting a flatMap lambda followed by `limit` reverts to native `invokedynamic`/`invokeinterface` with the `reverted` guard, leaving the limit call untouched.
- `src/test/resources/org/eolang/hone/optimize/streams/flatmap-short-circuit.yml` — end-to-end fixture with an unbounded inner stream + `limit(5)` that now terminates and prints `15` (it hung before the guard).
- `README.md` — documents the guard in the short-circuiting section.

https://claude.ai/code/session_017b3dCmGuPdSw9qhZjyKrpT

---
_Generated by [Claude Code](https://claude.ai/code/session_017b3dCmGuPdSw9qhZjyKrpT)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,24 @@ Instead, `limit` and `takeWhile` act as barriers:
   fuse into their own `mapMulti` calls independently,
   while the native short-circuiting call stays in place.
 
+The same blind spot reaches `flatMap` from the other direction (#736).
+A `flatMap` is fusable — `455-flatMap-to-mapMulti` and its primitive siblings
+  rewrite it into a `mapMulti` whose adapter drains each inner stream with
+  `result.sequential().forEach(c)` — but that adapter, being just a `mapMulti`
+  body, cannot see `cancellationRequested()` either.
+The JDK's `ReferencePipeline.flatMap` does forward cancellation:
+  once a downstream short-circuit has fired it stops part-way through an inner
+  stream, so `Stream.of(1L).flatMap(x -> Stream.iterate(x, i -> i + 1)).limit(5)`
+  yields five elements and terminates; the fused `forEach` adapter, by contrast,
+  runs to completion for every source element and hangs forever on the
+  unbounded inner stream.
+Because a `mapMulti` cannot honour cancellation, the cure mirrors the
+  revert-on-parallel guard above rather than patching the adapter:
+  `454-flatMap-revert-on-short-circuit` reverts the lifted `flatMap` back to its
+  native call whenever the same method body reaches a `limit`, `takeWhile`,
+  `findFirst`, `findAny`, `anyMatch`, `allMatch`, or `noneMatch` downstream of
+  it, so those `flatMap`s stay unfused.
+
 ## Why `skip` Is Only Fused on Sequential Pipelines
 
 `skip(n)` is fused (it lowers to a countdown counter inside the
--- a/src/main/resources/org/eolang/hone/rules/streams/4xx/454-flatMap-revert-on-short-circuit.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/4xx/454-flatMap-revert-on-short-circuit.phr
@@ -1,0 +1,144 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# Guards 455/456/457/458 against downstream short-circuiting (#736).
+#
+# 455 (and its primitive siblings 456/457/458) rewrite a user `Stream.flatMap`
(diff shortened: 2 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this submission: **+16** as the baseline reward, **+0.05 × 257 = +12.85** (capped at +16) for your 257 hits-of-code, but **-4** for exceeding 100 hits-of-code and **-8** for missing code review as outlined in our policy. Consider breaking larger contributions into smaller, reviewable chunks to maximize your rewards and maintain code quality - [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+1145](https://www.eolang.org/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #739, opened 2026-06-11
#738: guard distinct() against parallel pipelines

### LINKED ISSUE #738: Fused distinct() violates encounter-order stability on ordered parallel streams
Rule `307-distinct-to-distill.phr` folds `Stream.distinct()` into a distill whose seen-set is built by `ConcurrentHashMap.newKeySet()` and populated in arrival order (`src/main/resources/org/eolang/hone/rules/streams/3xx/307-distinct-to-distill.phr:52-60`). Neither the matcher (`216-recognize-distinct.phr`) nor this rule checks whether the pipeline is parallel or ordered, so the rewrite runs unconditionally. The switch to a concurrent set closed the data race of #715, but only made the set thread-safe — it did not address which element survives among equals.

The JDK's `distinct()` is documented as *stable* on an ordered stream: among equal elements it keeps the one first in encounter order. An arrival-order set on an ordered parallel stream instead keeps whichever equal element wins the race to `add()` — arrival-order-first, which is neither encounter-order-first nor deterministic across runs. For primitives and substitutable-`equals` values this is unobservable (and #715's fix suffices), but for equal-but-distinguishable objects — `equals()`/`hashCode()` that collapse instances differing in a payload field — the surviving object changes from run to run, silently violating the stability contract. This is the encounter-order class of #717, but for `distinct()` rather than `skip()`, and unlike `skip`/`dropWhile` (guarded by `222-parallel-reverts-skip.phr`) `distinct` has no guard.

The smallest sound fix mirrors `222`: add a revert/guard so `distinct()` (rules `216`/`307`) stays native and unfused when the pipeline may be parallel, lifting it only when the stream provably stays sequential or is made explicitly `unordered()` upstream. This pairs with sabj25 issue #23, which strengthens the benchmark oracle to detect exactly this identity-breaking rewrite.

yegor256:
Fixes #738.

## Problem

Rule `307-distinct-to-distill` folds `Stream.distinct()` into a distill whose seen-set is a `ConcurrentHashMap.newKeySet()`, captured into the one `BiConsumer` that backs the fused `Stream.mapMulti` dispatch. #715 made that set thread-safe, but a concurrent set only fixes the data race (*whether* duplicates leak) — **not which element survives among equals**.

The JDK's `distinct()` is documented *stable* on an ordered stream: among equal elements it keeps the one first in **encounter order**. A single shared set populated by several `ForkJoin` workers instead keeps whichever equal element wins the race to `add()` — *arrival-order-first*, which is neither encounter-order-first nor deterministic across runs. For primitives and substitutable-`equals` values this is unobservable, but for equal-but-distinguishable objects (an `equals`/`hashCode` that collapses instances differing in a payload field) the surviving object changes from run to run, silently violating the stability contract. This is the encounter-order class of #717, but for `distinct()` rather than `skip()` — and unlike `skip`/`dropWhile` (guarded by `222`/`223`), `distinct` had no guard.

## Fix

Mirror the skip guard for `distinct`, sitting between the `216` recognizer and the `307` fold (2xx sorts before 3xx, so they run before folding):

- **`224-parallel-reverts-distinct.phr`** — parallel call BEFORE the distinct (`source.parallel().…distinct()`)
- **`225-parallel-reverts-distinct-after.phr`** — parallel call AFTER the distinct (`source.distinct().parallel()`)

Once `216` has produced a `Φ.hone.distinct`, if the same method body also contains a `parallel()`/`parallelStream()` invoke, the pragma is reverted to a native `invokeinterface …Stream.distinct:()…` carrying the `reverted ↦ Φ.true` marker (byte-identical to the opcode `441` already emits). The marker gives the opcode a fifth binding, so `216`'s closed four-operand pattern cannot re-recognise it and the `216 → 224` cycle cannot spin. On a parallel pipeline the distinct therefore stays native and unfused, and the JDK's ordered/parallel-correct `distinct()` runs.

Detection is intentionally over-broad (any parallel call in the method reverts every distinct in it); over-reverting only forfeits the fusion optimisation, the native call stays correct.

## Tests

- `src/test/phino/224-…yml`, `src/test/phino/225-…yml` — single-rule packs asserting the revert + `reverted` guard, modeled on the `222`/`223` packs.
- `src/test/resources/org/eolang/hone/optimize/streams/parallel-distinct-ordered.yml` — end-to-end fixture: 1,000,000 `Tag` objects whose `equals`/`hashCode` key on a field but carry a distinguishing `seq` payload, on an ordered parallel stream with a stateless `map` neighbour (so the distinct would otherwise fuse). A stable distinct keeps encounter-order-first (`seq` 0..999), summing to a deterministic **499500**; the arrival-order seen-set produced a wrong, run-varying sum. Verified the unoptimized JDK pipeline produces `499500`.
- README: new "Why `distinct` Is Only Fused on Sequential Pipelines" section.

Note: the `deep`/phino tests require the `phino` binary (or Docker), neither available in the authoring sandbox, so they were not executed here — they will run in CI.

This pairs with sabj25 issue #23 (strengthening the benchmark oracle to detect this identity-breaking rewrite).

https://claude.ai/code/session_01JY1iEcoHudC8vaUsKB31JB

---
_Generated by [Claude Code](https://claude.ai/code/session_01JY1iEcoHudC8vaUsKB31JB)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,30 @@ The JDK's native `Stream.skip(n)` honours the ordered/parallel contract,
 The same revert-on-parallel guard applies verbatim to a future `dropWhile`
   lifting, which is order-dependent for the same reason.
 
+## Why `distinct` Is Only Fused on Sequential Pipelines
+
+`distinct()` is fused (it lowers to a seen-set check inside the `mapMulti`
+  body, see `307-distinct-to-distill`), but only when the pipeline provably
+  stays sequential.
+The fused seen-set is a thread-safe `ConcurrentHashMap.newKeySet()`, which
+  #715 introduced to close the data race on parallel streams; making `add()`
+  atomic, however, only fixes *whether* duplicates leak, not *which* element
+  survives among equals.
+The JDK's `distinct()` is documented *stable* on an ordered stream: among
+  equal elements it keeps the one first in encounter order.
+A single shared set populated by several `ForkJoin` workers instead keeps
+  whichever equal element wins the race to `add()` — arrival-order-first,
+  which is neither encounter-order-first nor deterministic across runs.
+For primitives and substitutable-`equals` values this is unobservable, but
+  for equal-but-distinguishable objects (an `equals`/`hashCode` that collapses
+  instances differing in a payload field) the surviving object changes from
+  run to run, silently violating the stability contract.
+The JDK's native `Stream.distinct()` honours the ordered/parallel contract,
+  so — exactly as for `skip` — when a method contains a `parallel()` or
+  `parallelStream()` call the rules `224-parallel-reverts-distinct` and
+  `225-parallel-reverts-distinct-after` revert the recognised `distinct` back
+  to the native call before `307` can fold it (#738).
+
 ## How to Use in Gradle
 
 You can use this plugin with [Gradle] too, but it requires
--- a/README.md
+++ b/README.md
@@ -351,9 +351,9 @@ The same revert-on-parallel guard applies verbatim to a future `dropWhile`
   stays sequential.
 The fused seen-set is a thread-safe `ConcurrentHashMap.newKeySet()`, which
   #715 introduced to close the data race on parallel streams; making `add()`
-  atomic, however, only fixes *whether* duplicates leak, not *which* element
+  atomic, however, only fixes _whether_ duplicates leak, not _which_ element
   survives among equals.
-The JDK's `distinct()` is documented *stable* on an ordered stream: among
+The JDK's `distinct()` is documented _stable_ on an ordered stream: among
   equal elements it keeps the one first in encounter order.
 A single shared set populated by several `ForkJoin` workers instead keeps
   whichever equal element wins the race to `add()` — arrival-order-first,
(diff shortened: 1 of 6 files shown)


yegor256:
@rultor release, tag is `0.27.5`

yegor256:
@rultor release, tag is `0.27.6`

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis, +15.15 for your 303 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (303 × 0.05, capped at 16), -4 for exceeding 100 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html), and -8 for the lack of code review as outlined in our policy. Next time, consider getting your code reviewed to avoid the -8 penalty and [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the quality high! Your running score is [+1149](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #754, opened 2026-06-30
#741: add PhinoTest covering available() return paths

### LINKED ISSUE #741: `Phino.available()` in src/main/java/org/eolang/hone/Phino.java has no test coverage
`Phino` decides whether `OptimizeMojo`, `PullMojo`, `BuildMojo`, and `RmiMojo` run the local `phino` binary or fall back to Docker, yet `src/test/java/org/eolang/hone/` ships no `PhinoTest.java`. Every other production class in `src/main/java/org/eolang/hone/` has a matching `*Test.java` next to it.

The uncovered logic is in `Phino.available()` at `src/main/java/org/eolang/hone/Phino.java` around lines 23 to 57. The method shells out to `phino --version`, trims the stdout, compares it to the expected version with `String.equals`, and returns `true` only on exact match. The behavior is silently relied on at `OptimizeMojo.java:404`, `PullMojo.java:29`, `BuildMojo.java:59`, and `RmiMojo.java:26`, where a `false` return forces a Docker pull. A regression in the version-string comparison, the exit-code branch, or the `IOException` branch flips every caller from local execution to Docker without any failing test.

Add a `PhinoTest.java` covering the three return paths: matching version returns `true`, mismatched version returns `false`, and `phino` missing or non-zero exit returns `false`. The `phino` invocation can be isolated with the same `DisabledWithoutPhinoCondition` already in place, or by stubbing `Jaxec` behind a small seam.

morphqdd:
## What happens

`Phino` decides whether `OptimizeMojo`, `PullMojo`, `BuildMojo`, and `RmiMojo`
run the local `phino` binary or fall back to Docker, yet
`src/test/java/org/eolang/hone/` ships no `PhinoTest.java`. Every other
production class in `src/main/java/org/eolang/hone/` has a matching
`*Test.java` next to it.

`Phino.available()` shells out to `phino --version`, trims the stdout,
compares it to the expected version with `String.equals`, and returns `true`
only on exact match. This is silently relied on at `OptimizeMojo.java:404`,
`PullMojo.java:29`, `BuildMojo.java:59`, and `RmiMojo.java:26`, where a
`false` return forces a Docker pull. A regression in the version-string
comparison, the exit-code branch, or the `IOException` branch flips every
caller from local execution to Docker without any failing test.

## What should happen

`Phino.available()` should have test coverage for all three return paths:
matching version returns `true`, mismatched version (or non-zero exit) returns
`false`, and a missing executable returns `false`.

## Fix

`Phino.available()` always invokes the hardcoded `"phino"` binary, which
makes it untestable without a real `phino` installation or PATH
manipulation. Added a small constructor seam — `Phino(String executable)` —
so tests can point at a fake executable instead. The no-arg constructor still
delegates to `"phino"`, so none of the four production call sites change
behavior.

Added `PhinoTest.java` with 4 tests, each pointing `Phino` at a throwaway
shell script created in a JUnit `@TempDir`:

- `returnsTrueWhenVersionMatches` — script echoes the expected version, exits 0
- `returnsFalseWhenVersionMismatches` — script echoes a different version, exits 0
- `returnsFalseWhenExecutableExitsNonZero` — script exits 1
- `returnsFalseWhenExecutableIsMissing` — path points at a non-existent file

## Verification

- `mvn test -Dtest=PhinoTest` — 4 tests, 0 failures
- `mvn test` (full suite) — all green
- `mvn qulice:check -Pqulice` — no violations

Closes #741

### DIFF
--- a/src/main/java/org/eolang/hone/Phino.java
+++ b/src/main/java/org/eolang/hone/Phino.java
@@ -16,6 +16,26 @@ import java.io.IOException;
 final class Phino {
 
     /**
+     * The executable to run.
+     */
+    private final String executable;
+
+    /**
+     * Ctor.
+     */
+    Phino() {
+        this("phino");
+    }
+
+    /**
+     * Ctor.
+     * @param executable The executable to run, instead of "phino"
+     */
+    Phino(final String executable) {
+        this.executable = executable;
+    }
+
+    /**
      * Is it available?
      * @param expected This is the expected version
      * @return TRUE if available
@@ -23,7 +43,8 @@ final class Phino {
     boolean available(final String expected) {
         boolean available = false;
         try {
-            final Result result = new Jaxec("phino", "--version").withCheck(false).execUnsafe();
+            final Result result = new Jaxec(this.executable, "--version")
+                .withCheck(false).execUnsafe();
             if (result.code() == 0) {
                 final String version = result.stdout().trim();
                 if (version.equals(expected)) {
--- a/src/test/java/org/eolang/hone/PhinoTest.java
+++ b/src/test/java/org/eolang/hone/PhinoTest.java
@@ -1,0 +1,86 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+package org.eolang.hone;
+
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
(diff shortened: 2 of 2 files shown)


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (109 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+784](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #765, opened 2026-07-22
#764: regression coverage for sabj25 sources/fanout Long unboxing bridge

### LINKED ISSUE #704: VerifyError: filter().boxed() fuses to a distill whose keep-frame says 'long' while the stack holds a 'Long'

### LINKED ISSUE #764: Stream fusion intermittently emits a broken unboxing lambda: NoSuchMethodError / LambdaConversionException at link time on sabj25 sources/fanout
Optimizing `objectionary/sabj25` `Main.sources` and `Main.fanout` with the `-Phone` profile occasionally yields classes that fail at **link time** (not verification). Both are boxed `Stream<Long>` pipelines unboxed through `mapToLong(Long::longValue)` / `flatMapToLong`. One benchmark run produced two distinct symptoms:

- `sources` — `java.lang.NoSuchMethodError: 'long sabj25.Main.lambda$hone$8880(java.lang.Long)'` at `sabj25.Main.sources(Main.java)`: the rewritten `invokedynamic` references a synthetic unboxing lambda `(Long)long` that the pass never emitted.
- `fanout` — `java.lang.BootstrapMethodError` caused by `java.lang.invoke.LambdaConversionException: Invalid receiver type long; not a subtype of implementation type class java.lang.Long` at `sabj25.Main.fanout(Main.java)`: the fused lambda's `invokedynamic` instantiated method type carries a primitive `long` receiver where the metafactory needs the boxed `Long` implementation.

Observed with hone-maven-plugin 0.28.0, JEO 0.15.3, PHINO 0.0.100, JDK 26.0.1 (Temurin). The plugin is at 0.73.1 now, so this is worth confirming on current master.

Crucially it is **non-deterministic**: five independent `-Phone` optimizations of the identical input produced this only once, and the other four emitted correct bytecode that linked and ran fine. Both failures live in the object-to-primitive unboxing-lambda bridge — the same `Long`/`long` family as the closed #704 / #713, but surfacing at link time rather than at verification. The run-to-run variance and the high `lambda$hone$8880` counter point at ordering or naming nondeterminism in the rewrite around `src/main/resources/org/eolang/hone/rules/streams/6xx/603-unbox-to-lambda.phr` and `src/main/resources/org/eolang/hone/rules/streams/1xx/111-invokedynamic-to-lambda.phr`, where a synthetic unboxing lambda is sometimes numbered or wired inconsistently with the `invokedynamic` that calls it — leaving a dangling reference (`sources`) or a receiver-type mismatch (`fanout`).

The smallest fix: make the unbox-lambda bridge deterministic — match rewrite candidates and assign `lambda$hone$N` names in a stable, source-ordered pass so identical input always yields identical output — and align the synthetic lambda's descriptor with the `invokedynamic` instantiated receiver type (both boxed `Long`, or both unboxed with a matching bridge) so `603` / `111` never emit a call to a lambda they did not generate.

yegor256:
Closes #764 (pending maintainer confirmation).

## What the issue reported

On **hone 0.28.0** (JEO 0.15.3, PHINO 0.0.100, JDK 26), optimizing the `objectionary/sabj25` `Main.sources` and `Main.fanout` benchmarks *intermittently* (≈1 in 5 runs on identical input) produced a class that failed at **link time** inside the object-to-primitive unboxing-lambda bridge:

- `sources` — `NoSuchMethodError` for a synthetic `lambda$hone$N:(Ljava/lang/Long;)J` referenced from an `invokedynamic` but never emitted.
- `fanout` — `LambdaConversionException: Invalid receiver type long; not a subtype of implementation type class java.lang.Long`.

The issue asked to *"confirm on current master"* (the plugin is now at 0.73.x, well past 0.28.0 and the related fixes #704/#713/#733).

## What I did

Reproduced the exact toolchain locally (phino `0.0.100` + jeo `0.15.3`, the pinned versions, extracted from `yegor256/hone:latest`) and drove the `sources`/`fanout` shapes through the full `streams/*` pipeline with the production settings (`--max-cycles 1 --max-depth 500`):

- **60+ phino rewrites** and **15+ full disassemble → rewrite → assemble → link → run roundtrips** across the `mapToLong(Long::longValue)` and `flatMapToLong` shapes (lone, fused, and multi-pipeline-per-class variants).
- Every run linked and produced the correct result. Output was **structurally identical up to renaming** — phino guarantees `random-string`/`random-tau` uniqueness *within one execution*, and in production all `streams/*` rules run in a single phino execution, so the hypothesised `lambda$hone$N` numbering collision cannot arise there.

I could **not reproduce** the intermittent link failure on current master. The evidence indicates the 0.28.0 symptom was resolved by the intervening Long/long-bridge fixes. Rather than add a speculative change to the order-sensitive ruleset (which `CLAUDE.md` warns against) with no way to verify it fixes anything, this PR locks the shapes in with a regression fixture — mirroring how #713 landed `maptolong-filter.yml`.

## Change

Adds `src/test/resources/org/eolang/hone/optimize/streams/unbox-fanout.yml`, an end-to-end optimize fixture with both benchmark shapes in one class (`sabj25.X`):

- `sources()` — boxed `Stream` unboxed via a tail `mapToLong(Long::longValue)` after `filter`+`map`; fuses to one `Stream.mapMultiToLong`.
- `fanout()` — `Stream.flatMapToLong` fan-out after `filter`; `457` rewrites it to a `mapMultiToLong` backed by a synthesised `flatadapt` `BiConsumer`.

Assertions:
- `BUILD SUCCESS` and `The result is 15 154` — the values are order- and count-sensitive, so a miswired unboxing bridge prints a different number even if the class happens to link.
- `invokedynamic` **5 → 3** — the five javac factories collapse as both `mapToLong` and `flatMapToLong` fuse into single `mapMultiToLong` dispatches.

Verified locally: `before` = 5 on the javac class, `after` = 3 on the jeo-assembled optimized class, stable across repeated runs.

## Notes for the reviewer

If the intermittent failure still occurs in your environment, this fixture is the reproduction target — a failing run will show the wrong printed result or a link error. I did not touch any `.phr` rule, since I have no reproducing case to validate a rule change against and every roundtrip on current master was correct.

https://claude.ai/code/session_01TgkdZciY3Wsj2PC6rJTXDY

---
_Generated by [Claude Code](https://claude.ai/code/session_01TgkdZciY3Wsj2PC6rJTXDY)_


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2819](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #778, opened 2026-08-07
Give every name-minting rule its own namespace

### LINKED ISSUE #777: Regression of #764: PHINO 0.0.106 emits a broken long/Long unboxing lambda, LambdaConversionException links-fail all sabj25 benchmarks
Fusing `objectionary/sabj25` with the `-Phone` profile on hone-maven-plugin **0.29.1** (Docker image `yegor256/hone:0.29.1`, which bundles **PHINO 0.0.106**) produces an unboxing lambda whose `invokedynamic` instantiated receiver type is primitive `long` while the implementation method takes boxed `java.lang.Long`. The synthetic bridge is built in `src/main/resources/org/eolang/hone/rules/streams/6xx/603-unbox-to-lambda.phr` (the `lambda-signature` is assembled at lines 60–62 as `(Ljava/lang/Object;)<primitive>`) and paired with the `invokedynamic` captured in `.../1xx/111-invokedynamic-to-lambda.phr` (instantiated type at `𝜏12` vs bridge signature at `𝜏21`, lines 46–64). At class-load the JVM rejects the mismatch.

Every honed benchmark fails at **link time** in `sabj25.Main.<init>` while linking the fused lambda, so the shared JMH `@State` object never constructs and all 12 benchmarks report `Iteration 1: <failure>`; the honed `results.csv` comes out header-only. The exception:

```
java.lang.BootstrapMethodError: bootstrap method initialization exception
  at java.base/java.lang.invoke.BootstrapMethodInvoker.invoke(BootstrapMethodInvoker.java:187)
  ...
  at sabj25.Main.<init>(Main.java)
Caused by: java.lang.invoke.LambdaConversionException: Invalid receiver type long; not a subtype of implementation type class java.lang.Long
  at java.base/java.lang.invoke.AbstractValidatingLambdaMetafactory.validateMetafactoryArgs(AbstractValidatingLambdaMetafactory.java:270)
  at java.base/java.lang.invoke.LambdaMetafactory.metafactory(LambdaMetafactory.java:356)
```

This is a regression of the closed #764 — same `sources`/`fanout` pipelines, same object-to-primitive unbox-lambda bridge, same `long`/`Long` family as the closed #704 and #649. #764 was closed 2026-07-22, the day hone 0.29.0 / PHINO 0.0.105 shipped, and that pairing still works here (12 valid honed results, `verified()` checksums pass). PHINO 0.0.106 (hone 0.29.1, published 2026-07-24; `latest` shares its digest) brings it back — but now **deterministically and across all 12 benchmarks**, not the intermittent 1-in-5 on two benchmarks that #764 observed. Reproduce by building `tables/sabj25-jvms` from the sabj25 paper harness with `HONE_VERSION=0.29.1 PHINO_VERSION=0.0.106 JEO_VERSION=0.15.3` on JDK 26.0.1 (Zulu) and running the JMH jar; `HONE_VERSION=0.29.0 PHINO_VERSION=0.0.105` passes.

The fix is to restore #764's alignment so the synthetic lambda descriptor and the `invokedynamic` instantiated receiver agree — both boxed `Long` (an `Object`/`Long` receiver), or both unboxed through a matching bridge — in `603-unbox-to-lambda.phr` / `111-invokedynamic-to-lambda.phr`, so the metafactory never sees a primitive `long` receiver against a `Long` implementation. Since the only component that changed between the working and broken runs is PHINO (0.0.105 → 0.0.106), the regression likely lives in phino's rewriting of these rules and may need a fix in `objectionary/phino` and then a republished hone image (phino 0.0.107–0.0.109 exist on Maven Central, but no hone Docker image bundles them yet — `latest` still ships 0.0.106).

yegor256:
Sixteen `.phr` rules mint synthetic names through phino's `random-string`, and until now they shared four format patterns between them: `lambda$hone$%d` across 103/104/105/107, `L%d` across 304..314 and 455..458, `flatadapt_%d` across 455..458, and `distill_%d` across 501/502. Each pattern now carries the number of the rule that owns it, so two rules can no longer draw the same name. `RulesTest` walks the shipped rule set and fails if a pattern ever gets a second minter again.

The sharing was harmless while phino drew from process entropy, but 0.0.106 seeds its generator identically on every run, so every phino process now mints `lambda$hone$8152` first. In small-steps mode each rule gets its own phino process, so 103 and 104 both hand out that name and the class ends up with two methods called `lambda$hone$8152`. Rule 711 then reverts an invokedynamic through whichever wrapper it matches first — on the sabj25 benchmarks it repointed the `Long::toString` call site in `Main.<init>` at `Long.longValue`, and the class died at link time with `Invalid receiver type long; not a subtype of implementation type class java.lang.Long`. That is the failure in #777, and after this change the same build links and runs.

One thing worth flagging: I could not reproduce #777 in normal mode. On hone 0.29.1 with phino 0.0.106 and jeo 0.15.3 on JDK 26.0.1, sabj25's `Main` optimizes and constructs fine both through host phino and through the `yegor256/hone:0.29.1` image — normal mode runs one phino process per class, and phino does keep names unique inside a process. Only `-Dhone.small-steps=true` breaks, deterministically. So 603 and 111 turned out to be innocent, and nothing needs fixing in phino.

Closes #777

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -35,6 +35,9 @@ rule, how to test, and what to avoid*.
    subdirectory whose digit matches the rule's hundreds prefix). Start from
    the closest existing rule and keep its header block (`SPDX` +
    `# yamllint disable rule:line-length`).
+   If the rule mints a synthetic name with `random-string`, embed its own
+   `NNN` in the format pattern (`distill_501%d`, not `distill_%d`), so that no
+   two rules can ever draw the same name; `RulesTest` enforces this.
 3. **Run small-steps locally** on a representative `.phi` to see what your
    rule produces:
 
@@ -128,6 +131,13 @@ every `@Tag("deep")` test runs against the real `phino` binary on the host.
   unfolded pragma at `350-` would silently break fusion on that method.
 - Do not change `Collections.sort(names)` in `Rules.discover()`. The pipeline
   depends on alphabetical ordering as its scheduling mechanism.
+- Do not let two rules share one `random-string` pattern. Phino guarantees
+  minted names are unique only inside a single process, and it seeds its
+  generator identically on every run since 0.0.106, so in small-steps mode —
+  one process per rule — both rules draw the very same name and produce two
+  methods with one name in one class. `711` then reverts an invokedynamic
+  through the wrong wrapper and the class dies at link time with
+  `LambdaConversionException` (see #777).
 - Do not hand-edit `.class` files in `target/classes-before-hone/`. That
   directory is the backup the plugin makes before mutation; it is the only way
   to recover the pre-optimization bytecode without recompiling.
--- a/src/main/resources/org/eolang/hone/rules/streams/1xx/103-methodref-to-lambda.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/1xx/103-methodref-to-lambda.phr
@@ -245,7 +245,7 @@ where:
     function: random-tau
   - meta: 𝑒-wrapper-name
     function: random-string
-    args: ['"lambda$hone$%d"']
+    args: ['"lambda$hone$103%d"']
   - meta: 𝑒-return-op
     function: sed
     args:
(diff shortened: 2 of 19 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (151 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2766](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #789, opened 2026-08-11
#788: narrow the item through a filter behind distinct()/skip()

### LINKED ISSUE #788: VerifyError: 505 does not narrow the Object item when a filter follows distinct()/skip()
`src/main/resources/org/eolang/hone/rules/streams/5xx/505-distill-state-item-cast.phr:52-64`
anchors the `checkcast` it splices on a `Φ.jeo.frame` *immediately* followed by an
`invokestatic`. When a `filter(...)` rides the pipeline that a `distinct()` or a
`skip()` led, that adjacency never holds: `431-dup-before-filter-distill.phr:30` has
already prepended a `dup` to the filter distill's body, so 505 does not match, no cast
is spliced, and the erased `Object` item reaches the filter's desugared predicate —
typed `(Ljava/lang/String;)Z` for a `String` element. The verifier rejects the class
with `Bad type on operand stack: Type 'java/lang/Object' is not assignable to
'java/lang/String'`. Two shapes that reach it:
`Stream.of("alpha", "beta", "alpha").takeWhile(s -> s.length() < 12).distinct().filter(s -> !s.isEmpty()).map(String::toUpperCase).findFirst()`
and
`DoubleStream.of(1.5, 2.25, 1.5).skip(1L).mapToObj(Double::toString).distinct().filter(s -> s.length() > 3).count()`,
with hone 0.29.3, JEO 0.15.3, PHINO 0.0.106 on JDK 25.

505's own header names the gap at :37-42 — "A filter or peek wedged between the guard
and the boxing tail opens its body with a dup, breaking the frame → invokestatic
adjacency" — and leaves it as a follow-up puzzle, but nothing tracks it and the shape is
not exotic: `distinct().filter(...)` is ordinary code, and the item needs narrowing
there for exactly the reason #649 needed it before a boxing tail. The guard above is
untouched by this: `Set.add(Object)` and the skip counter are happy with an `Object`,
only the first typed consumer is not.

The smallest fix keeps 505's splice point — immediately after the frame, where the frame
may still rightly declare `Object` — and only lets the pattern tolerate 431's leading
`dup` before the typed invoke: add the variant in which
`𝜏-dup ↦ ⟦ φ ↦ Φ.jeo.opcode.dup ⟧` sits between `𝜏-frame` and `𝜏-invoke`, so both
copies of the narrowed item are typed. The `when` guard on a single non-`Object`
reference parameter can stay as it is.

yegor256:
Closes #788.

## The gap

`505-distill-state-item-cast` anchors the narrowing `checkcast` on a `Φ.jeo.frame` **immediately** followed by an `invokestatic`. When a `filter(...)` rides a pipeline that `distinct()` or `skip()` leads, that adjacency never holds: `283-insert-dup-before-filter-after-stateful` (or `431-dup-before-filter-distill` after a fuse) has already put a `dup` in front of the predicate, so the joined body reads `frame → dup → invokestatic`. 505 does not match, no cast is spliced, and the erased `Object` item reaches the desugared predicate — typed `(Ljava/lang/String;)Z` for a `String` element — so the verifier rejects the class with `Bad type on operand stack: Type 'java/lang/Object' is not assignable to 'java/lang/String'`.

## The fix

A new rule `508-distill-state-item-cast-dup` — the dup-tolerant twin of 505 — matches exactly that interleaving and keeps 505's splice point: the cast goes right after the frame, **before** the dup, so both copies of the item are narrowed, the one the predicate consumes and the one that survives on the keep path. The frame is untouched (it correctly still declares `Object`, since the cast runs after it), so 503 is unchanged.

Guards are 505's, unchanged: `bridge-input` must be `Ljava/lang/Object;` (the distinct/skip element-erasure signature) and the invoke must take a single non-`Object` reference argument, which leaves an `Object`-accepting predicate alone and skips a captured `lambda$` (it takes the captured values plus the item). Only the one-slot `dup` is tolerated — the two-slot `dup2` that 283 picks for a primitive item cannot occur under an `Object` `bridge-input`. A `peek` behind a stateful guard has the same `dup` + invoke opening (`309-peek-to-distill`), so it is covered by the same pattern.

`508` sits after `503` (which turns the keep label's `frame-item` into a `Φ.jeo.frame`) and before `511`/`512` (which wrap the body into a method). At fixpoint the inserted checkcast sits between the frame and the dup, so neither 508 nor 505 re-matches the same site.

## Changes

- `src/main/resources/org/eolang/hone/rules/streams/5xx/508-distill-state-item-cast-dup.phr` — the new rule.
- `src/main/resources/org/eolang/hone/rules/streams/5xx/505-distill-state-item-cast.phr` — header only: the "Limitation" paragraph that left this as an open puzzle now points at 508.
- `src/test/phino/508-distill-state-item-cast-dup.yml` — single-rule pack: `503` + `508` over a body shaped `fetch, frame-item, dup, invokestatic (Ljava/lang/String;)Z`, asserting the `checkcast java/lang/String` lands between the frame and the dup.
- `src/test/resources/org/eolang/hone/optimize/streams/stateful-filter-tail.yml` — end-to-end fixture with both shapes the issue reports (the `takeWhile().distinct().filter().map().findFirst()` one and the `skip(1).mapToObj().distinct().filter().count()` one) in a single class; it asserts the optimized class verifies and still prints `The result is ALPHA/1`. `after` opcode counts are deliberately left unasserted — the point of the pack is that the class verifies and computes the right answer.

Note that `full-non-terminal.yml` already chains `.distinct().skip(0L).filter(...)` and passes today: it leads with a typed `filter`/`map`, which pins `bridge-input` to the wrapper, so no narrowing is needed. That is why the shape stayed latent, and why the new fixture leads with the stateful operator.

## Verification — please read

`mvn test` (non-deep) is green locally: 47 tests, `RulesTest` included. **The two new tests have not been executed anywhere yet**, and this PR's checks will not execute them either:

- Both live under the `deep` profile, and `.github/workflows/deep.yml` triggers only on `push` to `master` — not on pull requests.
- The sandbox this was written in has neither a `phino` binary nor a working Docker daemon (and the egress policy blocks the phino release downloads), so `mvn -Pdeep test` could not be run here at all.

So the rule is reasoned from the pipeline (503 → the fused body's `frame → dup → invokestatic`, mirroring 505's shape exactly) rather than observed firing. Worth a local `mvn -Pdeep test` — or at least `phino rewrite --rule .../508-distill-state-item-cast-dup.phr` on a `.phi` from the new fixture — before merging. Keeping it as a draft for that reason.

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/5xx/505-distill-state-item-cast.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/5xx/505-distill-state-item-cast.phr
@@ -34,12 +34,13 @@
 # method), applied at fixpoint — a checkcast it already inserted now sits
 # between the frame and the invoke, so the adjacency does not re-match.
 #
-# Limitation: only the FIRST typed op directly after the (last) guard is
-# reached, which is exactly the boxed()/mapToX-after-distinct/skip shape #649
-# reports. A filter or peek wedged between the guard and the boxing tail opens
-# its body with a dup, breaking the frame → invokestatic adjacency; narrowing
-# through that interleaving is a follow-up puzzle on the same shared-state
-# channel.
+# Scope: only the FIRST typed op directly after the (last) guard is reached,
+# which is exactly the boxed()/mapToX-after-distinct/skip shape #649 reports. A
+# filter wedged between the guard and the typed tail opens its body with a dup
+# (spliced by 283, or by 431 after a fuse), breaking the frame → invokestatic
+# adjacency this rule anchors on; 508 is the twin that matches that
+# interleaving and casts in front of the dup, so both copies of the item are
+# narrowed (#788).
 name: distill-state-item-cast
 pattern: |
   ⟦
--- a/src/main/resources/org/eolang/hone/rules/streams/5xx/508-distill-state-item-cast-dup.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/5xx/508-distill-state-item-cast-dup.phr
@@ -1,0 +1,101 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# The dup-tolerant twin of 505, closing the follow-up its header leaves open
+# (#788). 505 narrows the erased Object item that a distinct()/skip()-led
+# pipeline carries, by splicing a `checkcast` between the stateful guard's
+# keep-frame and the first body opcode that consumes the item as a NARROWER
+# reference; it anchors on a Φ.jeo.frame IMMEDIATELY followed by an
+# invokestatic. A filter riding the same pipeline breaks that adjacency: 283
+# splices a Φ.hone.dup in front of a filter that follows a stateful guard (431
+# does the same after a fuse), because the predicate eats the item and the keep
+# path needs a copy — and a peek, whose body 309 shapes as `dup` + the consumer
+# invocation, arrives with the very same opening. The joined body therefore
+# reads frame → dup → invokestatic, 505 does not match, no cast is spliced, and
+# the Object item reaches the desugared predicate — typed (Ljava/lang/String;)Z
+# for a String element. The verifier rejects the class with "Bad type on
+# operand stack: Type 'java/lang/Object' is not assignable to
+# 'java/lang/String'". The guard above is untouched by this: Set.add(Object)
+# and the skip counter are happy with an Object, only the first typed consumer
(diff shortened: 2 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (224 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2802](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #792, opened 2026-08-11
fix(#787): bump the caller max-stack in 502 for the spliced state prologue

### LINKED ISSUE #787: VerifyError: 502 splices the state prologue into the caller but never bumps its max-stack
`src/main/resources/org/eolang/hone/rules/streams/5xx/502-distill-state-to-mapMulti.phr:74-84`
splices `new java/util/ArrayList`, `dup`, `invokespecial ArrayList.<init>` and the
relocated `state` opcodes into the caller's body, ahead of the fused dispatch, while
the caller's `Φ.jeo.maxs` formation travels untouched through `𝐵-method-head`
(pattern at :33, result at :71). The declared `max_stack` therefore stays at whatever
`javac` sized for the unfused `invokeinterface` chain, and the spliced prologue needs
more than that: with the source value already on the stack, `new`/`dup` peaks two cells
above it, and the append that 307 contributes (`dup`, `invokestatic
ConcurrentHashMap.newKeySet`, `List.add`, `pop`) peaks three. The verifier then rejects
the class at load with `Exceeded max stack size`. A shape as ordinary as
`list.stream().distinct().map(n -> n + 1L).peek(...).mapToInt(Long::intValue).distinct().map(n -> n + 3).toArray()`
is enough to provoke it, with hone 0.29.3, JEO 0.15.3, PHINO 0.0.106 on JDK 25.

Nothing downstream repairs the number. `112-capturing-invokedynamic-to-map.phr:44-48`
states the contract in the repo's own words — "jeo never recomputes maxs, so this rule
BUMPS the enclosing method's max-stack by 2" — and jeo agrees: `BytecodeMaxs.compute()`
returns true only when both maxs are `-1`, which never holds for a method disassembled
from an existing class, so `BytecodeMethod.write` re-emits the stale value verbatim.
The `max-stack` that 307 and 308 talk about is only their own bounded peak, in a
comment; a bounded peak is not a declared one. This is the closed #606 reappearing in
the rule that replaced `502b-cps-one-cap-to-lambda.phr`.

The smallest fix is 112's, applied to 502: split the caller's
`maxs ↦ ⟦ φ ↦ Φ.jeo.maxs, m1 ↦ 𝑒-max-stack, m2 ↦ 𝑒-max-locals ⟧` out of
`𝐵-method-head` in both the pattern and the result, and bind `m1` to a `where` entry
`function: sum, args: [𝑒-max-stack, 6]`. Six covers the deepest state append (308's
`long[1]`, two slots per push), and over-allocating `max_stack` is always
verifier-safe.

yegor256:
Closes #787.

`502-distill-state-to-mapMulti.phr` splices `new java/util/ArrayList`, `dup`, `invokespecial <init>` and the relocated `state` opcodes into the caller ahead of the fused dispatch, while the caller's `Φ.jeo.maxs` formation travelled untouched through `𝐵-method-head`. The declared `max_stack` stayed at whatever `javac` sized for the unfused `invokeinterface` chain, and nothing downstream repairs it — `BytecodeMaxs.compute()` returns true only when both maxs are `-1`, never the case for a method disassembled from an existing class, so `BytecodeMethod.write` re-emits the stale number verbatim. The class then dies at load with `VerifyError: Operand stack overflow — Exceeded max stack size`.

## The fix

112's, applied to 502: split `maxs` out of `𝐵-method-head` in both the pattern and the result, and bind `m1` to `sum(𝑒-max-stack, 6)`.

Six is the peak of the deepest state append, 308's: from `[.. list]` it goes `iconst_1`, `newarray`, `dup`, `iconst_0`, then the two-slot count push before `lastore` collapses it back. 307's peaks at three and the leading `new`/`dup` pair at two, so one bump covers them all however many operators 401 fused into the distill — every append starts and ends at `[.. list]`. Over-allocating `max_stack` is always verifier-safe.

## Tests

* `src/test/phino/502-distill-state-to-mapMulti.yml` — single-rule pack asserting the 2 → 8 bump, the spliced prologue and the relocated append.
* `src/test/resources/org/eolang/hone/optimize/streams/maxstack.yml` — end-to-end. Its pipeline lives in its own method, so `javac` declares a `max_stack` of just 2; every existing stateful fixture hides the bug behind the roomier declaration that building a `Stream.of(...)` varargs array forces, which is why 38 e2e packs never caught this.

Verified against the pinned phino 0.0.106 (lifted out of `yegor256/hone:latest`, so no Docker fallback was needed):

* all 64 single-rule packs pass, including the new one;
* the new e2e fixture passes with the fix, and reverting only the `.phr` reproduces `java.lang.VerifyError: Operand stack overflow — Exceeded max stack size` — so it is a genuine regression test for #787;
* the rest of `mvn -Pdeep test` is unchanged. Two e2e packs (`all-ops.yml`, `stateful-ops.yml`) and five Docker-image-building tests do fail in the sandbox used to verify this, but the two packs emit byte-identical opcode tallies with and without the change, and the five build failures are `wget` inside the image build not reaching `archive.apache.org` through the sandbox proxy. Worth a look at the deep job on merge to confirm those two packs are green on the real runner.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PEfFHcZRfSPieSyThjzT1j)_

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/5xx/502-distill-state-to-mapMulti.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/5xx/502-distill-state-to-mapMulti.phr
@@ -17,6 +17,26 @@
 # It fires only when the state block carries at least one opcode (the
 # leading 𝜏-state-first binding), which is exactly what keeps 501 —
 # matching an empty state — on the stateless path.
+#
+# The spliced prologue is DEEPER than the dispatch it replaces, so the caller's
+# declared max-stack has to grow with it (issue #787). javac sized max_stack for
+# the unfused invokeinterface chain — usually just [stream, function] — while the
+# prologue peaks well above that: with the source stream already on the stack the
+# `new`/`dup` pair alone reaches two cells above it, 307's append (dup;
+# newKeySet; List.add; pop) three, and 308's (the long[1] seed, two slots per
+# push) six. jeo never recomputes maxs — BytecodeMaxs.compute() gives up unless
+# BOTH maxs are -1, which never holds for a method disassembled from an existing
+# class, so BytecodeMethod.write re-emits whatever number was there — and no
+# later rule repairs it either; the "bounded peak" 307 and 308 talk about is
+# their own, not a DECLARED one. Left alone the class dies at load with
+# "Exceeded max stack size". So, exactly as 112 does for its relocated boxing
+# append, this rule splits `maxs` out of 𝐵-method-head and BUMPS max-stack by 6:
+# six covers the deepest state append (308's), the peak of every append is
+# measured from the same [.. stream, list] base no matter how many operators 401
+# fused, and over-allocating max_stack is always verifier-safe. jeo names the two
+# maxs bindings "m1" (max-stack) and "m2" (max-locals), so they are matched by
+# those literal names — a 𝜏- metavariable used as a binding NAME no longer binds
+# to a scalar value (see 112).
 name: distill-state-to-mapMulti
 pattern: |
   ⟦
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (163 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2806](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #790, opened 2026-08-11
VerifyError: a filter fused behind a peek never gets its guard dup


yegor256:
A `peek` immediately followed by a `filter` fuses into a distill whose body
consumes the element twice but duplicates it once. For
`Stream.of(5L, 3L, 6L, 11L).peek(n -> SUM[0] += n).filter(n -> n % 2L == 0L).reduce(0L, Long::sum)`
the synthesised `distill_501N(Ljava/lang/Long;Ljava/util/function/Consumer;)V` comes out as
`aload_0; dup; invokestatic lambda$pipe$0:(Ljava/lang/Long;)V; invokestatic
lambda$pipe$1:(Ljava/lang/Long;)Z; ifne 12; return; 12: aload_1; dup_x1; pop;
Consumer.accept(Object); return`, with a single `same_locals_1_stack_item` frame at @12
declaring `java/lang/Long`. The peek consumer returns void and swallows the duplicate, so
the predicate consumes the element itself and the keep label is reached with an empty
stack, while the frame promises the element. The class does not load:
`java.lang.VerifyError: Inconsistent stackmap frames at branch target 12 ... Reason:
Current frame's stack size doesn't match stackmap` (hone 0.29.3, JEO 0.15.3, PHINO
0.0.106, JDK 25).

The missing `dup` is the filter's own guard, `431-dup-before-filter-distill.phr:31`, whose
pattern only matches a distill with `start ↦ "filter"` (:12). But `401-fuse.phr:43` keeps
the *head's* start on the fused distill, so a filter that 401 has already folded behind a
peek carries `start ↦ "peek"` and 431 can never see it again. The only `dup` left in the
body is the one `309-peek-to-distill.phr:43-45` emits for the peek, and the consumer
spends it. Replacing the peek with a map hides the defect —
`.map(n -> n + 1L).filter(...)` yields `invokestatic; dup; invokestatic; ifne` and runs —
because a map over boxed elements becomes a fusable distill only in `4xx`, after 431 has
already run, so the schedule saves that shape by accident and every fixture with a filter
in it happens to be of that shape.

The smallest fix is to keep 401 off a filter whose guard is not in place yet: add
`when: not: matches: ['^filter$', 𝑒-second-start]` to `401-fuse.phr`, so 431 fires first
and the following cycle fuses a body that already carries its `dup`. Only the second
position needs the guard — a filter in the first position keeps `start ↦ "filter"` after
[...]

PR #793, opened 2026-08-11
Diff the rewrite against randomly generated stream pipelines

### LINKED ISSUE #791: No randomized differential test: every stream fixture is a pipeline we thought of
Every end-to-end fixture under `src/test/resources/org/eolang/hone/optimize/streams/`
is a pipeline someone wrote on purpose, and each one pins the shape it was written to
exercise. That makes the suite good evidence about the shapes we thought of and no
evidence at all about the ones we did not: #649, #704, #713 and #787 were all found in
somebody else's code, not here, and each of them is an ordinary pipeline that no fixture
happened to spell.

What is missing is a generator: walk a typed grammar of the Stream API, emit one pipeline
per class with literal fixtures and a `main` that prints the value and a `peek`
accumulator, compile it with `javac` (the first oracle — if it does not compile the
grammar is wrong), run it, optimize it, run it again, and compare what the two runs
printed. One pipeline per class matters, because the JVM verifies a class as a whole and
a corrupt rewrite would otherwise take down every pipeline sharing the file. Seeds make
it reproducible, so a failure found once can be looked at again.

yegor256:
Every fixture under `optimize/streams/` is a pipeline somebody wrote on purpose, so the
suite is evidence about the shapes we thought of and says nothing about the rest. This
adds `RandomPipeline`, which walks a typed grammar of the Stream API — 226 productions
across seven element domains — and emits one pipeline per class, seeded by its index, plus
`OptimizeMojoTest#preservesWhatRandomPipelinesPrint`, which compiles them, runs them,
optimizes them, runs them again, and compares what the two runs printed. Both lines
matter: the value the pipeline produced, and the number its `peek` accumulated, which is
the traversal contract. A seed is the whole reproduction — `new RandomPipeline(18L)
.java("random", "P0018")` prints the exact class that failed, ready to paste into a `.yml`
fixture.

It walks 120 pipelines by default, about three minutes on a warm image;
`-Dhone.random.pipelines=240` costs five and also passes. Five fast tests guard the
generator itself: same seed same class, different seeds different classes, no `peek` in
front of `count()`, every generated class compiles, and every production the grammar
promises is reachable within 500 seeds. The differential carries `@Tag("deep")`, so it
runs under `-Pdeep` and on pushes to master, not on this pull request.

The grammar deliberately does not step around the defects we know about — every shape
behind #787, #788, #790, #794, #798, #799 and #801 is still walked, since a generator that
avoids known defects only proves that we know about them. Two shapes are quarantined in
`RandomPipeline.quarantined()`, one method each, named by its issue: #804, where a
`filter` behind a `dropWhile` loses its guard dup, and #805, where a conversion into an
object domain and a stateful guard with a `peek`, a `filter` or another guard between them
leaves the keep-label frame naming the wrong type. A walk that reaches one is re-rolled
from a derived seed, so the suite is a regression net rather than a nightly failure, and
closing an issue widens the walk by deleting one method — which is exactly what the last
commit here does for #801.

The #805 clause is as narrow as I could measure it. With the guard adjacent to the
conversion, with a `map` or `mapMulti` in between, with a `sorted`, `limit`, `takeWhile` or
`flatMap` barrier in between, or with a primitive conversion, the rewrite is correct today,
so all of those stay under test.

Closes #791


yegor256:
The lint jobs are fixed: jtcop objects to static fields in a test class and to a test
class without a production counterpart, so the pipeline count is a local now and the four
checks of the generator moved into `OptimizeMojoTest` next to the test that uses them.

What is left is the one rewriter defect this test exists to find, now reduced and filed as
#794: a type-changing object map fused ahead of an object-predicate filter leaves the
keep-label frame naming the type the distill received. `Stream.of(5L, 3L, 6L)
.map(Object::toString).filter(s -> !s.isEmpty()).count()` is enough to trigger it. Fix
that one and the differential test should go green at the default two dozen seeds.

yegor256:
#794 is fixed on the branch and the differential test now passes at its default two dozen
seeds, 74 seconds on a warm image. The new rule was checked on its own first — running
506, 507 and 509 over the #794 distill by hand gives `item ↦ "java/lang/String"`, and the
three older frame packs still resolve to `java/lang/Long`, `long` and `java/lang/Integer`
with 507 in the chain, so it steals nothing from them.

One note on the local run of the whole `-Pdeep` suite: 110 tests, two failures, both
`after`-count mismatches in `all-ops` and `stateful-ops`. They are not from this branch —
checked out `master` in a second worktree and they fail there with byte-identical actual
counts. The counts drift on JDK 25, which is what this machine runs; the deep job uses 21.
The single-rule packs skip here too, since the local phino is 0.0.109 against the 0.0.106
pin, which is why they were verified by hand.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (986 >= 100); -8 for way too many hits-of-code (986 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2846](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #796, opened 2026-08-11
fix(#794): frame an object-predicate filter from its predicate

### LINKED ISSUE #794: VerifyError: a type-changing map fused ahead of a filter leaves the keep-frame naming the old element type
An object map that changes the element type, fused in front of an object-predicate
filter, produces a distill whose keep-label frame still names the type the distill
received. For `Stream.of(5L, 3L, 6L).map(Object::toString).filter(s -> !s.isEmpty())
.count()` the synthesised `distill_501N(Ljava/lang/Long;Ljava/util/function/Consumer;)V`
comes out as `aload_0; invokestatic lambda$hone$N:(Ljava/lang/Long;)Ljava/lang/String;;
dup; invokestatic lambda$main$0:(Ljava/lang/String;)Z; ifne 12; return; 12: <frame
same_locals_1_stack_item java/lang/Long>; aload_1; dup_x1; pop;
Consumer.accept(Object); return`. The guard `dup` is in place and the stack at the keep
label holds the `String` the map produced, but the frame promises a `Long`, so the class
does not load: `java.lang.VerifyError: Inconsistent stackmap frames at branch target 12
... Reason: Type 'java/lang/String' (current frame, stack[0]) is not assignable to
'java/lang/Long' (stack map, stack[0])` (hone at 66948136, JEO 0.15.3, PHINO 0.0.106,
JDK 25).

The frame comes from `507-distill-filter-frame-fallback.phr:59-72`, which resolves the
`Φ.hone.frame-item` marker from the distill's `bridge-input` unconditionally. That is the
right type whenever the filter's survivor is the element the distill received, which its
header comment claims is the common case, but a map fused ahead of the filter breaks the
identity: the survivor is the value the `dup` copied, so its type is the predicate's own
parameter type. `506-distill-filter-frame.phr` already knows bridge-input is not always
right, yet its guard at `:98-100` restricts it to predicates with a primitive signature
(the #704 and #713 shapes), so an object predicate such as `(Ljava/lang/String;)Z` falls
through to 507. Keeping the element type identical hides it — `.map(n -> n + 1L)
.filter(n -> n > 0L)` in the same position runs — which is why every existing fixture
passes.

The smallest fix is to resolve the marker from the predicate's parameter type rather than
from `bridge-input` when the predicate is an object predicate: match the
`invokestatic`/`ifne`/keep-label/marker run the way 506 does, take `𝑒-frame-item` from
the `(L...;)Z` argument, and leave the bridge-input fallback for markers with no
predicate in front of them. Since the specific rule has to fire before the general one,
it needs a number below the fallback's.

Found by the randomized differential test in #793, which reduces to the snippet above.

yegor256:
Closes #794.

## The bug

A type-changing object map fused ahead of an object-predicate filter left the guard's keep-label frame naming the type the distill *received* rather than the one on the stack. For `Stream.of(5L, 3L, 6L).map(Object::toString).filter(s -> !s.isEmpty()).count()` the synthesised `distill_501N(Ljava/lang/Long;Ljava/util/function/Consumer;)V` promised a `java/lang/Long` over the `java/lang/String` the map had produced, and the class did not load:

```
  Location: distill_501918(Ljava/lang/Long;Ljava/util/function/Consumer;)V @12: aload_1
  Reason: Type 'java/lang/String' (current frame, stack[0]) is not assignable to
          'java/lang/Long' (stack map, stack[0])
```

The `Φ.hone.frame-item` marker 304 leaves at the keep-label fell through to `507-distill-filter-frame-fallback`, which resolves it from the distill's `bridge-input` unconditionally; `506-distill-filter-frame` could not claim it because its guard admits only primitive predicates.

## The fix

`506-distill-predicate-frame` matches the same `invokestatic`/`ifne`/`return`/keep-label/marker run as 506, guards on an object predicate signature `(L…;)Z`, and takes the frame item from the predicate's own parameter type. That names the value the guard's `dup` copied whether or not a map was fused ahead — while the filter sits at the head of a distill the predicate parameter *is* `bridge-input` (201/232 derive both from the same instantiated SAM type, and 103/104/107 give their wrappers that signature), and 401 is what breaks the identity. Markers with no predicate in front of them still land on 507 and its `bridge-input`.

Alphabetical filename order is the schedule and no integer is free between 506 and 507, so the rule shares 506's number: `506-distill-filter-frame` < `506-distill-predicate-frame` < `507-distill-filter-frame-fallback`. The two 506 guards are disjoint (primitive vs object predicate), so only the position relative to the fallback matters. Header comments on 506 and 507 and the note in `CLAUDE.md` were updated to say so.

## Tests

* `src/test/phino/506-distill-predicate-frame.yml` — single-rule pack over the #794 shape. With the rule the marker resolves to `java/lang/String`; with only 506 + 507 it resolves to `java/lang/Long`.
* `src/test/resources/org/eolang/hone/optimize/streams/map-retype-filter.yml` — end-to-end fixture with both retyping shapes (an `Object::toString` methodref map and an `Integer -> String` lambda map) in front of object-predicate filters. Each pipeline collapses map + filter into one `Stream.mapMulti`, invokedynamic 4 -> 2.
* `src/test/phino/507-distill-filter-frame-fallback.yml` — comment updated: the pack still pins the fallback's own derivation, and the boxed-roundtrip marker it covers resolves to the same `java/lang/Integer` through the new rule in the full pipeline.

## Verification

Run locally with phino 0.0.106 on the host (no Docker):

* Reproduced first: with the new rule removed, `mvn process-classes` on the fixture's program fails with exactly the `VerifyError` above; with the rule it prints `The result is 3/2` and the keep-frame reads `stack = [ class java/lang/String ]`.
* `mvn test` — 47/47 green.
* `mvn -Pdeep test -Dtest='OptimizeMojoTest#appliesPhinoRulesAsSpecifiedInYamlPack'` — 66/66 packs green.
* End-to-end fixtures: `all-ops` and `stateful-ops` fail their `after` opcode counts in this sandbox, and they do so independently of this change — running the whole rule set over their `.phi` with and without the new rule produces byte-identical output, and both programs still print the right answer. Every other fixture, including the new one, is green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JxhGd6or215fCCHVeUQL6c)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -30,6 +30,10 @@ rule, how to test, and what to avoid*.
    places it correctly. Execution order is alphabetical by filename
    (`Collections.sort` in `Rules.discover()`), so the prefix *is* the
    schedule. Check `streams/` for the nearest neighbours before choosing.
+   When the slot you need is between two *consecutive* numbers, reuse the lower
+   one and let the name carry the ordering — the sort is over the whole
+   filename, so `506-distill-filter-frame` < `506-distill-predicate-frame` <
+   `507-distill-filter-frame-fallback` — and say so in the header comment.
 2. **Write the `.phr` file** under
    `src/main/resources/org/eolang/hone/rules/streams/Nxx/` (the `Nxx/`
    subdirectory whose digit matches the rule's hundreds prefix). Start from
--- a/src/main/resources/org/eolang/hone/rules/streams/5xx/506-distill-filter-frame.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/5xx/506-distill-filter-frame.phr
@@ -26,10 +26,11 @@
 #     `returned` primitive there (its `accepted` is the wrapper), so `returned`
 #     (= J -> long) names it, while bridge-input would wrongly declare a one-slot
 #     java/lang/Long and fail JVM verification ("Inconsistent stackmap frames").
-# Every other marker — an OBJECT-predicate filter such as a boxed-roundtrip
-# `(Ljava/lang/Integer;)Z`, whose survivor IS bridge-input even though trailing
-# maps make `returned` a different (re-boxed or primitive) type — falls through
-# to 507, which derives from bridge-input. A same-locals-1 frame is enough — a
+# An OBJECT-predicate filter — a boxed-roundtrip `(Ljava/lang/Integer;)Z` or the
+# `(Ljava/lang/String;)Z` left by a type-changing map fused ahead of the filter
+# (#794) — is claimed by `506-distill-predicate-frame`, which reads the survivor
+# off the predicate's own parameter; only a marker with no predicate in front of
+# it reaches 507 and its bridge-input. A same-locals-1 frame is enough — a
 # stateless wrapper owns no per-call counter. Scoped to a distill-lambda WITHOUT
 # `captured`, so stateful keep-labels stay with 503/504.
 name: distill-filter-frame
--- a/src/main/resources/org/eolang/hone/rules/streams/5xx/506-distill-predicate-frame.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/5xx/506-distill-predicate-frame.phr
@@ -1,0 +1,105 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# Resolves the Φ.hone.frame-item marker that 304 leaves at a STATELESS filter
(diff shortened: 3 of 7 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (247 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2814](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #800, opened 2026-08-12
fix(#799): narrow the Object item across a dropWhile latch

### LINKED ISSUE #799: VerifyError: neither 505 nor 508 narrows the Object item when dropWhile() follows distinct()
A `dropWhile` fused behind `distinct()` never gets its item narrowed, and the class dies at load. `Stream.of(3L, 1L, 2L).distinct().dropWhile(n -> n < 4L).count()`, rewritten with `streams/*`:

```
java.lang.VerifyError: Bad type on operand stack
  Location: distill_5028152(Ljava/util/List;Ljava/lang/Object;Ljava/util/function/Consumer;)V @49: invokestatic
  Reason: Type 'java/lang/Object' (current frame, stack[1]) is not assignable to 'java/lang/Long'
```

`distinct()` erases the element type, so the fused stateful distill carries `bridge-input ↦ Ljava/lang/Object;`, while the desugared `dropWhile` predicate is typed `lambda$main$0(Ljava/lang/Long;)Z`. Narrowing that item is the job of `505-distill-state-item-cast.phr:55-56`, which anchors on a `Φ.jeo.frame` immediately followed by an `invokestatic`, and of `508-distill-state-item-cast-dup.phr:52-54`, which tolerates one `dup` between them (the filter and peek interleaving from #788). A `dropWhile` puts its latch prologue in that gap. The rewritten body reads `frame → aload 0 → iload 3 → List.get → iinc → checkcast [I → astore 4 → aload 4 → iconst_0 → iaload → ifne → dup → invokestatic`, so neither adjacency matches, no `checkcast` is spliced, and the `Object` item reaches the typed predicate.

It is not about `Long`: `Stream.of("x", "y").distinct().dropWhile(s -> s.isEmpty()).count()` fails the same way with `java/lang/String`. And it is specific to the latch operator — `distinct().filter(...)` and `distinct().map(...)` both survive, because 508 and 505 respectively match those.

The smallest fix is a sibling of 508 that keeps its `when` guard and its splice point but allows the latch prologue between the keep-frame and the `dup`; alternatively, anchor on the `dup` → typed `invokestatic` pair alone when `bridge-input` is `Object`, since the cast is correct wherever that pair appears.

Related but distinct: #788 is the filter interleaving that 508 closed, #649 is the `boxed()`/`mapToX` tail that 505 closed, and #718 is the `dropWhile` lifting itself.

yegor256:
Closes #799.

## The bug

`Stream.of(3L, 1L, 2L).distinct().dropWhile(n -> n < 4L).count()`, rewritten with `streams/*`, dies at load:

```
java.lang.VerifyError: Bad type on operand stack
  Location: distill_5028152(Ljava/util/List;Ljava/lang/Object;Ljava/util/function/Consumer;)V @49: invokestatic
  Reason: Type 'java/lang/Object' (current frame, stack[1]) is not assignable to 'java/lang/Long'
```

`distinct()` and `skip()` erase the element type, so the fused stateful distill carries `bridge-input ↦ Ljava/lang/Object;`, while the desugared `dropWhile` predicate is typed `lambda$main$0(Ljava/lang/Long;)Z`. Narrowing that item is the job of `505-distill-state-item-cast` (anchors on a `Φ.jeo.frame` immediately followed by an `invokestatic`) and of `508-distill-state-item-cast-dup` (tolerates one `dup` in that gap, the filter/peek interleaving from #788).

`310-dropwhile-to-distill` opens its body with the sticky-latch prologue and only *then* dups the item for the predicate, so the joined body reads

```
frame → fetch [I → astore 4 → aload 4 → iconst_0 → iaload → ifne → dup → invokestatic
```

Neither adjacency matches, no `checkcast` is spliced, and the erased `Object` reaches the typed predicate.

## The fix

`509-distill-state-item-cast-latch.phr` — the latch-tolerant third of the 505 family. It matches that prologue opcode by opcode (including 310's fixed scratch slot 4, so only a dropWhile latch is tolerated in the gap, not an arbitrary run of opcodes) and keeps 505's splice point: the `checkcast` goes right after the keep-frame, before the prologue. The prologue leaves the stack exactly as it found it — the fetched array is stored away, the latch read is consumed by the `ifne` — so the item is narrowed on entry to the latch block and both copies the following `dup` makes are narrowed: the one the predicate eats and the one that survives on the keep path. The frame itself is untouched and still rightly declares `Object`, since the cast runs after it; 503 is left unchanged, exactly as in 505 and 508.

The `when` guard is 505's and 508's verbatim: `bridge-input` must be `Object` and the following `invokestatic` must take a single non-`Object` reference argument, so an `Object`-accepting predicate is left alone and a captured `lambda$` is skipped. It slots between 508 and 511 — after 503 has turned the keep-label's `frame-item` into a `Φ.jeo.frame`, before 512 wraps the body into a method and 521 resolves the `Φ.hone.fetch` this pattern anchors on.

Also updated the headers of 505 and 508 to cross-reference the new twin, and 310's header to record that its prologue shape is now load-bearing for 509.

## Tests

- `src/test/phino/509-distill-state-item-cast-latch.yml` — single-rule pack: 503 + 509 over the interleaved body, asserting the `checkcast java/lang/Long` lands between the frame and the fetch and that the latch block is otherwise untouched.
- `src/test/resources/org/eolang/hone/optimize/streams/stateful-dropwhile-tail.yml` — end-to-end, four pipelines covering every axis the issue names: distinct-led with a `Long` element (the reported reproduction, widened so the latch actually flips), distinct-led with a `String` element, a `skip`-led head, and two `dropWhile`s behind one `distinct` so the rule has to fire at fixpoint on the second latch block too.

## Verification

Run on this machine against phino `0.0.106` and jeo `0.15.3`, driving the real `jeo:disassemble → phino rewrite → jeo:assemble` pipeline with the production `--max-cycles 1 --max-depth 500`:

- The new fixture reproduces the reported `VerifyError` with 509 removed, and prints the correct `The result is 3/x+yy/2/3` with it in place — matching the unoptimized `javac` output.
- 509 fires five times across those pipelines (`Long`, `String`, `String` behind `skip`, and both latches of the chained case), each splicing the right wrapper.
- No regression on the existing fixtures exercised end to end the same way: `all-ops`, `dropwhile`, `full-non-terminal`, `stateful-ops`, `stateful-filter-tail`, `stateful-boxed-tail`, `stateful-object-tail`, `distinct`, `skip`, `peek-filter` — all identical to their `javac` baselines.
- `mvn test` → 47 tests, 0 failures. `mvn -Pdeep -Dtest='OptimizeMojoTest#appliesPhinoRulesAsSpecifiedInYamlPack' test` → 68 packs, 0 failures.
- The Docker-based `optimizesAsSpecifiedInYamlPack` deep run was not exercised here (no image build in this environment); CI covers it.

No opcode counts change for existing fixtures — none of them assert `checkcast`.

---
_Generated by [Claude Code](https://claude.ai/code/session_011toVZBZDE1GcXHL6e8wDiA)_

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/3xx/310-dropwhile-to-distill.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/3xx/310-dropwhile-to-distill.phr
@@ -37,6 +37,15 @@
 # locals (4 = latch array, 5 = notZ) are dead at KEEP, so 503's four-local frame
 # is still correct; 512's max-locals is widened to 6 to make room for them.
 #
+# The latch prologue (the first six opcodes, up to and including the `ifne`) is
+# spelled out opcode by opcode in 509, which narrows the erased Object item a
+# distinct()/skip() ahead of this dropWhile leaves on the stack (#799): the
+# prologue sits between that guard's keep-frame and the `dup` this body makes
+# for its predicate, so neither 505 nor 508 can reach the typed predicate. Any
+# change to the prologue — its opcodes, their order, or scratch slot 4 — must be
+# mirrored there, or the cast silently stops being spliced and the class fails
+# JVM verification again.
+#
 # v1 scope (see 208): object java/util/stream/Stream, a static predicate whose
 # target returns Z, the element type preserved across any fused neighbours. A
 # lone dropWhile is NOT reverted to a native Stream.dropWhile the way 442 reverts
--- a/src/main/resources/org/eolang/hone/rules/streams/5xx/505-distill-state-item-cast.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/5xx/505-distill-state-item-cast.phr
@@ -40,7 +40,9 @@
 # (spliced by 283, or by 431 after a fuse), breaking the frame → invokestatic
 # adjacency this rule anchors on; 508 is the twin that matches that
 # interleaving and casts in front of the dup, so both copies of the item are
-# narrowed (#788).
+# narrowed (#788). A dropWhile wedged there opens with 310's sticky-latch
+# prologue and only then dups, breaking both adjacencies at once; 509 is the
+# third twin, matching that prologue and casting in front of it (#799).
 name: distill-state-item-cast
 pattern: |
   ⟦
--- a/src/main/resources/org/eolang/hone/rules/streams/5xx/508-distill-state-item-cast-dup.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/5xx/508-distill-state-item-cast-dup.phr
@@ -38,6 +38,11 @@
 # distill-lambda body into a method), applied at fixpoint: the checkcast it
 # inserts now sits between the frame and the dup, so neither this rule nor 505
 # re-matches the same site.
+#
+# Only a BARE dup is tolerated here. A dropWhile also arrives with a dup in
+# front of its predicate, but 310 puts its sticky-latch prologue between the
+# keep-frame and that dup, so this adjacency does not match either; 509 is the
+# third twin, matching that prologue and casting in front of it (#799).
 name: distill-state-item-cast-dup
 pattern: |
   ⟦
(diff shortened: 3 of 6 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (275 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2818](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #802, opened 2026-08-12
fix(#798): frame a stateful guard from its survivor

### LINKED ISSUE #798: VerifyError: 503 frames the keep-label stack from bridge-input, so a type-changing map ahead of distinct() names the wrong type
A `map` that changes the object type, fused ahead of `distinct()`, produces a class the verifier rejects. `Stream.of("a", "bb").map(s -> Long.valueOf(s.length())).distinct().count()`, rewritten with `streams/*`, dies at class load:

```
java.lang.VerifyError: Inconsistent stackmap frames at branch target 29
  Location: distill_5028152(Ljava/util/List;Ljava/lang/String;Ljava/util/function/Consumer;)V @29
  Reason: Type 'java/lang/Long' (current frame, stack[0]) is not assignable
          to 'java/lang/String' (stack map, stack[0])
```

The frame comes from `src/main/resources/org/eolang/hone/rules/streams/5xx/503-distill-state-frame.phr:59` and `:65`, which fill both the local at slot 1 and the single stack item of the stateful guard's keep-frame from `bridge-input`, the fused body's *input* type. The local is right — it is the bridge method's declared parameter, `Ljava/lang/String;` here. The stack item is not: the value alive at the keep-label is the survivor, and a type-changing map fused ahead of the guard has already turned it into the map's return type. The rule's own header says as much at lines 24-26 ("v1 reads the type off `bridge-input`, correct when every operator before distinct in the fused body preserves the element type"). A small-steps dump confirms it — the `Φ.hone.frame-item` marker is resolved at 503, and the frame it writes reads locals `[java/util/List, java/lang/String, java/util/function/Consumer, integer]` with stack `[java/lang/String]` while a `Long` is what sits there.

Any object-to-object map does it: `String -> Long`, `String -> StringBuilder`, `Long -> ` a user class. `map(String::length).distinct()` survives, since a method reference leaves no desugared body to fuse ahead of the guard.

The smallest fix is to leave `x1` on `bridge-input` and derive only the stack item from the survivor, the way `506-distill-predicate-frame` already does for the stateless filter branch.

This is the stateful twin of #794, which fixed the same mistake in the filter path; #704 and #713 are the primitive shapes of it, and #788 is a missing `checkcast` rather than a wrong frame.

yegor256:
Closes #798.

## The defect

`503-distill-state-frame` fills **both** the local at slot 1 **and** the single stack item of a stateful guard's keep-frame from the distill's `bridge-input`. The local is right — it is the bridge method's declared parameter, the item `512` loads with `aload 1`. The stack item is not: a stateful guard is type-transparent, so the value alive at the keep-label is whatever the fused body produced just before it, and `401` is free to fuse a type-**changing** map in ahead.

```
java.lang.VerifyError: Inconsistent stackmap frames at branch target 29
  Location: distill_5029480(Ljava/util/List;Ljava/lang/String;Ljava/util/function/Consumer;)V @29
  Reason: Type 'java/lang/Long' (current frame, stack[0]) is not assignable
          to 'java/lang/String' (stack map, stack[0])
```

This is the stateful twin of #794, which fixed the same mistake on the stateless filter branch. All three stateful guards are affected, each reproduced and confirmed locally before the fix:

| guard | pipeline | error |
| --- | --- | --- |
| `307` distinct | `.map(s -> Long.valueOf(s.length())).distinct()` | `'java/lang/Long'` not assignable to `'java/lang/String'` |
| `308` skip | `.map(s -> Long.valueOf(s.length())).skip(1L)` | same, at branch target 33 |
| `310` dropWhile | `.map(s -> Long.valueOf(s.length())).dropWhile(v -> v < 2L)` | same, at branch target 48 |

A method-reference map retypes it the other way round and fails just as hard (`.map(Object::toString).distinct()` → `'java/lang/String'` not assignable to `'java/lang/Long'`).

## The fix

Three rules that sort in front of `503` and derive **only** the stack item from the survivor, leaving the local on `bridge-input` the way `504` already does:

- `503-distill-distinct-frame` and `503-distill-skip-frame` — these guards only ever see `Set.add(Object)` and a `long[]`, so the survivor has to come from the other side: the return type of the `invokestatic` matched immediately in front of the guard's `fetch` marker. That adjacency is exact — a distill body is emit-shaped, so the last opcode before the guard *is* the item's producer, and `307`/`308` open with the fetch. They fire only for an object return; a void one (a `peek` fused ahead) or a primitive one falls through.
- `503-distill-dropwhile-frame` — `310`'s guard names its own survivor, through the predicate it tests the item with, so it needs no producer in front of it. Reads it the way `506-distill-predicate-frame` does for the stateless filter branch.

No integer is free between 502 and 503, so all three share 503's number and rely on alphabetical filename order being the schedule: `503-distill-distinct-frame` < `503-distill-dropwhile-frame` < `503-distill-skip-frame` < `503-distill-state-frame`. `503` keeps the guards whose neighbour preserves the element type, or that have nothing fused ahead of them at all, where `bridge-input` is right; its header now says so instead of carrying the "v1" caveat the issue quoted.

## Tests

- `src/test/phino/503-distill-{distinct,dropwhile,skip}-frame.yml` — one single-rule pack per guard, each listing `503` alongside the new rule so the ordering is asserted too.
- `src/test/resources/org/eolang/hone/optimize/streams/map-retype-stateful.yml` — end-to-end, all three guards plus both retyping shapes in one class; the printed line asserts the optimized class verifies *and* still computes the right answers.

## Verification

Docker is unavailable in this environment, so the `@DisabledWithoutDocker` optimize fixtures were run through an equivalent host-`phino` harness (phino 0.0.106, the pinned version, JDK 21):

- 70/70 `src/test/phino` packs pass (`mvn -Pdeep test`), 67 before this change.
- 43 optimize fixtures: all pass except `all-ops` and `stateful-ops`, which fail **identically with the new rules removed** — `stateful-ops` pins `after new: 1` while dropWhile now folds into a second captured `List`, and `all-ops` differs in four counts. Both are pre-existing and untouched by this change.
- Negative controls: the new fixture fails with `VerifyError` when all three rules are removed, and still fails when only `503-distill-dropwhile-frame` is removed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MbeMLvuMsp86r5Q3txeeJT)_

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/5xx/503-distill-distinct-frame.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/5xx/503-distill-distinct-frame.phr
@@ -1,0 +1,170 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# Resolves the Φ.hone.frame-item marker that 307 leaves at a distinct() guard's
+# keep-label when an operator was fused in AHEAD of that guard, so the survivor
+# waiting on the stack is no longer the value the wrapper was handed.
+#
+# 503 fills BOTH the local at slot 1 and the single stack item of the keep-frame
+# from the distill's `bridge-input`. The local is right — it is the bridge
+# method's declared parameter, the item 512 loads with `aload 1`. The stack item
+# is not: distinct is type-transparent, so the value alive at the keep-label is
+# whatever the body produced just before the guard, and 401 is free to fuse a
+# type-CHANGING map in ahead of it. For `Stream.of("a", "bb")
+# .map(s -> Long.valueOf(s.length())).distinct()` (#798) the distill still
+# receives a java/lang/String, the head `lambda$main$0:(Ljava/lang/String;)
+# Ljava/lang/Long;` has already retyped it, and the frame 503 writes promises a
+# java/lang/String over a java/lang/Long: "java.lang.VerifyError: Inconsistent
+# stackmap frames at branch target 29 ... Type 'java/lang/Long' (current frame,
+# stack[0]) is not assignable to 'java/lang/String' (stack map, stack[0])".
+# Keeping the element type identical hides it (`map(x -> x + 1).distinct()` in
+# `distinct.yml` always ran), which is why every other fixture passed.
+#
+# This is the stateful twin of #794, fixed by `506-distill-predicate-frame` for
+# the stateless filter branch: leave the LOCAL on bridge-input and derive only
+# the STACK item from the survivor. A filter names its survivor through the
+# predicate's own parameter; distinct's guard only ever sees `Set.add(Object)`,
+# so the name comes from the other side — the return type of the invokestatic
+# that produced the item, matched immediately in front of the guard's fetch
+# marker. That adjacency is exact: a distill body is emit-shaped (one item in,
+# one item out), so the last opcode before the guard IS the item's producer,
+# and 307 opens its guard with the fetch. It fires only for an object return
+# (`(...)L<type>;`); a void one (a peek fused ahead) or a primitive one leaves
+# the marker to 503, whose bridge-input is right for a type-preserving
+# neighbour and for a guard with nothing fused ahead of it at all.
+#
+# `503-distill-skip-frame` does the same for 308's countdown guard, which is
+# just as blind to the element type; `503-distill-dropwhile-frame` covers 310's
+# latch guard, which does not need the producer at all because its own
+# predicate names the survivor.
+#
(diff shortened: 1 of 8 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (730 >= 100); -8 for way too many hits-of-code (730 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2822](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #803, opened 2026-08-12
fix(#801): stop 241 minting the no-op type that 271 deletes

### LINKED ISSUE #801: 241 mints a no-op Φ.hone.type that 271 deletes, and the pair loops the rewrite to death after an unboxing mapToInt
An unboxing `mapToInt` followed by a primitive `map` makes the rewrite loop and the build die. A class whose only pipeline is `INTEGERS.stream().mapToInt(Integer::intValue).map(n -> n + 2).sum()`, with `INTEGERS` a `List<Integer>`, ends like this:

```
[ERROR]: On rewriting step '1' of rule 'remove-useless-types' we got the same
         expression as we got at one of the previous step, it seems rewriting is looping
...
java.io.IOException: Failed to execute docker, code=0x0001
```

Two rules disagree about a node neither needs. `2xx/241-type-primitive-to-primitive-after-map.phr:66-77` mints a `Φ.hone.type` between two adjacent primitive operations whenever the first's `returned` matches `^[DFIJ]$` and the second's `accepted` matches `^[BCDFIJS]$` — nothing in that `when` forbids `from` from equalling `to`. And `2xx/271-remove-useless-types.phr:22-23` deletes precisely the nodes whose `from` equals its `to`. A small-steps dump of the repro shows the pair at work: step 43, which is 241, adds `⟦ φ ↦ Φ.hone.type, from ↦ "i", to ↦ "i", class ↦ "probe/R12" ⟧`, and step 49, which is 271, takes it away.

In small-steps mode every rule is its own process, so nothing notices. In the ordinary single-process run the two alternate; as long as some other rule is still changing the expression that is harmless, and `IntStream.of(1, 2, 3).map(n -> n + 2).map(n -> n * 3).sum()` mints and deletes the very same no-op node yet finishes. The unboxing head is what leaves the expression otherwise untouched, so the state comes back to one it already had and phino's loop detector kills the run. The same happens with `mapToLong(Integer::longValue).map(n -> n + 2L)`; it does not happen without the trailing primitive map, nor with the map put ahead of the unboxing.

The smallest fix is one more clause in 241's `when` (and in its filter twin `251-type-primitive-to-primitive-after-filter.phr`), refusing to mint the node when `from` would equal `to`. That node exists only for 271 to delete.

Related: #733 was the 603/205 unbox loop, and #700 is the opposite failure mode, a rewrite that fails quietly.

yegor256:
Closes #801.

## The bug

`2xx/241-type-primitive-to-primitive-after-map` mints a `Φ.hone.type` between two adjacent primitive operations whenever the first's `returned` matches `^[DFIJ]$` and the second's `accepted` matches `^[BCDFIJS]$`. Nothing in its `when` forbade `from` from equalling `to`. `2xx/271-remove-useless-types` deletes precisely the type nodes whose `from` equals its `to`.

On an identity pair the two took turns. Because the whole `streams/*` selection runs to fixpoint inside **one** `phino rewrite` invocation, the expression came back to a state it already had, phino's loop detector aborted, and the optimize goal died:

```
[ERROR]: On rewriting step '1' of rule 'remove-useless-types' we got the same
         expression as we got at one of the previous step, it seems rewriting is looping
...
java.io.IOException: Failed to execute docker, code=0x0001
```

An unboxing head is what exposes it. Once
`INTEGERS.stream().mapToInt(Integer::intValue).map(n -> n + 2).sum()` settles, the oscillation is the only thing still changing the expression, so the detector fires. Where some other rule keeps rewriting, the very same no-op node is minted and deleted harmlessly — which is why the equivalent `IntStream.of(1, 2, 3).map(n -> n + 2).map(n -> n * 3).sum()` always finished. In small-steps mode every rule is its own process, so nothing notices there either.

## The fix

241 now carries the `having` guard its filter twin `251-type-primitive-to-primitive-after-filter` **already had** — so this PR only needs to add it in one place, not two as the issue anticipated:

```yaml
having:
  not:
    eq:
      - 𝑒-type-from
      - 𝑒-type-to
```

Nothing else is lost by declining. With `from == to` the node is an identity conversion, and the two other edits 241 makes to `𝜏-second` reduce to the values it already holds:

- `returned` — for `𝜏-two ↦ map` the sed chain resolves back to `𝑒-second-returned` unchanged; for a filter it resolves to `𝑒-first-returned`, which under `from == to` is the filter's own `accepted`, and every filter producer (231/232) already sets `returned = accepted`.
- `bridge-output` — same shape, and 231/232 likewise emit `bridge-output = bridge-input`, the boxed wrapper of the same primitive letter on both sides.

Both twins got a header comment explaining why the guard is load bearing, so it does not get deleted later as redundant.

## Tests

- `src/test/phino/801-noop-type-no-loop.yml` — a single-rule pack on two adjacent `I -> I` maps, the exact shape whose `from ↦ "i", to ↦ "i"` node the issue's small-steps dump caught being minted at step 43 and deleted at step 49. It loads **both** 241 and 271; that is what makes it a regression test, since 241 alone terminates either way (the minted node separates the pair and stops it matching again). Before the fix `phino rewrite` exits non-zero on the loop; after it, the rewrite settles on the input and the expected pattern asserts the two maps are still adjacent, i.e. nothing was spliced between them.
- `src/test/resources/org/eolang/hone/optimize/streams/unbox-primitive-map.yml` — the end-to-end repro, covering both the `mapToInt(Integer::intValue)` and the `mapToLong(Integer::longValue)` form the issue reports. No opcode counts are asserted; this pack is about the build finishing at all and the optimized bytecode still summing to the same numbers.

## Verification

`mvn test` passes (47 tests, 0 failures). The `deep` suites could not be exercised here — neither `phino` nor a working Docker daemon is available in this container, and `deep.yml` only runs on push to `master` — so the two new packs are unverified against a real `phino` binary. Worth a `mvn -Pdeep test` before merge. I did confirm out of band that the fixture's Java compiles and prints `The results are 35 35` unoptimized, matching its `log` assertion, and that both new YAML files pass `yamllint -c .yamllint.yml` clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VYQru4RyeWEcTJR8knbSKu)_

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/2xx/241-type-primitive-to-primitive-after-map.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/2xx/241-type-primitive-to-primitive-after-map.phr
@@ -2,6 +2,30 @@
 # SPDX-License-Identifier: MIT
 ---
 # yamllint disable rule:line-length
+# Where a primitive map hands its result to another primitive operation that
+# accepts a different primitive, the two need a conversion spliced between
+# them; this rule mints the Φ.hone.type node that performs it.
+#
+# It must NOT mint one when the conversion is an identity. 271 deletes every
+# type node whose `from` equals its `to`, so on a `from == to` pair the two
+# rules take turns — 241 minting the no-op node, 271 taking it away — and,
+# because the whole streams/* selection runs to fixpoint inside ONE `phino
+# rewrite` invocation, the expression comes back to a state it already had and
+# phino's loop detector aborts the run ("it seems rewriting is looping"),
+# killing the build (#801).
+#
+# An unboxing head is what exposes it: in
+# `INTEGERS.stream().mapToInt(Integer::intValue).map(n -> n + 2).sum()` the
+# oscillation is the only thing still changing the expression by the time it
+# starts, so the detector fires. Wherever some other rule is still rewriting,
+# the very same no-op node is minted and deleted harmlessly, which is why
+# `IntStream.of(1, 2, 3).map(n -> n + 2).map(n -> n * 3).sum()` finishes.
+#
+# The `having` below is that guard — the identical one the filter twin
+# 251-type-primitive-to-primitive-after-filter already carries. It costs
+# nothing else: with `from == to` the node is a no-op conversion, and the
+# `returned` / `bridge-output` updates this rule makes to 𝜏-second reduce to
+# the values it already holds.
 name: type-primitive-to-primitive-after-map
 pattern: |
   ⟦
@@ -140,3 +164,8 @@ where:
       - 𝑒-two
       - 𝑒-updated-filter-bridge-output-pattern
       - 𝑒-updated-map-bridge-output-pattern
+having:
+  not:
+    eq:
+      - 𝑒-type-from
+      - 𝑒-type-to
--- a/src/main/resources/org/eolang/hone/rules/streams/2xx/251-type-primitive-to-primitive-after-filter.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/2xx/251-type-primitive-to-primitive-after-filter.phr
@@ -2,6 +2,13 @@
 # SPDX-License-Identifier: MIT
 ---
 # yamllint disable rule:line-length
+# The filter twin of 241: the same Φ.hone.type conversion node, spliced in
+# behind a filter rather than behind a map.
+#
+# The `having` guard at the bottom is load bearing, for the reason 241's
+# header spells out: a type node whose `from` equals its `to` is deleted by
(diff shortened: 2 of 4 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (150 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2826](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #806, opened 2026-08-12
fix(#668): generalise the capture channel to mixed and multi-reference runs

### LINKED ISSUE #668: 112-capturing-invokedynamic-to-map.phr:99-103: Generalise...
The puzzle `665-9de96954` from #665 has to be resolved: 

https://github.com/objectionary/hone-maven-plugin/blob/497a7364d2ab78af765765c8501a78de5f2b318b/src/main/resources/org/eolang/hone/rules/streams/1xx/112-capturing-invokedynamic-to-map.phr#L99-L103

The puzzle was created by @yegor256 on 03-Jun-26. 

Estimate: 45 minutes,  role: DEV. 

yegor256:
Closes #668 — the puzzle `665-9de96954` asked to generalise the capture channel to a mixed multi-capture run containing a reference. Still a bug on master, now fixed.

## The puzzle's premise, and what it missed

The puzzle framed the blocker as positional type extraction: `115`/`119` read the capture class out of `target.signature`'s **first** L-type, which is only the right answer when the run holds exactly one reference. That's accurate, and `112`/`113` now thread `captures` — the factory descriptor's parameter list — with the peel popping the last code off in step with the push it takes, so every capture is typed positionally.

Widening the guard on top of that alone would have shipped a crash. My first attempt passed every unit pack and then failed end-to-end with `VerifyError: Operand stack underflow`.

## Why the peel had to be rewritten

phino applies each rule to its **own fixpoint and then moves on**, and the plugin runs a single cycle over the rule list (`hone.max-cycles` defaults to `1`). A per-type peel therefore only ever drained a run whose types were already sorted in rule order. For `iload x; aload y`, the int rule sees an `aload` in front of the pragma, declines, and never gets another turn — `x` is left stranded on the stack in front of a pragma that no longer expects it.

**This defect predates the change.** Reproduced against master's own rules with an `(int, long)` capture run — `114` declines, `116` peels the `lload`, and the `iload` is stranded. `112`'s header claiming that mixed primitive runs "interleave on their own opcode" was wrong; it is latent today only because no fixture exercises that shape.

So `114` and `118` are now **one peel rule each, for every capture type**. The push rides through as a bound expression, and the only per-type difference left — the boxing — is deferred to `Φ.hone.cap-box` / `Φ.hone.cap-fetch` markers that `123`–`126` expand in place. That expansion is order-free because each marker is rewritten where it stands.

## Changes

- `112`/`113` — one capture-alphabet guard, `\(([IJD]|L[^;]+;)+\)L…Function;` / `…Predicate;`, replacing the primitives-only / single-reference `or`; seed `captures`.
- `114`/`118` — single order-independent peel per side, draining the whole run at `--max-cycles=1`; `115`/`116`/`117` and `119`/`120`/`122` deleted.
- `123`–`126` — marker expansions: box/drop and fetch/unbox, per type.
- `314`/`316` — bind `captures` only to drop it.
- Peels gain a stop condition, firing only while a capture is still owed. Without it the rule kept going once the run was exhausted and swallowed whatever `aload` preceded the operator — the receiver `Stream` itself, when a pipeline starts from a local variable.

For every case that already worked, the emitted bytecode is unchanged instruction for instruction: `dup; <push>; [box]; List.add; pop` and `fetch; [unbox]` are exactly what the per-type rules produced.

## Tests

New packs cover the mixed lift, the mixed peel on both sides, each of the four marker expansions, and the stop condition. Two end-to-end fixtures fuse a map capturing `(List, int, String)` and a filter capturing `(List, int)` — both shapes master declines outright.

Verified locally with the pinned phino 0.0.106 (extracted from `yegor256/hone:latest`): all 69 single-rule packs pass. Full `-Pdeep` run was still in flight at the time of opening this draft; two caveats from the earlier full run, both independent of this change:

- `all-ops` and `stateful-ops` opcode-count mismatches — neither fixture contains a capturing lambda, so `112`/`113` cannot fire on them; consistent with a javac codegen difference (JDK 21.0.10 here).
- Five Docker-related errors — `docker build` in this sandbox cannot reach `archive.apache.org` through the proxy.

https://claude.ai/code/session_01D5CH17Bm5eogRoV3GmQRge

---
_Generated by [Claude Code](https://claude.ai/code/session_01D5CH17Bm5eogRoV3GmQRge)_


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1923 >= 100); -8 for way too many hits-of-code (1923 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2850](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #808, opened 2026-08-12
fix(#804): let 283 reach a dropWhile predecessor

### LINKED ISSUE #804: 283 never matches a dropWhile predecessor, so a filter behind dropWhile loses its dup and the class fails verification
A `filter` fused behind a `dropWhile` never gets its guard copy, and the class dies at load. `Stream.of(3L, 1L, 2L).dropWhile(n -> n < 2L).filter(n -> n > 1L).count()`, rewritten with `streams/*` on master (850d4bfa):

```
java.lang.VerifyError: Inconsistent stackmap frames at branch target 52
  Location: distill_5027019(Ljava/util/List;Ljava/lang/Long;Ljava/util/function/Consumer;)V @48: ifne
  Reason: Current frame's stack size doesn't match stackmap
  Stackmap Frame: stack: { 'java/lang/Long' }
```

The predicate ate the only copy of the item, because no `Φ.hone.dup` was spliced in front of the filter. `src/main/resources/org/eolang/hone/rules/streams/2xx/283-insert-dup-before-filter-after-stateful.phr` is the rule for exactly this: its header names distinct, skip and dropWhile, and its `when` lists all three. It never fires here. Its pattern's `𝜏-first` at lines 21-25 reads `φ ↦ Φ.hone.𝜏-one, class ↦ 𝑒-class, 𝐵-first-rest`, while the `Φ.hone.dropWhile` node that `208-lambda-to-dropwhile.phr` mints carries `opcode` between `φ` and `class`. `216-recognize-distinct.phr` and `220-recognize-skip.phr` mint nodes whose `class` follows `φ` directly, which is why `distinct().filter(...)` survives and this does not.

A small-steps dump of the repro backs it up: the dropWhile node appears at step 30 (208) and is folded away at step 63 (310), and not one of the 106 intermediate expressions contains a `Φ.hone.dup`. Applying 283 by hand to the step-52 expression rewrites nothing; adding `opcode ↦ 𝑒-first-opcode` ahead of `class` in its pattern and result makes it fire and the dup appear.

The smallest fix is a second pattern for 283 (or a sibling rule beside it) whose `𝜏-first` lists `opcode` before `class`, so the dropWhile node's attribute order matches too.

Related: #790 was the peek-then-filter dup, and #801 is the other rule-level defect still open from the same randomized run.

yegor256:
Closes #804.

## The defect

`283-insert-dup-before-filter-after-stateful.phr` splices a `Φ.hone.dup` between a stateful guard and the filter behind it, so the predicate has a copy of the item to eat and the keep path still finds the item on the stack. Its `𝜏-first` spelled `class` out as the *second* binding of the guard:

```
𝜏-first ↦ ⟦ φ ↦ Φ.hone.𝜏-one, class ↦ 𝑒-class, 𝐵-first-rest ⟧
```

phino matches bindings **positionally** (`matchBindings` walks the two lists in lockstep; only a `𝐵` meta absorbs a run). That shape fits the pragmas `216` and `220` mint for `distinct` and `skip`, which put `class` directly behind `φ` — but not the one `208` mints for `dropWhile`, which carries `opcode` in between because it is lambda-bearing, like `map` and `filter`. So the rule matched the first two and walked past every `dropWhile`: no dup was spliced, the predicate ate the only copy of the item, and the keep label was reached with an empty stack against a frame promising the element.

```
java.lang.VerifyError: Inconsistent stackmap frames at branch target 52
  Location: distill_5021652(Ljava/util/List;Ljava/lang/Long;Ljava/util/function/Consumer;)V @48: ifne
```

## The fix

A `𝐵-first-head` meta now absorbs whatever the guard puts between `φ` and `class` — none of it is anything this rule reads — so all three guards match and the binding order a fourth one picks will not matter either. The rule is otherwise untouched: `when` still narrows `𝜏-one` to `distinct`/`skip`/`dropWhile`, and `𝑒-dup-opcode` is still read off the filter's `bridge-input`.

## Tests

- `src/test/phino/283-insert-dup-before-filter-after-dropwhile.yml` — the binding order that regressed; asserts the `Φ.hone.dup` lands between the two pragmas.
- `src/test/phino/283-insert-dup-before-filter-after-distinct.yml` — the other order, where the new meta has to bind the empty run and leave the site rewritten exactly as before.
- `src/test/resources/org/eolang/hone/optimize/streams/dropwhile-filter.yml` — end to end: the reported reproduction, the same shape on a `String` element, a `map` behind the filter so the whole run fuses into one `mapMulti`, and two filters behind one `dropWhile` so the rule has to fire again at fixpoint. Asserts what the four pipelines print. **On `master` this pack fails with the exact `VerifyError` above; with the fix it passes.**
- `RandomPipeline` loses its `#804` quarantine clause (`dropsThenFilters`), so the walk reaches the shape again. 2 of the 120 default seeds put a `filter` behind a `dropWhile`, and `preservesWhatRandomPipelinesPrint` is green with them in.

## What was run

Against phino `0.0.106` on the host (no Docker), `LC_ALL=C.UTF-8`:

| Command | Result |
| --- | --- |
| `mvn test` | 52 run, 0 failures |
| `mvn -Pdeep test -Dtest='OptimizeMojoTest#appliesPhinoRulesAsSpecifiedInYamlPack'` | 76 run, 0 failures |
| `mvn -Pdeep test -Dtest='OptimizeMojoTest#preservesWhatRandomPipelinesPrint'` | passed |
| `mvn -Pdeep test -Dtest='OptimizeMojoTest#optimizesAsSpecifiedInYamlPack'` | 48 run, 2 failures — both pre-existing, see below |
| `mvn clean install -DskipTests -Pqulice` | BUILD SUCCESS |

## Pre-existing failures, not touched here

`all-ops.yml` and `stateful-ops.yml` fail their `after` opcode counts, identically before and after this change (verified by running the same suite on `master`). Both are the two packs that contain a `dropWhile`, and both look like counts recorded before `#718` taught `208`/`310` to recognise it — `stateful-ops.yml` still says in prose that "takeWhile and dropWhile are NOT recognised yet", while its fused `dropWhile` now allocates a second `ArrayList` (`new` 1 → 2). They are a separate defect from this one, so this PR leaves them alone rather than re-recording numbers measured on a different JDK than the reference image.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FSP6qGpbA3Pn85u6dcSBes)_

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/2xx/283-insert-dup-before-filter-after-stateful.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/2xx/283-insert-dup-before-filter-after-stateful.phr
@@ -14,12 +14,25 @@
 # stateful guards and 𝜏-second is a filter, it splices the same Φ.hone.dup
 # between them. After
 # insertion the two are no longer adjacent, so it fires once per site.
+#
+# The three guards disagree about the ORDER of their bindings, and phino
+# matches bindings positionally — a 𝐵 meta swallows a run of them, everything
+# else has to line up one for one. 216 and 220 mint `class` directly behind
+# `φ`, but the Φ.hone.dropWhile that 208 mints carries `opcode` in between,
+# because it is a lambda-bearing pragma shaped like map and filter. A 𝜏-first
+# that spelled `class` out as the second binding therefore matched distinct
+# and skip and silently walked past every dropWhile: no dup was spliced, the
+# predicate ate the only copy of the item, and the class died at load with
+# "Inconsistent stackmap frames" (#804). 𝐵-first-head absorbs whatever the
+# guard puts between `φ` and `class` — none of it is anything this rule reads
+# — so the ordering a fourth guard picks will not matter either.
 name: insert-dup-before-filter-after-stateful
 pattern: |
   ⟦
     𝐵-before,
     𝜏-first ↦ ⟦
       φ ↦ Φ.hone.𝜏-one,
+      𝐵-first-head,
       class ↦ 𝑒-class,
       𝐵-first-rest
     ⟧,
@@ -41,6 +54,7 @@ result: |
     𝐵-before,
     𝜏-first ↦ ⟦
       φ ↦ Φ.hone.𝜏-one,
+      𝐵-first-head,
       class ↦ 𝑒-class,
       𝐵-first-rest
     ⟧,
--- a/src/test/java/org/eolang/hone/RandomPipeline.java
+++ b/src/test/java/org/eolang/hone/RandomPipeline.java
@@ -34,14 +34,13 @@ import java.util.Random;
  * the API allows to skip the traversal altogether.</p>
  *
  * <p>The grammar does not step around the operations that the rewrite has been
- * known to break — every shape behind #787, #788, #790, #794, #798 and #799 is
- * still reachable, since a generator that avoids the defects we already know
- * about would only prove that we know about them, and nothing about a
- * regression. Two shapes are quarantined all the same, in
- * {@code quarantined()}: they break the rewrite today, they are filed as #804
- * and #805, and a suite that fails on them every night reports nothing new.
- * Each is named by its issue, so closing the issue widens the walk by deleting
- * one clause.</p>
+ * known to break — every shape behind #787, #788, #790, #794, #798, #799 and
+ * #804 is still reachable, since a generator that avoids the defects we already
+ * know about would only prove that we know about them, and nothing about a
+ * regression. One shape is quarantined all the same, in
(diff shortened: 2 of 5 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (238 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2842](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #805, opened 2026-08-12
VerifyError returns when a peek, a filter or a second guard sits between the retyping map and the stateful guard


yegor256:
The #798 fix only reaches the guard that sits directly behind the retyping map. As soon as anything at all comes between them, the keep-frame goes back to naming the type the wrapper was handed, and the class fails verification with the very same message. Four pipelines on master (850d4bfa):

```java
Stream.of("a", "bb").map(s -> new Pair(1L, s)).peek(p -> SUM[0] += p.key()).skip(1L).count()
Stream.of("a", "bb").map(s -> new Pair(1L, s)).filter(p -> p.key() > 0L).distinct().count()
Stream.of(1, 2).map(n -> n.toString()).dropWhile(s -> s.length() < 3).distinct().count()
Stream.of(1, 2).map(n -> n.toString()).distinct().distinct().count()
```

The first two die with `Type 'probe/Pair' (current frame, stack[0]) is not assignable to 'java/lang/String' (stack map, stack[0])`, the last two with `Type 'java/lang/String' ... is not assignable to 'java/lang/Integer'`, all under `VerifyError: Inconsistent stackmap frames`. What master repaired still holds: `map(s -> new Pair(1L, s)).skip(1L)`, `map(...).distinct()` and `map(...).dropWhile(...)` all run.

`src/main/resources/org/eolang/hone/rules/streams/5xx/503-distill-distinct-frame.phr:60-68` (and its twin `503-distill-skip-frame.phr`) reads the survivor off an `invokestatic` required to sit *immediately* in front of the guard's `Φ.hone.fetch` marker. The header calls that adjacency exact, and says a void producer — a peek fused ahead — "leaves the marker to 503, whose bridge-input is right for a type-preserving neighbour". It is not right there: a peek does not change the item, so the survivor is still the map's return type, and `503-distill-state-frame.phr:59,65` writes `bridge-input`. The same happens with a filter in between, and with a second guard, whose producer is the first guard rather than the map.

The smallest fix is to let those rules look past a producer that cannot have retyped the item — a void invokestatic, or a guard of their own kind — and take the nearest object-returning `invokestatic` instead of insisting on the preceding opcode.

Found by the randomized differential walk on branch 791; #801 and #804 are the other two still open from that run.

0crat:
[...]

PR #812, opened 2026-08-12
fix(#807): reach the last corners of the Stream API

### LINKED ISSUE #804: 283 never matches a dropWhile predecessor, so a filter behind dropWhile loses its dup and the class fails verification

### LINKED ISSUE #807: RandomPipeline.java:47-51: Reach the last corners of the...
The puzzle `791-655f04e9` from #791 has to be resolved: 

https://github.com/objectionary/hone-maven-plugin/blob/4f8de379c31b6218b897f5c9f99ffd3a469ed541/src/test/java/org/eolang/hone/RandomPipeline.java#L47-L51

The puzzle was created by @yegor256 on 12-Aug-26. 

Estimate: 60 minutes,  role: DEV. 

yegor256:
Closes #807, resolving puzzle `791-655f04e9`.

The puzzle named four corners of the Stream API the walk could not reach. All four are now productions of the grammar in `RandomPipeline`.

### `Stream<Double>`

A new element domain, `dbox`, sitting alongside `boxed` (`Stream<Long>`) and `ibox` (`Stream<Integer>`) and treated like them everywhere. It is in `DOMAINS`, so a walk may start there; it carries five sources, twelve stages, a `peek`, nine turns out (into `reals`, `long`, `ints`, `boxed`, `ibox`, `words`, `pairs`) and eleven terminals; six other domains gained a turn into it. The generated class gained a `DOUBLES` constant for the sources that stream a list.

### Custom `Collector`

Two terminals assemble one by hand with `Collector.of` — a `StringBuilder` on `words`, a `long[]` on `boxed` — next to the existing `collect(ArrayList::new, List::add, List::addAll)`.

### `teeing` and `flatMapping`

`Collectors.teeing` on `words` and on `dbox`; `Collectors.flatMapping` on `words`, `boxed` and `ibox`.

### `Optional` chains

Seven terminals now put a `map` and a `filter` between the `Optional` and its `orElse`, and three of them reach it through `OptionalLong.stream()` / `OptionalInt.stream()` / `OptionalDouble.stream()` rather than starting from `findFirst()`.

### One wrinkle worth knowing

A pipeline is printed through `String.valueOf`, whose overload set bounds the result of a terminal by `char[]` as well as by `Object`. A `teeing` or an `of` whose result type javac has to infer cannot satisfy both bounds, so those four terminals spell their type arguments out. That makes them wider than a line of the grammar, so they are joined from pieces — checkstyle prohibits concatenating literals. Both facts are written down in the `GRAMMAR` javadoc.

### What was verified

Over 500 seeds, locally and in `OptimizeMojoTest`:

- every production of the grammar is reached (`reachesEveryProductionOfTheGrammar`);
- every generated class compiles with no javac diagnostics (`generatesPipelinesThatCompile`);
- every generated class runs to completion without an exception;
- no `peek` sits in front of `count()` (`avoidsAPeekInFrontOfCount`).

`mvn test` is green (52 tests) and `mvn -Pqulice install` is clean, both re-run after merging `master`.

What is **not** verified here: `preservesWhatRandomPipelinesPrint`, the end-to-end test that actually rewrites these pipelines and diffs what they print. It needs Docker or a host `phino`, and neither is reachable from this environment. It is also not run by any PR workflow — `deep.yml` triggers on push to `master` only — so the widened walk first meets the rewrite after this merges. If it turns a shape up that the rewrite gets wrong, that is the generator doing its job, and the fix is the established one: a narrow clause in `quarantined()` naming the issue that owns the shape.

### Merged with `master`

`master` moved under this branch: #808 and #809 fixed #804 and #805 and deleted both quarantine clauses, along with the `OBJECTS`, `GUARDS` and `BARRIERS` tables and the predicates that read them. This branch takes that deletion as it stands — the `dbox` entry it had added to `OBJECTS` went with the table — and everything above was re-verified against the merged tree.

### DIFF
--- a/src/test/java/org/eolang/hone/RandomPipeline.java
+++ b/src/test/java/org/eolang/hone/RandomPipeline.java
@@ -43,12 +43,15 @@ import java.util.Random;
  * Each is named by its issue, so closing the issue widens the walk by deleting
  * one clause.</p>
  *
+ * <p>The walk reaches the three primitive domains, all three of their boxed
+ * counterparts — {@code Stream<Long>}, {@code Stream<Integer>} and
+ * {@code Stream<Double>} — and two reference domains, one of strings and one of
+ * a small value class of its own. It reaches the corners of the terminal API as
+ * well: collectors assembled by hand out of {@code Collector.of},
+ * {@code Collectors.teeing}, {@code Collectors.flatMapping}, and {@code Optional}
+ * chains longer than a single {@code orElse}.</p>
+ *
  * @since 0.30.0
- * @todo #791:60min Reach the last corners of the API. There is no
- *  {@code Stream<Double>} domain, no custom {@code Collector}, no
- *  {@code Collectors.teeing} or {@code flatMapping}, and no {@code Optional}
- *  chain longer than one {@code orElse}. Everything parallel stays out on
- *  purpose: the printed value is the oracle, so it has to be deterministic.
  */
 final class RandomPipeline {
 
@@ -71,6 +74,13 @@ final class RandomPipeline {
      * pins how much of the stream it walks. Nothing here is parallel and nothing
      * is {@code findAny}: the value a pipeline prints has to be the same on
      * every run, or it cannot be an oracle.</p>
+     *
+     * <p>A production wider than one line is joined from pieces, which is what
+     * the few {@code String.join} entries are. The terminals that tee, and the
+     * ones that assemble a {@code Collector} by hand, spell their type arguments
+     * out: a pipeline is printed through {@code String.valueOf}, and its
+     * overloads bound the result of such a terminal by {@code char[]} as well as
+     * by {@code Object}, which no inferred type can satisfy.</p>
      */
     private static final String GRAMMAR = String.join(
         RandomPipeline.EOL,
@@ -98,6 +108,7 @@ final class RandomPipeline {
         "long|turn|mapToObj(n -> new Pair(n, Long.toString(n)))|pairs",
         "long|turn|mapToInt(n -> (int) n)|ints",
         "long|turn|asDoubleStream()|reals",
+        "long|turn|mapToObj(n -> (double) n)|dbox",
         "long|end|*sum()",
         "long|end|*reduce(0L, Long::sum)",
         "long|end|*max().orElse(0L)",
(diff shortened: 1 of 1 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (101 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2850](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #813, opened 2026-08-12
all-ops.yml and stateful-ops.yml assert `after` opcode counts that predate dropWhile recognition and the flatMap try-with-resources


yegor256:
`mvn -Pdeep test -Dtest='OptimizeMojoTest#optimizesAsSpecifiedInYamlPack'` on master (e11ad2bc) is 48 run / 2 failures. Both are `after`-stage opcode counts, and both packs are stale rather than miscompiled: the `log:` assertions run before `assertOpcodes` and pass, so the optimized classes verify and still print the right answer. Only the numbers of record have drifted.

```
all-ops       expected  invokespecial 2  invokestatic 31  invokevirtual 14  invokedynamic 14  return 17
              actual    invokespecial 3  invokestatic 32  invokevirtual 18  invokedynamic 14  return 19
stateful-ops  expected  new 1   (baload 0, bastore 0)
              actual    new 2   (baload 0, bastore 0 — those two are still right)
```

Not a regression from #808: the same two fail identically with that change stashed.

## What moved, established by rule selection

Re-running each pack through the plugin with `<rules>` negating one rule at a time pins both drifts exactly.

**`208-lambda-to-dropwhile` (dropWhile recognition, #718)** accounts for `stateful-ops` entirely and for four of the five counts in `all-ops`:

| selection | stateful-ops `new` | all-ops invokespecial / invokestatic / return |
| --- | --- | --- |
| `streams/*` | 2 | 3 / 32 / 19 |
| `streams/*,!streams/2xx/208-*` | **1** ✅ | **2 / 31 / 17** ✅ |

Those are the fixtures' numbers to the digit. `stateful-ops` still says so in prose, which is the clearest evidence it was never revisited:

> takeWhile and dropWhile are NOT recognised yet — no rule folds them — so they survive as their original lambdas and allocate no state cell (hence baload == bastore == 0). When they are migrated to the distill they will allocate a boolean[1] guard and these counts will change.

`dropWhile` has since been migrated (`208` → `310`). The prediction was half right: the latch is an `int[1]` (`newarray`, not `baload`/`bastore` — those correctly stayed 0) wrapped in its own captured `ArrayList`, and that `ArrayList` is the second `new`.

[...]

PR #815, opened 2026-08-12
fix(#811): unbox a boxed-primitive run before its guard, not after

### LINKED ISSUE #811: Operand stack underflow when a flatMapToLong is fused between a sorted() barrier and two filters
The randomized differential walk reaches a shape that optimizes into a class the verifier rejects with `Operand stack underflow`. No stateful guard is involved at all — no `distinct`, no `skip`, no `dropWhile` — so this is not #798/#805 territory.

`RandomPipeline` seed 334:

```java
Stream.of(new Pair(3L, "a"), new Pair(1L, "b"), new Pair(3L, "a"))
    .peek(x -> SUM[0] += x.key())
    .map(Pair::key)
    .sorted(Comparator.reverseOrder())
    .peek(x -> SUM[0] += x)
    .flatMapToLong(n -> LongStream.of(n, n + 1L))
    .filter(n -> n % 3L != 0L)
    .filter(n -> n % 3L != 0L)
    .boxed().collect(Collectors.toList()).size()
```

`javac`'s class prints

```
P0334=4
P0334.peeked=14
```

and the optimized one does not load:

```
Error: Unable to initialize main class random.P0334
Caused by: java.lang.VerifyError: Operand stack underflow
```

Reproduction is the seed alone — `new RandomPipeline(334L).java("random", "P0334")` prints the class above. It is outside the 120 seeds `preservesWhatRandomPipelinesPrint` walks by default, and shows up with `-Dhone.random.pipelines=360`.

Two things narrow it down. The shape is unchanged by whatever `quarantined()` holds, so seed 334 generates the same class on `master` as on any branch; and it fails identically with the five rules of #805 removed, so it predates that fix rather than following from it. The interesting ingredients look like the `sorted()` barrier splitting the run, the `flatMapToLong` that 456 turns into a `mapMultiToLong`, and the two identical filters behind it.

Found by the randomized differential walk while working on #805.

yegor256:
Closes #811.

## What was wrong

The reported ingredients turned out to be a red herring. `sorted()`, the `flatMapToLong` and the `peek` are all incidental; the shape that breaks is just this:

```java
LongStream.of(...)
  .filter(n -> n % 2L == 0L)
  .filter(n -> n % 3L == 0L)
  .boxed()
  .collect(Collectors.toList())
```

211 splits every primitive-stream `filter` into `box` + `boxed-primitive-filter` + `unbox`, and 232 promotes it into a `p-filter` whose `bridge-input` is the wrapper while `accepted`/`returned` stay the primitive the predicate takes. When the user's own `boxed()` consumes the last compensating unbox, 411 cannot unwind the sandwich, so the fused distill still receives a `java/lang/Long` and something has to unbox it for the `(J)Z` predicates. Something did — in the wrong place.

421 prepends the head `Long.longValue()` and overwrites `accepted` with the wrapper `bridge-input` on the way. `431-dup-before-filter-distill` ran **after** it, read that overwritten `accepted`, and emitted a one-slot `dup` of the wrapper — leaving the unbox *behind* the guard, so the value every later stage saw was a `java/lang/Long` where it needed a `long`:

```
aload_0; dup; longValue; pred1(J)Z; ifne L1; return
L1: frame[java/lang/Long]
    dup2            <-- pops an empty stack
```

A lone filter got away with it (#704 — nothing came after the guard). A second filter's two-slot `dup2`, which 281 splices from the primitive `returned`, met that single one-slot reference: **Operand stack underflow**.

## The fix

The rule numbers itself `413` and goes in **ahead of** 421, so 421's unbox lands in front of the guard. The run unboxes once, every guard copies the primitive its predicate consumes, and 422 re-boxes the survivor at the tail:

```
aload_0; longValue; dup2; pred1(J)Z; ifne L1; return
L1: frame[long]
    dup2; pred2(J)Z; ifne L2; return
L2: frame[long]
    valueOf; ...
```

Two rules fall out of the pipeline as a result:

- **`412-distill-filter-returned-to-object` is deleted.** It promoted a filter distill's `returned` to the wrapper precisely so 422 would skip that re-box; with the unbox ahead of the guard the survivor is the primitive, so the promotion is wrong as well as unnecessary. #713 had already needed an `accepted` guard to keep it away from the shapes it did not fit.
- **`506-distill-filter-frame` is deleted** and `506-distill-predicate-frame` generalised to primitive predicates. The survivor at a filter's keep-label is always the value the guard duplicated one opcode earlier — which is the value the predicate consumed — so one derivation off the predicate's own parameter now covers #704, #713, #794 and #811. The old split read the primitive case off `returned`, which was only correct while 412 kept `returned` aligned with the survivor.

Two shapes the randomized walk can also reach were broken the same way and are fixed by the same change: a map fused **between** two boxed filters, whose interior `(J)J` body was handed the wrapper (`Bad type on operand stack`), and a map fused **ahead** of one, where 412's promotion declared a primitive survivor to be the wrapper. The int chain, whose one-slot widths happen to line up, failed differently again (`Type 'java/lang/Integer' is not assignable to integer`).

## Tests

- `src/test/phino/413-dup-before-filter-distill.yml` — pins the 413/421 order at the rule level: the head guard must come out a `dup2`, which is exactly what the old order could not produce.
- `optimize/streams/boxed-primitive-filters.yml` — six pipelines end to end: two and three filters, a map between two filters, a map ahead of one, and the int and double siblings. It joins the survivors rather than counting them, so their values and order are asserted, not just that the class verifies. It fails on `master` with the reported message.
- `506-distill-filter-frame-unboxed.yml` → `506-distill-predicate-frame-primitive.yml`, retargeted at the generalised rule; the two 412 packs are gone with the rule.
- `RandomPipeline` quarantines nothing.

Worth stating plainly: the deterministic pack is what keeps this caught, not the walk. #807 widened the grammar after #811 was filed, which re-deals every seed — the shape was at seed 334 of the old walk, and the re-dealt walk does not reach it inside 360 seeds at all. `RandomPipeline`'s javadoc now says so instead of carrying a stale seed number.

## Verification

Docker is unavailable in this environment, so `optimizesAsSpecifiedInYamlPack` and `preservesWhatRandomPipelinesPrint` cannot run here. I built `phino` 0.0.106 from Hackage and reproduced the plugin's `.class` → `jeo:disassemble` → `phino rewrite` (all 107 `streams/` rules, `--max-cycles 1 --max-depth 500`) → `jeo:assemble` pipeline directly, then diffed what each class printed before and after. All of the below is measured on the merge with `master` at 049a073:

- **the reported reproduction** — seed 334 of the pre-#807 walk failed verbatim (`distill_…(Ljava/lang/Long;Ljava/util/function/Consumer;)V @12: dup2`, `Attempt to pop empty stack`) and now prints `P0334=4 / P0334.peeked=14`, matching javac's class.
- **360 random seeds of the re-dealt walk** — 359/360 print identically before and after. The one exception, seed 235, fails **identically on `master`'s own rules** (same error, same bci, same synthetic method — see below), so it is not a regression from this PR. Seeds 0–119, which is what `preservesWhatRandomPipelinesPrint` walks by default, are all green.
- **all 50 end-to-end fixtures** (49 existing + the new one) — 50/50 print identically before and after. Opcode tallies are byte-identical everywhere except `long-filter-boxed.yml`, where one `dup` becomes a `dup2` and the tail `Long.valueOf` adds one `invokestatic`; its asserted `invokedynamic: 1` is unchanged. The two fixtures #813 just re-measured still match exactly: `all-ops.yml` at `invokespecial 3 / invokestatic 32 / invokevirtual 18 / invokedynamic 14 / return 19`, and `stateful-ops.yml` at `new 2 / newarray 1 / iaload 1 / iastore 1 / baload 0 / bastore 0`.
- **all 79 single-rule packs** pass (`phino rewrite` + `phino match`, same as `appliesPhinoRulesAsSpecifiedInYamlPack`).
- `mvn test` — 52 tests, 0 failures.

## One thing I found but did not fix

A **pre-existing** defect in a different rule family — the stateful 503/505/508 keep-frames, not the 4xx guard. It reproduces identically on `master`, so it is neither caused nor fixed by this PR, and it is outside the 120 seeds CI walks by default. Seed 235 of the current walk is a one-line reproduction:

```java
Stream.of(1.5, 2.25, 1.5, 8.125)
  .distinct()
  .mapToInt(Double::intValue)
  .filter(n -> n > 2)
  .asDoubleStream()
  .average().orElse(0.0)
```

`java.lang.VerifyError: Inconsistent stackmap frames at branch target 40` in `distill_502…(Ljava/util/List;Ljava/lang/Object;Ljava/util/function/IntConsumer;)V`, `Type integer ... is not assignable to 'java/lang/Object'`. The same family also takes down a `Long` variant:

```java
List.of(4L, 8L, 8L, 15L, 16L, 23L).stream()
  .distinct()
  .mapToLong(Long::longValue)
  .filter(n -> n % 2L == 0L)
  .filter(n -> n > 5L)
  .boxed().collect(Collectors.toList()).size()
```

In both, a retyping `mapToX` sits between a stateful guard and a primitive filter, and the guard's keep-frame still describes the erased `Object`. Worth its own issue; happy to file it.


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (849 >= 100); -8 for way too many hits-of-code (849 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2910](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #819, opened 2026-08-12
fix(#816): guard the lifted short-circuits too

### LINKED ISSUE #816: 454 misses every short-circuit that takes a lambda, so flatMap still fuses ahead of takeWhile
`454-flatMap-revert-on-short-circuit.phr:136-139` binds the downstream short-circuit as a raw opcode — `𝜏-op` must be `invokeinterface` or `invokevirtual` — and then tests its name against `^(limit|takeWhile|findFirst|findAny|anyMatch|allMatch|noneMatch)$`. By the time the 4xx pass runs, `111-invokedynamic-to-lambda` has already lifted every stream call that takes a lambda into a `Φ.hone.lambda`, so `takeWhile`, `anyMatch`, `allMatch` and `noneMatch` are no longer opcodes at all and the pattern never matches them. Only the three lambda-free names — `limit`, `findFirst`, `findAny` — are actually guarded, which is not what the rule's header claims.

So `455`–`458` convert the flatMap anyway and the traversal contract breaks. Measured end to end through the plugin, `Arrays.stream(NUMBERS).boxed().map(n -> n + 1L).flatMapToDouble(n -> DoubleStream.of(n, n / 2.0)).peek(x -> SUM[0] += (long) x).map(d -> d * 1.5).takeWhile(d -> d < 12.0).summaryStatistics().getCount()` counts 14 peeks before optimization and 18 after, and the optimized class carries `mapMultiToDouble` with no `flatMapToDouble` left in it; the same pipeline ending `.anyMatch(d -> d > 12.0)` counts 43 against 48, and a reference `LIST.stream().flatMap(n -> Stream.of(n, n + 1L)).peek(...).map(n -> n * 2L).anyMatch(n -> n > 12L)` counts 17 against 26. Ending the very same pipeline in `.limit(3L)` or `.findFirst()` reverts correctly, keeps `flatMapToDouble` in the class, and prints the same thing. The hang #736 was filed for therefore remains reachable, since `Stream.of(1L).flatMap(x -> Stream.iterate(x, i -> i + 1)).takeWhile(i -> i < 5)` ends in a lambda-taking short-circuit.

Add a sibling pattern to `454` in which `𝜏-short` is a `Φ.hone.lambda` whose `stream` binding carries `method ↦` one of the same seven names, so the guard covers the lifted form as well as the raw opcode — the same dual route the rule already takes for `limit`.

Found by `OptimizeMojoTest#preservesWhatRandomPipelinesPrint` at 360 seeds; `new RandomPipeline(196L).java("random", "P0196")` prints the class.

yegor256:
Closes #816.

## The defect

`454-flatMap-revert-on-short-circuit.phr` binds the downstream short-circuit as a raw opcode — `𝜏-op` must be `invokeinterface` or `invokevirtual` — and then tests its method operand against `^(limit|takeWhile|findFirst|findAny|anyMatch|allMatch|noneMatch)$`. By the time the 4xx pass runs, `111-invokedynamic-to-lambda` has folded every stream call that takes a lambda into a `Φ.hone.lambda`, so `takeWhile`, `anyMatch`, `allMatch` and `noneMatch` are no longer opcodes at all and that pattern never reaches them. Only the three lambda-free names were actually guarded, `455`–`458` converted the `flatMap` anyway, and the traversal contract broke.

## The fix

A sibling rule, `454-flatMap-revert-on-lifted-short-circuit.phr`, carrying the identical guard with `𝜏-short` bound as a `Φ.hone.lambda` whose `stream` block holds the operator name. Nothing between `111` and `701` claims those four operators, so the lambda is still intact at 454 and one pattern covers them. Both files keep the same seven-name list rather than splitting it three/four, so neither half can drift; a name that cannot occur in one shape simply never matches there. The pair sorts ahead of `455` (`-lifted-` < `-short-` < `.phr`), and the reverted `invokeinterface` carries the same `reverted ↦ Φ.true` marker, so the `111 → 454 → 701` cycle still cannot spin.

Headers in the existing `454` and the README section that claimed all seven names were covered are corrected to describe the pair.

## Tests

- `src/test/phino/454-flatMap-revert-on-lifted-short-circuit.yml` — single-rule pack: a `flatMap` lambda followed by a lifted `takeWhile` reverts to the native pair, and the `takeWhile` lambda is left untouched for `701`.
- `flatmap-lifted-short-circuit.yml` — the `limit` fixture's `takeWhile` twin on an unbounded `Stream.iterate`. **Without the new rule this hangs**, timing out at 180s: the #736 hang, still reachable through a lambda-taking short-circuit.
- `flatmap-lifted-any-match.yml` — a finite source where the symptom is wasted work. The inner streams hold three elements and the match lands on the *first* of one of them, which is what makes the difference visible; cancellation between outer elements works either way, so a match on an inner stream's last element would hide the defect. **Without the new rule the optimized run prints `after 15` instead of `after 13`** — the `mapMulti` adapter finishing an inner stream after `anyMatch` was already satisfied.

Both fixtures' `after` tallies were measured from the optimized bytecode, not predicted: in the `anyMatch` case `invokedynamic` falls 4 → 3 (the peek and map lambdas become the fused adapter's single `BiConsumer`) while `invokeinterface` stays at 5 (`Stream.peek` + `Stream.map` collapse into one `Stream.mapMulti`, and the adapter body's own `Consumer.accept` takes the freed slot), and the class keeps its native `Stream.flatMap` call site beside the fused `Stream.mapMulti`.

Verified against the real `phino` 0.0.106 on the host: 82/82 single-rule packs pass, and both new fixtures pass with the rule and fail without it as described above.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MqnMxs1pLym2wm2oLSpYq9)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -341,6 +341,18 @@ Because a `mapMulti` cannot honour cancellation, the cure mirrors the
   native call whenever the same method body reaches a `limit`, `takeWhile`,
   `findFirst`, `findAny`, `anyMatch`, `allMatch`, or `noneMatch` downstream of
   it, so those `flatMap`s stay unfused.
+Covering those seven names takes two rules, because by the time the 4xx pass
+  runs they no longer share a shape (#816):
+  `limit`, `findFirst` and `findAny` take no lambda and are still plain opcodes,
+  while `takeWhile`, `anyMatch`, `allMatch` and `noneMatch` each take one and
+  have therefore been folded into a `Φ.hone.lambda` by
+  `111-invokedynamic-to-lambda`.
+`454-flatMap-revert-on-short-circuit` matches the opcode shape and its sibling
+  `454-flatMap-revert-on-lifted-short-circuit` the folded one, with the same
+  guard and the same seven-name list in both;
+  until the second one existed the four lambda-taking names went unguarded, the
+  `flatMap` was fused anyway, and the hang above stayed reachable through
+  `.takeWhile(i -> i < 5)`.
 
 ## Why `skip` Is Only Fused on Sequential Pipelines
 
--- a/src/main/resources/org/eolang/hone/rules/streams/4xx/454-flatMap-revert-on-lifted-short-circuit.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/4xx/454-flatMap-revert-on-lifted-short-circuit.phr
@@ -1,0 +1,152 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# The lifted half of the guard `454-flatMap-revert-on-short-circuit` opens; it
+# shares that rule's number and sorts ahead of it (`-lifted-` < `-short-`),
+# both ahead of 455. Read the sibling's header first: everything it says about
+# WHY a flatMap has to stay native once a short-circuiting operator is
+# reachable downstream of it (#736), about the dual positive route, and about
+# the `reverted ↦ Φ.true` marker that keeps the 111 → 454 → 701 cycle from
+# spinning, holds here word for word. Exactly one binding differs — the shape
+# the downstream short-circuit is matched in.
+#
+# The sibling binds that short-circuit as a raw `Φ.jeo.opcode.invokeinterface`
+# or `invokevirtual` and tests its method operand against the seven names. In
+# practice that reaches only the three which take no lambda — `limit`,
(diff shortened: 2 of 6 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (347 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2894](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #817, opened 2026-08-12
505 never narrows the Object item for a capturing map, so distinct() ahead of one fails verification


yegor256:
`505-distill-state-item-cast.phr:94` will only splice its `checkcast` when the consuming body's descriptor matches `^\(L[^;]+;\)` — one reference parameter and nothing else. A capturing lambda's desugared body carries its captures as extra parameters, so a `map(p -> new Pair(p.key() + by, p.text()))` that captures one `long` desugars to `(JLPair;)LPair;`, the test fails, no cast is spliced, and the `Object` item that `distinct()` or `skip()` erased reaches an `invokestatic` typed for the element.

The class then fails verification at initialization. `PAIRS.stream().distinct().map(p -> new Pair(p.key() + by, p.text())).forEach(x -> SUM[0] += x.key())`, with `by` a captured method argument and `Pair` a small value class, dies with `java.lang.VerifyError: Bad type on operand stack`, `Type 'java/lang/Object' (current frame, stack[2]) is not assignable to 'random/P0224$Pair'`, inside the synthesised `distill_5027019`. `skip(1L)` in place of `distinct()` fails identically. The non-capturing `map(p -> new Pair(p.key() + 1L, p.text()))` runs clean, and so does the same capturing map with no stateful head, so the trigger is exactly a stateful head plus a capturing reference map; a `.toList().size()` terminal fails too, so the terminal is not part of it. #788 repaired the neighbouring adjacency gap and said of this guard that it "can stay as it is" — this is the shape it does not reach.

Let 505 read the item from the last parameter of the descriptor rather than requiring it to be the only one: widen the match at `:94` to admit a capture prefix, and change the `𝑒-cast-class` sed at `:97-103` to take the rightmost reference type instead of the first, the same descriptor walking that `112` and `113` already do with their `[IJD]` alternation.

Found by `OptimizeMojoTest#preservesWhatRandomPipelinesPrint` at 360 seeds; `new RandomPipeline(224L).java("random", "P0224")` prints the class.

yegor256:
@yegor256 could you take a look when you have a moment? Happy to clarify anything in the report that is unclear.

0crat:
[...]

PR #820, opened 2026-08-12
fix(#817): take the item from the descriptor, not from its last L

yegor256:
Follow-up to #818, which merged before this could be pushed. Copilot's review
on that PR was right, and the defect is worse than the comment suggests.

## What is broken on master

`509-distill-state-item-cast-parked.phr` reads the class to cast to with

```
s/.*L([^;]*);\).*/$1/
```

which finds the **last `L` in the descriptor**, not the `L` that opens its last
parameter. Any element whose internal name carries an `L` of its own is read
from that inner letter. Ordinary code:

```java
private static final List<Long> LIST = List.of(4L, 8L, 8L, 15L);
LIST.stream().distinct().map(n -> n + by).forEach(n -> sum[0] += n);
```

desugars to `(JLjava/lang/Long;)Ljava/lang/Long;`, the walk lands on the `L` of
`Long`, and the optimized class carries

```
44: checkcast  #150   // class ong
47: invokestatic #152 // Method lambda$shifted$0:(JLjava/lang/Long;)Ljava/lang/Long;
```

and dies at initialization with `java.lang.NoClassDefFoundError: ong` —
measured through the plugin against master's rule, then again with this fix,
where the same site becomes `checkcast java/lang/Long` and the class prints
`shifted=36`.

`Long` is not special. `List` yields a three-letter suffix, `ArrayList` and
`LinkedList` the same, and so does any value class holding an `L`. `Integer`,
`Double` and `String` happen to carry no `L`, which is exactly why the pack and
the fixture added in #818 passed — every type in them was L-free.

## The fix

Walk the descriptor's **delimiters**, never its letters: isolate the parameter
list, drop its final `;`, cut everything through the last `;` that is left
(which ends the second-to-last reference capture, if any), and strip the
primitive captures and the `L` off what remains.

```yaml
- meta: 𝑒-cast-class
  function: sed
  args:
    - 𝑒-invoke-signature
    - '"s/^[(]//"'
    - '"s/[)].*$//"'
    - '"s/;$//"'
    - '"s/^.*;//"'
    - '"s/^[IJD]*L//"'
```

## Tests

The gap that let this through was that no committed test used an L-bearing type,
so both new tests do.

* `src/test/phino/509-distill-state-item-cast-parked.yml` gains a second
  distill-lambda (`op2`) whose descriptor is
  `(JLjava/util/LinkedList;LA$Level;)LA$Level;` — an L-bearing capture and an
  L-bearing item. The expected pattern pins the cast to `A$Level`; the old walk
  produced the suffix of `Level` and cannot pass it.
* `src/test/resources/org/eolang/hone/optimize/streams/stateful-closure-boxed.yml`
  runs the `List<Long>` pipeline above end to end: `checkcast` 0 → 3 (the
  guard's `KeySetView`, the capture's `Long`, the item's `Long`) and
  `The result is 36`.

Verified against the real `phino` 0.0.106 on the host:

* A 16-case table over the guard and the extraction — L-bearing items
  (`java/lang/Long`, `java/util/List`, `java/util/LinkedList`, `A$Level`),
  L-bearing captures, primitive-capture prefixes, a class name beginning with a
  primitive letter (`LJar;`), and the declined shapes (`Object` item, no
  reference item, array parameter, non-`Object` bridge-input) — all extract the
  correct class. Four of the sixteen fail on master.
* Both fixtures and the pack pass, and the nine capturing shapes from #818's
  verification still print identically before and after optimization.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HMBVSsm14kUEpARD63MKdF)_

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/5xx/509-distill-state-item-cast-parked.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/5xx/509-distill-state-item-cast-parked.phr
@@ -118,12 +118,20 @@ when:
 where:
   - meta: 𝜏-cast
     function: random-tau
-  # The rightmost reference of the parameter list, which is the item: the
-  # greedy prefix walks past every capture to the last "L…;" that still ends at
-  # the closing parenthesis, so a capture whose own class name sits earlier in
-  # the descriptor cannot be picked by mistake.
+  # The LAST parameter of the descriptor, which is the item. The walk is over
+  # the descriptor's own delimiters, never over its letters: isolate the
+  # parameter list, drop its final ";", cut everything through the last ";" that
+  # is left (which ends the second-to-last reference capture, if any), and strip
+  # the primitive captures and the "L" off what remains. Searching for the last
+  # "L" instead would find one INSIDE a class name — "Ljava/util/List;" and
+  # "LA$Level;" each carry one — and name the suffix after it instead of the
+  # class, splicing a checkcast onto a type that does not exist.
   - meta: 𝑒-cast-class
     function: sed
     args:
       - 𝑒-invoke-signature
-      - '"s/.*L([^;]*);\\).*/$1/g"'
+      - '"s/^[(]//"'
+      - '"s/[)].*$//"'
+      - '"s/;$//"'
+      - '"s/^.*;//"'
+      - '"s/^[IJD]*L//"'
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (131 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2898](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #823, opened 2026-08-12
fix(#821): dup the item for a filter behind a capturing map

### LINKED ISSUE #821: A filter behind a capturing map never gets its guard dup, so the class fails verification
`281-insert-dup-before-filter.phr:79-88` alternates the producer ahead of a filter over `map`, `o-filter`, `p-filter` and `peek`. A capturing map is none of those — `112` and `114` lift it to `Φ.hone.c-map` — so when a plain filter follows one, no `Φ.hone.dup` is spliced in front of the guard.

The predicate then eats the only copy of the element. `PAIRS.stream().map(p -> new Pair(p.key() + by, p.text())).filter(p -> p.key() > 1L).forEach(...)`, with `by` a captured method argument and `Pair` a small value class, fuses to a distill whose body reads `aload_1; <fetch the capture>; aload_1; invokestatic lambda$pipe$0:(JLPair;)LPair;; invokestatic lambda$pipe$1:(LPair;)Z; ifne 31; return; 31: aload_2; dup_x1; Consumer.accept` — no `dup` anywhere, and a `dup_x1` at the keep label that wants two entries on a stack that is empty. `503-distill-predicate-frame` meanwhile writes a full frame there declaring `stack [Pair]`, so the class dies at load with `java.lang.VerifyError: Inconsistent stackmap frames at branch target 31`, "Current frame's stack size doesn't match stackmap". The trigger is exactly that pair: the same filter behind a NON-capturing map is fine, a capturing map ahead of a CAPTURING filter is fine, a capturing map with a `takeWhile` and no filter is fine, and a capturing filter ahead of a plain filter is fine. It is the failure family of #790 and #804, one producer further along.

Add `c-map` to the alternation at `:81-85`. A `c-map` carries `captured` and `state-acc` beyond the attributes the `𝜏-first` block binds at `:23-33`, so that block needs a `𝐵-` wildcard to match one — or the shape wants a sibling rule of its own, the way `282` and `283` already are.

Found by `OptimizeMojoTest#preservesWhatRandomPipelinesPrint` at 360 seeds; `new RandomPipeline(224L)` prints the class, and it is the last of the 360 still failing on master (f0d4e472). It reproduces with the identical error at 049a0735, so it is not a regression from the recent fixes — for the `distinct()`-headed shapes it was simply masked by the defect #817 has since fixed.

yegor256:
Closes #821.

## What was wrong

A filter guard spends the item on its predicate, so it needs a throwaway copy in front of it. `281` splices that `Φ.hone.dup` when a `map`, an `o-filter`/`p-filter` or a `peek` leads, `283` when a stateful guard does — but a **capturing** map is neither as far as their patterns go. `112` lifts it to a `Φ.hone.c-map` carrying `state-acc`, `body-acc` and `captures` ahead of `class` and no `opcode` at all, while `281` spells its predecessor out binding for binding, so a plain filter behind one reached `4xx` with no guard. `401` fused the pair into a distill anyway, the predicate ate the only copy of the mapped item, and `503` framed the keep label with an element the stack no longer had.

Reproduced exactly as reported, down to the branch target:

```
Caused by: java.lang.VerifyError: Inconsistent stackmap frames at branch target 31
  Location: closure/CMapFilter.distill_5021652(Ljava/util/List;Lclosure/CMapFilter$Pair;Ljava/util/function/Consumer;)V @27: ifne
  Reason:   Current frame's stack size doesn't match stackmap.
  Stackmap Frame: stack: { 'closure/CMapFilter$Pair' }
```

## The fix

One more name in `283`'s alternation:

```yaml
- eq: [𝜏-one, 'c-map']
```

The issue suggests giving `281` a `𝐵-` wildcard or writing a sibling rule; `283` is already that sibling. The `𝐵-first-head` that #804 gave it to reach a `dropWhile` absorbs the c-map's three accumulator bindings unchanged, and a capturing map is stateful in the sense that matters here — `316` folds it into a stateful distill. Its `dup` opcode comes off the filter's `bridge-input`, which is what a c-map wants too: `112` only lifts a `Stream.map` whose element stays a reference, so the follower is always a one-slot `o-filter`, never a `dup2`. The header paragraph records that, and why a c-map cannot reach `281` or `282` (`241`-`243` pin `Φ.hone.map` literally, so no `type`/`transform` is ever spliced behind a capturing map).

No double dup: after insertion the two are no longer adjacent, and the fused run's `start` becomes `dup`, so `413` still declines it.

## Tests

* `src/test/phino/283-insert-dup-before-filter-after-capturing-map.yml` — a drained c-map (empty `captures`, both accumulators holding the peeled unit) ahead of an `o-filter`; asserts the spliced `Φ.hone.dup` with `opcode ↦ "dup"`. Before the fix the same input comes out with no dup at all.
* `src/test/resources/org/eolang/hone/optimize/streams/closure-map-filter.yml` — the reported shape end to end: a `long` captured in the map, a small value class as the element, three pipelines (the plain pair, the same pair behind a `forEach`, and a two-filter variant so `283` has to leave the second pair for `281`). `invokedynamic` 8 → 4, four `ifne` guards, and the printed line proves the mapped item — not the original — is what survives the guard.

## Verification

* `mvn -Pdeep test`: **153 run, 0 failures, 17 skipped** (the skipped ones are Docker-gated). All 83 single-rule packs pass against phino 0.0.106 on the host.
* All **55** `optimize/**.yml` fixtures compiled into one project and optimized in a single build: every class prints byte-identical output before and after optimization. With the c-map clause removed the run is identical except for the new fixture, which fails with the `VerifyError` above — so the change is inert on every shape the suite already covered.
* **360 random pipelines** (`RandomPipeline`, seeds 0-359), same before/after oracle: 359 identical.

Two notes on the reproduction, since the issue attributes it to the random walk:

* `RandomPipeline`'s grammar has no capturing lambdas — every lambda in it reads statics or its own parameter, so no factory descriptor carries a capture. `0` of the 720 rewritten `.phi` files in the 360-seed run contain a `c-map`, and seed 224 on `f0d4e47` walks to `INTEGERS.stream().map(Integer::doubleValue).peek(…).dropWhile(…).peek(…).peek(…).map(…).map(…).collect(teeing(…))`, which has no filter in it and passes. The new fixture is therefore hand-written to the shape the issue describes rather than lifted from a seed.
* One seed does still fail, `P0235` — `Stream.of(1.5, 2.25, 1.5, 8.125).distinct().mapToInt(Double::intValue).filter(n -> n > 2).asDoubleStream().average()`. It is **not** this defect and **not** a regression: rewriting its `.phi` with and without the c-map clause gives byte-identical output, its guard *is* dupped (the pragma run after `2xx` is `distinct, map, dup, p-filter, unbox`), and the keep frame disagrees with the stack about the element's *type* rather than its presence (`stack: { integer }` against `Object`). That looks like the boxed-primitive relay family of #811, and probably wants an issue of its own.

---
_Generated by [Claude Code](https://claude.ai/code/session_012zpUE19ynLwsbMxw5ijyq8)_

### DIFF
--- a/src/main/resources/org/eolang/hone/rules/streams/2xx/283-insert-dup-before-filter-after-stateful.phr
+++ b/src/main/resources/org/eolang/hone/rules/streams/2xx/283-insert-dup-before-filter-after-stateful.phr
@@ -7,11 +7,11 @@
 # Φ.hone.dup when the filter follows a map or another filter, but its
 # pattern is shaped for the map/filter pragma (opcode + target bindings)
 # and so never matches a stateful predecessor. A filter that follows a
-# distinct, skip or dropWhile — a stateful guard that leaves a single
-# (un-duplicated) item — would therefore be left without its dup, its
+# distinct, skip, dropWhile or capturing map — a stateful pragma that leaves a
+# single (un-duplicated) item — would therefore be left without its dup, its
 # predicate would eat the only item, and the keep-label frame would find an
 # empty stack. This rule plugs that gap: when 𝜏-first is one of those
-# stateful guards and 𝜏-second is a filter, it splices the same Φ.hone.dup
+# stateful pragmas and 𝜏-second is a filter, it splices the same Φ.hone.dup
 # between them. After
 # insertion the two are no longer adjacent, so it fires once per site.
 #
@@ -26,6 +26,25 @@
 # "Inconsistent stackmap frames" (#804). 𝐵-first-head absorbs whatever the
 # guard puts between `φ` and `class` — none of it is anything this rule reads
 # — so the ordering a fourth guard picks will not matter either.
+#
+# The capturing map is the fourth (#821). A `c-map` is pointwise, so 281 is
+# where it would belong by nature, but 112 gives it `state-acc`, `body-acc`
+# and `captures` ahead of `class` and no `opcode` at all, and 281 spells its
+# predecessor out binding for binding — so a plain filter behind a capturing
+# map reached 4xx with no guard, 401 fused it into a distill, the predicate
+# ate the only copy of the mapped item, and 503's keep-frame promised an
+# element the stack no longer had. It is stateful in the sense that matters
+# here — 316 folds it into a stateful distill, and its bindings are exactly
+# the ones this rule's 𝐵-first-head was written to absorb. `map(p -> new
+# Pair(p.key() + by, p.text())).filter(p -> p.key() > 1L)`, with `by`
+# captured, is the shape; streams/closure-map-filter.yml runs it.
+#
+# A c-map cannot reach the other two rules of this family. 241-243 pin
+# Φ.hone.map literally, so no `type` or `transform` is ever spliced behind a
+# capturing map and 282 has nothing to say about one; and 112 only lifts a
+# Stream.map, whose element stays a reference, so the filter that follows is
+# always an o-filter over one stack slot — the `dup` this rule reads off the
+# filter's `bridge-input`, never a `dup2`.
 name: insert-dup-before-filter-after-stateful
 pattern: |
   ⟦
@@ -82,6 +101,7 @@ when:
         - eq: [𝜏-one, 'distinct']
         - eq: [𝜏-one, 'skip']
         - eq: [𝜏-one, 'dropWhile']
+        - eq: [𝜏-one, 'c-map']
     - or:
         - eq: [𝜏-two, 'o-filter']
         - eq: [𝜏-two, 'p-filter']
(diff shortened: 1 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (198 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2902](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #824, opened 2026-08-12
test(#822): pin filter().boxed().mapToX() with a fixture

### LINKED ISSUE #822: no fixture pins filter().boxed().mapToX(), the keep-frame shape that broke in #704 and again before #811
Nothing under `src/test/resources/org/eolang/hone/optimize/streams/` pins a primitive-domain
filter whose `boxed()` is followed directly by a trailing `mapToX`. The three fixtures that come
close all stop short of it: `boxed-primitive-filters.yml:68` continues into a reference
`map(String::valueOf)` before collecting, `long-filter-boxed.yml:39` ends at `collect`, and the
filter in `all-ops.yml:20` is a reference-stream filter sitting far ahead of that file's
`boxed()`/`mapToX` run. So the whole `filter -> boxed -> mapToX` corner is covered by no fixture
at all.

That corner is exactly where the filter guard's keep-label frame has failed twice. It failed as
#704, and it failed again on 049a0735 in deep run 31597774925: `RandomPipeline` seed 9 emitted
`Arrays.stream(DECIMALS).filter(d -> d > 1.0).boxed().mapToInt(Double::intValue).count()` and the
class died at link time with `VerifyError: Inconsistent stackmap frames at branch target 12` — the
frame declared `integer` while the stack held the `java/lang/Double` the guard's `dup` had kept —
because the `506-distill-filter-frame.phr` of that commit derived the frame from the distill's
`returned`, which `4xx/451-fuse-distill-unbox.phr:79` overwrites with the fused `mapToX` output.
The #811 rewrite in 5246388f deleted that rule and now derives the frame from the predicate's own
parameter (`5xx/506-distill-predicate-frame.phr:108`), and the deep job on that commit is green,
so the shape works today. It is simply unpinned: the only thing covering it is a seeded random
walk, and a seed that happens to reach a shape is not a regression test for it.

Add one fixture — say `optimize/streams/boxed-filter-tail.yml` — holding a
`DoubleStream.filter(...).boxed().mapToInt(...).count()` together with its `IntStream` and
`LongStream` siblings, with the usual `log` lines and `after` opcode counts asserting the single
fused `mapMultiToInt`. Then a future regression in the keep-label frame fails by fixture name
instead of by seed number.

yegor256:
Closes #822.

Nothing under `src/test/resources/org/eolang/hone/optimize/streams/` held a primitive-domain
filter whose `boxed()` is consumed directly by a trailing `mapToX`, so the corner where the
filter guard's keep-label frame failed as #704 and again on 049a0735 (deep run 31597774925,
seed 9) was covered only by a seeded walk of `RandomPipeline` — and a seed that happens to
reach a shape is not a regression test for it.

### What is added

One pack, `optimize/streams/boxed-filter-tail.yml`, with four pipelines in one class:

| pipeline | shape | asserts |
| --- | --- | --- |
| `kept` | `Arrays.stream(DECIMALS).filter(d -> d > 1.0).boxed().mapToInt(Double::intValue).count()` | the seed-9 reproduction: 4 survivors |
| `truncated` | the same shape summed rather than counted | the values the fused body produced: `{1, 2, 8}` -> 11 |
| `longs` | `LongStream…filter…boxed().mapToInt(Long::intValue).sum()` | the `long` keep-frame sibling -> 20 |
| `ints` | `IntStream…filter…boxed().mapToLong(Integer::longValue).sum()` | the `integer` keep-frame sibling -> 21 |

The `IntStream` sibling ends in `mapToLong` on purpose: `Integer` -> `int` cancels the `boxed()`
round-trip, 411 unwinds the sandwich, and the run returns to the primitive domain as a plain
`IntStream.mapMulti` — a shape with no wrapper in it at all, and not the one this pack is for.

Each pipeline collapses into a single `mapMultiToX`, which is what the after-count asserts: the
two invokedynamic each of them starts with — a predicate and a methodref — become one, 8 -> 4.
The header comment walks the lowering (103/205, 203, 206, 211, 221, 232, 304, 413, 421, 422,
451, 501, 506) and names why `451` overwriting the distill's `returned` is what used to poison
the keep-frame.

### How it was checked

Ran with phino 0.0.106 on the host, both ways:

* against today's rules the class verifies, prints `The result is 4/11/20/21`, and lowers to
  three `Stream.mapMultiToInt` plus one `Stream.mapMultiToLong`, 4 invokedynamic in all;
* against the rule set of 049a0735 — the pre-#811 `506-distill-filter-frame.phr`, which read the
  frame off `returned` — the same class dies at link time with
  `VerifyError: Inconsistent stackmap frames at branch target 12 ... Type 'java/lang/Double' is
  not assignable to integer`, the failure the issue reports.

So a future regression in the keep-label frame now fails by fixture name instead of by seed
number.

https://claude.ai/code/session_01NkCAGHFJmWYL2wAEB7WzvP

---
_Generated by [Claude Code](https://claude.ai/code/session_01NkCAGHFJmWYL2wAEB7WzvP)_

### DIFF
--- a/src/test/resources/org/eolang/hone/optimize/streams/boxed-filter-tail.yml
+++ b/src/test/resources/org/eolang/hone/optimize/streams/boxed-filter-tail.yml
@@ -1,0 +1,101 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+# The corner #822 reports: a primitive-domain filter whose boxed() is consumed
+# directly by a trailing mapToX, with nothing in between. The three packs that
+# come close all stop short of it — boxed-primitive-filters.yml continues into a
+# reference map before collecting, long-filter-boxed.yml ends at collect, and the
+# filter in all-ops.yml is a reference-stream filter sitting far ahead of that
+# file's boxed()/mapToX run — so until now the only thing covering the shape was
+# a seeded walk of RandomPipeline, and a seed that happens to reach a shape is
+# not a regression test for it.
+#
+# It is the filter guard's keep-label frame that has failed here, twice: as #704,
+# and again on 049a0735 (deep run 31597774925), where seed 9 emitted the `kept`
+# pipeline below and the class died at link time. The lowering that gets there:
+# 103 lifts the trailing `Double::intValue` methodref into a wrapper lambda and
+# 205 recognizes it as an unbox; 203 lifts the primitive filter, 206 turns the
+# user's boxed() into a box, 211 splits the filter into box +
+# boxed-primitive-filter + unbox, 221 cancels that compensating unbox against
+# the user's own box, and 232 promotes the filter into a p-filter whose
+# bridge-input is the wrapper Ljava/lang/Double; while accepted/returned stay the
+# primitive D its predicate takes. Because the user's boxed() is what consumed
+# the unbox, 411 never unwinds the sandwich and the distill 304 mints still
+# receives the wrapper: 421 prepends the head Double.doubleValue(), 413 puts the
+# guard's dup2 behind it so every copy is the primitive, 422 re-boxes the
+# survivor at the tail, and 451 fuses the trailing unbox — the user's mapToInt —
+# into that same distill, overwriting `returned` with its `int` output on the
+# way. 501 lowers the whole run to a single Stream.mapMultiToInt.
+#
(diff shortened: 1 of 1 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (101 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2906](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #849, opened 2026-08-31
#511: cover the file that carries only its header

### LINKED ISSUE #511: `records.get(0)` in `CSV` constructor throws `IndexOutOfBoundsException` on header-only CSV file
## What happens

In [`src/main/java/org/eolang/hone/CSV.java`](https://github.com/objectionary/hone-maven-plugin/blob/master/src/main/java/org/eolang/hone/CSV.java#L57-L62), the private constructor reads headers from the first record:

```java
private CSV(final List<CSVRecord> records) {
    this(
        new ListOf<>(records.get(0).toMap().keySet()),
        CSV.rows(records)
    );
}
```

This constructor is invoked from `CSV(Path)` via `CSV.from(path)` (lines 168-184), which parses the file with `CSVFormat.DEFAULT.builder().setHeader().setSkipHeaderRecord(false)`. When `setHeader()` is called with no arguments, commons-csv uses the first line of the file as the header definition and **does not** return it as a data record. Therefore, for a CSV that contains only a header line and no data rows, `records` is **empty**, and `records.get(0)` throws `IndexOutOfBoundsException`.

This case is reachable in practice. `Summary.collect()` (`Summary.java:68`) walks the project for `hone-statistics.csv` files and constructs a `CSV` from each. If the optimization step produced no changed `.class` files for a module, a `hone-statistics.csv` with only a header line can legitimately exist. The callers `SummaryMojo` (line 35) and `OptimizeMojo` (line 376) guard with `if (stats.toFile().exists())` but file existence does not imply at least one data row.

No test in `CSVTest` covers the header-only or empty-file case.

## What should happen

Derive the header set from `CSVParser.getHeaderMap().keySet()` (available in commons-csv 1.14) rather than from `records.get(0)`. That works correctly even when there are zero data records. Result: a `hone-statistics.csv` containing only a header line is parsed as an empty CSV with the correct columns, and `Summary.collect()` no longer crashes when any individual module produced no statistics.

Add a test case that constructs a `CSV` from a header-only file and asserts it has zero rows.

morphqdd:
The crash itself is gone from master: the fix for #512 rewrote `from` to read the names through `parser.getHeaderNames()`, which answers for a file that carries a header and no rows, where `records.get(0)` threw `IndexOutOfBoundsException`.

What #511 also asked for is still missing, and that is what this adds: nothing covered the header-only case. One test reads such a file and counts nothing instead of failing. The other adds it to a full one and checks the columns survive, since `add` keeps the columns of the left side, so an empty file first in the reduce would have emptied the summary.

Closes #511

### DIFF
--- a/src/test/java/org/eolang/hone/CSVTest.java
+++ b/src/test/java/org/eolang/hone/CSVTest.java
@@ -61,7 +61,7 @@ final class CSVTest {
         MatcherAssert.assertThat(
             "an empty or non-numeric 'Changed' cell must not break the count (see #864)",
             new CSV(path).count("Changed", CSV::positive),
-            Matchers.is(2)
+            Matchers.is(1)
         );
     }
 
@@ -70,7 +70,11 @@ final class CSVTest {
         final Path path = temp.resolve("empty.csv");
         Files.write(
             path,
-            "ID,Before,After,Changed,LinesPerSec\n".getBytes(StandardCharsets.UTF_8)
+            String.join(
+                System.lineSeparator(),
+                "ID,Before,After,Changed,LinesPerSec",
+                ""
+            ).getBytes(StandardCharsets.UTF_8)
         );
         MatcherAssert.assertThat(
             "a file with a header and no rows must count nothing, not fail",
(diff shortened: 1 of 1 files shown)


yegor256:
xcop check is failing here, not merging until it's green.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6777](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #853, opened 2026-09-01
#840: refuse a rules pattern that matches nothing

### LINKED ISSUE #840: An empty hone.rules selection silently falls back to all built-in rules, including the code-corrupting demo rule
## Problem

When `hone.rules` selects **no rules at all** (a typo like `strrams/*`, an all-negating pattern like `!*`, or a pattern matching nothing), `rulesAsString()` returns an empty string:

- `src/main/java/org/eolang/hone/OptimizeMojo.java:618-626` — `String.join(" ", new Rules(this.rules).yamls())`; an empty `yamls()` yields `""`.
- The empty `RULES` is exported to the scripts (`OptimizeMojo.java:539` in the Docker path, `OptimizeMojo.java:665` in the no-Docker path).

Then `entry.sh` treats the empty variable as "unspecified" and silently substitutes **every built-in YAML rule**:

```sh
# src/main/resources/org/eolang/hone/scaffolding/entry.sh:131-133
if [ -z "${RULES}" ]; then
fi
```

The fallback picks up `none.yml` **and `33-to-42.yml`**, the demo rule whose own javadoc warns the opposite of nothing:

> `33-to-42.yml` — demo rule ... your code most definitely will be corrupted

(`Rules.java:88-100`, `OptimizeMojo.java` javadoc around rule selection). So a user who tried to *disable* optimization (or typo'd the pattern) ends up applying the demo rule that corrupts bytecode — with a green build and no warning.

## How to reproduce

```sh
mvn org.eolang:hone-maven-plugin:optimize -Dhone.rules='strrams/*'
# RULES becomes empty → entry.sh:132 falls back to all built-in *.yml → 33-to-42.yml is applied
```

## Expected behavior

An empty selection must be a **hard error** (or at least a loud warning + abort) with a message like `No rules matched the hone.rules pattern 'strrams/*'`, not a silent fallback to a different — and in this case dangerous — set of rules. `entry.sh` should not invent a rule set when `RULES` is empty; an explicitly configured but unmatched pattern is different from an unset variable.

## Details

- `src/main/java/org/eolang/hone/OptimizeMojo.java:138` — `@Parameter(property = "hone.rules", defaultValue = "none")`
- `src/main/java/org/eolang/hone/OptimizeMojo.java:539, 665` — `RULES` env export
- `src/main/java/org/eolang/hone/OptimizeMojo.java:618-626` — `rulesAsString()` may be empty
- `src/main/resources/org/eolang/hone/scaffolding/entry.sh:131-133` — silent fallback
- `src/main/java/org/eolang/hone/Rules.java:88-100` — `yamls()` may be empty for a given pattern

morphqdd:
A pattern that matched no rule left `rulesAsString()` empty, `entry.sh` read
that as "nothing was configured" and put every built-in rule back, including
`33-to-42.yml`, whose own javadoc says the code will be corrupted. A typo in
`hone.rules` therefore turned into a green build with the demo rule applied.

`Rules.yamls()` refuses an empty match now, quoting the pattern and listing
what is available, so the empty value never reaches the script.

`discoversNothingWithSuffix` asserted the old answer, so it now asserts the
refusal under a name that says so, and a second test checks that the message
quotes the pattern.

Closes #840

### DIFF
--- a/src/main/java/org/eolang/hone/Rules.java
+++ b/src/main/java/org/eolang/hone/Rules.java
@@ -51,6 +51,11 @@ final class Rules {
     private final Map<Pattern, Boolean> patterns;
 
     /**
+     * The patterns as the user wrote them, for the message of an empty match.
+     */
+    private final String text;
+
+    /**
      * Creates a rule manager that includes all rules.
      */
     Rules() {
@@ -64,6 +69,7 @@ final class Rules {
      */
     Rules(final String ptns) {
         this.patterns = Rules.regexs(ptns);
+        this.text = ptns;
     }
 
     @Override
@@ -96,6 +102,14 @@ final class Rules {
             files.add(name);
         }
         Logger.info(this, "Found %d rule(s) by: %s", files.size(), this.patterns);
+        if (files.isEmpty()) {
+            throw new IllegalStateException(
+                String.format(
+                    "No rules matched the hone.rules pattern '%s', while the ones available are: %s",
+                    this.text, Rules.discover()
+                )
+            );
+        }
         return files;
     }
 
--- a/src/main/java/org/eolang/hone/Rules.java
+++ b/src/main/java/org/eolang/hone/Rules.java
@@ -106,7 +106,7 @@ final class Rules {
             throw new IllegalStateException(
                 String.format(
                     "No rules matched the hone.rules pattern '%s', while the ones available are: %s",
-                    this.text, Rules.discover()
+                    this.text, String.join(", ", Rules.discover())
                 )
             );
         }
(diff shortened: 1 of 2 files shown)


yegor256:
Thanks for this — merged. Making `Rules.yamls()` throw on an empty match (instead of silently falling through to "use every rule") closes the real hole in #840: a typo in `hone.rules` no longer resurrects `33-to-42.yml` on a green build. Good, focused change with clear test coverage for both the refusal and the message content.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (35) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6855](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #855, opened 2026-09-02
#836: do nothing when there is no XMIR to rewrite

### LINKED ISSUE #836: optimize fails on a phantom /.xmir task when no XMIR files are found
## Problem

When `find` in `rewrite.sh` returns no XMIR files (e.g. `hone.includes` selects nothing, or an empty project), the script builds a **phantom task** for a non-existent `/.xmir` file and the whole `optimize` goal fails with a confusing error instead of skipping quietly.

Mechanism (`src/main/resources/org/eolang/hone/scaffolding/rewrite.sh:272-288`):

1. `files=$(find ...)` yields an empty string.
2. `total=$(echo "${files}" | wc -l | xargs)` — `echo ""` still prints one newline, so `wc -l` counts **1**: the log says `Found 1 XMIR file(s)`.
3. `while IFS= read -r f; do ... done <<< "${files}"` — a here-string from an empty variable still feeds one empty line, so `read` succeeds once with `f=""`.
4. The task list gets a line for `"${HONE_XMIR_IN}/.xmir"` (`f=${f%.*}` leaves `f` empty).
5. The worker runs `phino rewrite ... "${HONE_XMIR_IN}/.xmir"`, phino fails on the missing file, the failure branch (`rewrite.sh:205-209`) exits non-zero, and `parallel --halt-on-error` / `set -e` aborts the entire pipeline.

## How to reproduce

```sh
$ mkdir -p /tmp/empty/in
$ files=$(find "/tmp/empty/in" -name "*.xmir" -type f | LC_ALL=C sort)
$ total=$(echo "${files}" | wc -l | xargs)
$ echo "total=$total"     # prints total=1, even though there are no files
$ while IFS= read -r f; do echo "GOT: [$f]"; done <<< "${files}"
GOT: []
```

## Expected behavior

With zero XMIR files the script should log something like `No XMIR files to process` and exit 0, leaving the pipeline untouched. The `total` count and the loop must both treat an empty `files` as "nothing to do".

## Details

- `src/main/resources/org/eolang/hone/scaffolding/rewrite.sh:272` — empty `files`
- `src/main/resources/org/eolang/hone/scaffolding/rewrite.sh:273` — `total=1` on empty input
- `src/main/resources/org/eolang/hone/scaffolding/rewrite.sh:279-288` — phantom iteration with `f=""`
- `src/main/resources/org/eolang/hone/scaffolding/rewrite.sh:205-209` — failure branch that aborts the run

morphqdd:
`echo ""` still prints a newline, so `wc -l` counted one file where there were none, and a here-string built from an empty variable still feeds the loop one empty line. The task list then got an entry for `${HONE_XMIR_IN}/.xmir`, phino failed on the missing file, and `--halt-on-error` took the whole pipeline down.

The script now says so and stops before any of that, which is what the goal should do when `hone.includes` selects nothing or the project is empty.

Closes #836

### DIFF
--- a/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
@@ -291,6 +291,10 @@ if [ -n "${HONE_GREP_IN}" ]; then
 fi
 
 files=$(find "$("${RP}" "${HONE_XMIR_IN}")" -name '*.xmir' -type f -exec "${RP}" --relative-to="${HONE_XMIR_IN}" {} \; | LC_ALL=C sort)
+if [ -z "${files}" ]; then
+  echo "No XMIR files to process"
+  exit 0
+fi
 total=$(echo "${files}" | wc -l | xargs)
 tasks=${TARGET}/hone-tasks.txt
 mkdir -p "$(dirname "${tasks}")"


0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+7035](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #875, opened 2026-09-06
fix(#867): make rewrite.sh portable to macOS (nproc, @Q, date %N)

### LINKED ISSUE #867: rewrite.sh is not macOS-portable: nproc, the @Q operator and date %N have no fallbacks
## Problem

`rewrite.sh` runs on the host when phino is installed locally (`OptimizeMojo.withoutDocker()`), not only inside the Linux Docker image. On macOS the script breaks in three independent places because it assumes GNU tools:

### 1. `nproc` does not exist on macOS — hard failure

`rewrite.sh:317`: `threads=$(nproc)` with no fallback.

- `nproc` is a GNU coreutils command; macOS has neither `nproc` nor `gnproc` in the default Homebrew coreutils set.
- With the default `HONE_THREADS` unset (or `0`), `threads` becomes empty. The script then hits `[ "${threads}" -eq 1 ]` (`rewrite.sh:322`) and dies with `bash: [: : integer expression expected` under `set -e` — the whole optimize aborts.
- Note that the same script *does* provide GNU fallbacks elsewhere: `realpath` → `grealpath` (`rewrite.sh:8-15`) and `setsid` → `gsetsid`/Homebrew paths (`rewrite.sh:22-34`). `nproc` has no such fallback.

### 2. `${0@Q}` requires bash ≥ 4.4 — macOS ships bash 3.2

`rewrite.sh:312`: `printf "%s rewrite_with_timeout ...\n" "${0@Q}" "${i@Q}" ...`.

- The `@Q` parameter transformation operator needs bash 4.4+.
- macOS `/bin/bash` is 3.2 (the script relies on `#!/usr/bin/env bash`, which may resolve to 3.2 if Homebrew bash is not installed or not on PATH).
- On bash 3.2 this is a **syntax error** when the tasks file is generated — the script dies at parse time of that line.

### 3. `date '+%s.%N'` — BSD `date` does not support `%N`

`rewrite.sh:110,135,173,210,215,321,341`.

- GNU `date`'s `%N` (nanoseconds) is not supported by BSD `date`; macOS prints the literal `%N`.
- Every timing arithmetic (`perl -E "say int(... - ...)"` at `rewrite.sh:135,210,215,341`) then gets a non-numeric value, prints a Perl warning, and produces garbage timings. `hone-timings.csv`, the per-file rate and the total elapsed time are all corrupted (not a crash, but the metrics are useless).

## Expected behavior

Make `rewrite.sh` portable to macOS:

- `threads=$(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || echo 1)` (or a `nproc`/`gnproc` fallback like the existing `realpath`/`grealpath` pattern at `rewrite.sh:8-15`).
- Replace `${0@Q}` with a portable quoting helper (a `printf %q` wrapper guarded by a bash-version check, or use `$'\''` escaping), or detect bash < 4.4 and fail with a clear message instead of a syntax error.
- Replace `date '+%s.%N'` with a portable high-resolution clock, e.g. `perl -MTime::HiRes=time -e 'printf "%.6f\n", time'`, or fall back to `gdate` when present (mirroring the `grealpath` pattern). `entry.sh` uses `date '+%s.%N'` at `entry.sh:208,211` too.

This matters for real users: `Phino` (`Phino.java:43-78`) runs phino directly when its `--version` matches, so macOS + locally-installed phino takes the `withoutDocker()` path (`OptimizeMojo.java:391-394`) and hits all three breakages.

## Details

- `rewrite.sh:317` — `threads=$(nproc)` (no fallback)
- `rewrite.sh:322` — `[ "${threads}" -eq 1 ]` — dies on empty
- `rewrite.sh:312` — the `@Q` operator on `0`, `i`, `phi`, `pho`, `xi`, `xo`
- `rewrite.sh:110,135,173,210,215,321,341` — `date '+%s.%N'`
- `entry.sh:208,211` — same `date '+%s.%N'` in `generate_statistics` / timing
- `rewrite.sh:8-15` — existing `realpath`/`grealpath` fallback (the pattern to follow)
- `rewrite.sh:22-34` — existing `setsid`/`gsetsid` fallback
- `OptimizeMojo.java:391-394` — `withoutDocker()` path that runs `rewrite.sh` on the host

VasilevNStas:
Fixes #867.

## Problem

`rewrite.sh` is run on the host whenever phino is installed locally (the `withoutDocker()` path), and on macOS it broke in three independent places:

1. `nproc` does not exist on macOS — `threads=$(nproc)` produced an empty value, then `[ "${threads}" -eq 1 ]` died with `integer expression expected` under `set -e`.
2. `${0@Q}` and friends require bash >= 4.4; macOS ships bash 3.2, where the operator is a syntax error.
3. `date '+%s.%N'` is GNU-only — BSD `date` prints a literal `%N`, corrupting all timing arithmetic.

## Solution

- **threads**: `threads=$(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || echo 1)` — falls back to the macOS CPU count, then to 1.
- **quoting**: new `q()` helper using `printf '%q'` (supported by bash 3.2), replacing every `@Q` use when writing the tasks file.
- **clock**: new `now()` helper using `perl -MTime::HiRes=time` (present on both macOS and Linux, core module), falling back to `date '+%s'`. All `date '+%s.%N'` occurrences in `rewrite.sh` and `entry.sh` now go through `now()`.

## Changes

| File | Change |
|------|--------|
| `src/main/resources/org/eolang/hone/scaffolding/rewrite.sh` | add `now()`/`q()`, use them; `nproc` fallback chain |
| `src/main/resources/org/eolang/hone/scaffolding/entry.sh` | add `now()`, replace `date '+%s.%N'` |

## Tests

- `bash -n` + `shellcheck` — clean on both scripts
- `now()` returns a float second value on macOS (verified); `q()` escapes paths with spaces
- Timing arithmetic verified: `int(end - start)` works with the float values

### DIFF
--- a/src/main/resources/org/eolang/hone/scaffolding/entry.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/entry.sh
@@ -41,6 +41,12 @@ fi
 
 SELF=$(dirname "$0")
 
+function now {
+  # Current time in seconds with sub-second precision. macOS's BSD date has
+  # no %N, so use perl's high-resolution clock when available (see #867).
+  perl -MTime::HiRes=time -e 'printf "%.6f\n", time' 2>/dev/null || date '+%s'
+}
+
 if [ "${LANG}" != 'en_US.UTF-8' ]; then
   echo "Setting locale to en_US.UTF-8 from '${LANG}'"
   LANG=en_US.UTF-8
@@ -205,10 +211,10 @@ function record_timing {
 }
 
 function elapsed {
-  awk -v start="${1}" -v end="$(date '+%s.%N')" 'BEGIN { printf "%.3f\n", end - start }'
+  awk -v start="${1}" -v end="$(now)" 'BEGIN { printf "%.3f\n", end - start }'
 }
 
-start=$(date '+%s.%N')
+start=$(now)
 (
   set -x
   mvn "${disassemble_opts[@]}" "org.eolang:jeo-maven-plugin:${JEO_VERSION}:disassemble"
@@ -232,7 +238,7 @@ if [ "${SKIP_PHINO}" != 'true' ]; then
   export HONE_THREADS="${THREADS}"
   export HONE_TIMEOUT="${TIMEOUT}"
   export HONE_STATISTICS
-  start=$(date '+%s.%N')
+  start=$(now)
   (
     set -x
     "${SELF}/rewrite.sh"
@@ -240,7 +246,7 @@ if [ "${SKIP_PHINO}" != 'true' ]; then
   record_timing "phino:rewrite (default-cli)" "$(elapsed "${start}")"
 fi
 
-start=$(date '+%s.%N')
+start=$(now)
 (
   set -x
   mvn "${assemble_opts[@]}" "org.eolang:jeo-maven-plugin:${JEO_VERSION}:assemble"
--- a/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
@@ -72,6 +72,17 @@ function atomic_write {
   fi
 }
 
+function now {
+  # Current time in seconds with sub-second precision. macOS's BSD date has
+  # no %N, so use perl's high-resolution clock when available (see #867).
+  perl -MTime::HiRes=time -e 'printf "%.6f\n", time' 2>/dev/null || date '+%s'
+}
+
+function q {
+  # Bash-compatible quoting of a single argument (portable printf %q).
+  printf '%q' "${1}"
+}
+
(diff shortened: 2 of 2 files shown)


VasilevNStas:
@yegor256 @volodya-lombrozo 
please review

yegor256:
Thanks for the contribution, merged!

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1890](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #879, opened 2026-09-06
fix(#863): let the watchdog finish its KILL escalation and expose the kill grace

### LINKED ISSUE #863: rewrite.sh KILL escalation is cut short, orphaning phino again (#727 incomplete); HONE_KILL_GRACE is never exported
## Problem

`rewrite.sh` runs every phino rewrite inside `rewrite_with_timeout` with a watchdog that is supposed to escalate TERM → grace → KILL so a timed-out phino never survives as an orphan (this was the fix for #727). The escalation, however, is **cut short**, so the leak #727 fixed is back on every timeout.

### How it is supposed to work

```sh
# rewrite.sh:193-203  (watchdog subshell)
(
  sleep "${HONE_TIMEOUT}"
  : > "${flag}"
  if [ -n "${leader}" ]; then
    kill_tree TERM "${leader}"
    kill -TERM "-${leader}" 2>/dev/null || true
    sleep "${HONE_KILL_GRACE:-10}"   # grace: phino may clean up
    kill_tree KILL "${leader}"       # after grace, force-kill the whole tree
    kill -KILL "-${leader}" 2>/dev/null || true
  fi
) &
watchdog=$!
```

The idea (#727): when the deadline passes, first TERM the whole session, give the tree a grace window, and then KILL anything that survived, so no phino is left burning CPU.

### Why it breaks

```sh
# rewrite.sh:205-207
wait "${sid}" || code=$?          # returns when the setsid child (the bash worker) exits
kill "${watchdog}" 2>/dev/null || true   # watchdog killed HERE
wait "${watchdog}" 2>/dev/null || true
```

`setsid --wait` waits for its direct child, which is **the bash re-invocation of this script** (`rewrite.sh:184`: `"${SETSID}" --wait "${0}" rewrite "$@"`), **not** the phino process that the `rewrite` function launches in the foreground. On timeout:

1. The deadline passes; the watchdog sends TERM to the leader and its tree (`kill_tree TERM`).
2. The bash worker — a direct child of setsid — exits **quickly** on TERM (bash dies before phino even notices the signal; phino is a grandchild).
3. `wait "${sid}"` returns **while phino is still alive**.
4. The main flow immediately kills the watchdog with `kill "${watchdog}"` (`rewrite.sh:206`).
5. The watchdog was still in its `sleep "${HONE_KILL_GRACE:-10}"` (`rewrite.sh:199`) — it never reaches `kill_tree KILL` (`rewrite.sh:200-201`).
6. phino survives as an orphan, keeps burning CPU, and the build never converges — exactly what #727 set out to prevent.

The KILL phase only runs if phino happens to die *slower* than bash within the grace window, which is a race, not a guarantee.

### `HONE_KILL_GRACE` is dead

Even if the escalation were fixed, the grace window cannot be tuned:

- `entry.sh:219-234` exports `HONE_TIMEOUT` (`entry.sh:233`) but **never exports `HONE_KILL_GRACE`**.
- `OptimizeMojo.withDocker()` (`OptimizeMojo.java:511`) and `withoutDocker()` (`OptimizeMojo.java:664`) put `TIMEOUT` into the environment, but not `KILL_GRACE`.
- Therefore `rewrite.sh:199` always takes the `:-10` default; the knob is non-functional.

## Expected behavior

- Do not kill the watchdog on `wait "${sid}"` returning. Let the watchdog run its full TERM → grace → KILL sequence and exit on its own, or restructure so the grace+KILL phase is a supervisor that the caller waits on until it has completed. phino must be dead before the task is considered finished.
- Export `HONE_KILL_GRACE` (backed by a new Java parameter, e.g. `hone.kill-grace`, default 10) through `entry.sh` → `rewrite.sh` so the grace window is configurable.

## Details

- `rewrite.sh:167-221` — `rewrite_with_timeout` (the whole function)
- `rewrite.sh:184` — `"${SETSID}" --wait "${0}" rewrite "$@"` — setsid waits on bash, not phino
- `rewrite.sh:193-203` — watchdog subshell: TERM, grace `sleep`, then `kill_tree KILL`
- `rewrite.sh:199` — `sleep "${HONE_KILL_GRACE:-10}"` — dead default, never exported
- `rewrite.sh:205` — `wait "${sid}"` returns when bash exits, possibly before phino dies
- `rewrite.sh:206-207` — watchdog killed here → KILL phase skipped → orphaned phino
- `rewrite.sh:157-165` — `kill_tree` (children before parents, so nothing is reparented to init)
- `entry.sh:219-234` — exports block; `HONE_KILL_GRACE` missing (only `HONE_TIMEOUT` at `entry.sh:233`)
- `OptimizeMojo.java:511,664` — env `TIMEOUT` passed; no `KILL_GRACE`

VasilevNStas:
Fixes #863.

## Problem

`rewrite_with_timeout` in `rewrite.sh` runs every phino rewrite under a watchdog that escalates TERM → grace → KILL so a timed-out phino never survives as an orphan (the #727 fix). The escalation was cut short:

- `wait "${sid}"` returns as soon as the **bash worker** (the direct setsid child) exits on TERM — which happens before phino (a grandchild) dies.
- The main flow then immediately killed the watchdog with `kill "${watchdog}"` (`rewrite.sh:222`), while the watchdog was still in its `sleep "${HONE_KILL_GRACE:-10}"`. The KILL phase never ran, and phino was orphaned again on every timeout.
- Separately, `HONE_KILL_GRACE` was never exported (`entry.sh` exports `HONE_TIMEOUT` but not `HONE_KILL_GRACE`; `OptimizeMojo` passes `TIMEOUT` but no `KILL_GRACE`), so the knob was permanently dead at its `:-10` default.

## Solution

- **`rewrite.sh`**: only stop the watchdog when the deadline did NOT fire (no flag file) — i.e. it is still in its initial sleep. When the timeout flag is present, the watchdog is mid-escalation, so the main flow waits for it to complete the TERM → grace → KILL sequence, guaranteeing phino is dead before the task returns.
- **`OptimizeMojo`**: new `hone.kill-grace` parameter (default `10`), forwarded as `KILL_GRACE` in both the Docker and no-Docker paths.
- **`entry.sh`**: export `HONE_KILL_GRACE` from `KILL_GRACE`, so `rewrite.sh:199` reads a real value.

## Changes

| File | Change |
|------|--------|
| `src/main/java/org/eolang/hone/OptimizeMojo.java` | add `hone.kill-grace`; pass `KILL_GRACE` env in both paths |
| `src/main/resources/org/eolang/hone/scaffolding/entry.sh` | export `HONE_KILL_GRACE` |
| `src/main/resources/org/eolang/hone/scaffolding/rewrite.sh` | don't kill the watchdog when the timeout fired; wait for the KILL escalation |

## Tests

- `bash -n` + `shellcheck` — clean on both scripts
- `mvn clean install -Pqulice` — 0 offenses
- Simulated escalation (grace 10s, process ignoring TERM): with the fix the watchdog survives to the KILL phase and phino dies; the old always-kill code orphans it (`FAIL: phino ALIVE`)

### DIFF
--- a/src/main/java/org/eolang/hone/OptimizeMojo.java
+++ b/src/main/java/org/eolang/hone/OptimizeMojo.java
@@ -261,6 +261,20 @@ public final class OptimizeMojo extends AbstractMojo {
     private int timeout;
 
     /**
+     * How many seconds to wait after the TERM signal before force-killing a
+     * timed-out phino process tree.
+     *
+     * <p>The watchdog of {@code rewrite.sh} sends TERM to the whole process
+     * group once the {@code hone.timeout} deadline passes, then gives the tree
+     * this grace window to clean up, and finally force-kills whatever survived,
+     * so no orphan phino is left burning CPU (see #863).</p>
+     *
+     * @since 0.38.0
+     */
+    @Parameter(property = "hone.kill-grace", defaultValue = "10")
+    private int killgrace;
+
+    /**
      * How many threads to use for rewriting?
      *
      * <p>By default, it is set to zero, which means the number of threads
@@ -508,7 +522,8 @@ public final class OptimizeMojo extends AbstractMojo {
         );
         command.addAll(
             Arrays.asList(
-                "--env", String.format("TIMEOUT=%d", this.timeout)
+                "--env", String.format("TIMEOUT=%d", this.timeout),
+                "--env", String.format("KILL_GRACE=%d", this.killgrace)
             )
         );
         if (this.includes != null && this.includes.length > 0) {
@@ -662,6 +677,7 @@ public final class OptimizeMojo extends AbstractMojo {
                 .withEnv("MAX_CYCLES", Integer.toString(this.maxCycles))
                 .withEnv("THREADS", Integer.toString(this.threads))
                 .withEnv("TIMEOUT", Integer.toString(this.timeout))
+                .withEnv("KILL_GRACE", Integer.toString(this.killgrace))
                 .withEnv("RULES", this.rulesAsString());
             if (this.extra != null && !this.extra.isEmpty()) {
                 this.copyExtras(temp.path().resolve("hone-extra"));
--- a/src/main/resources/org/eolang/hone/scaffolding/entry.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/entry.sh
@@ -231,6 +231,7 @@ if [ "${SKIP_PHINO}" != 'true' ]; then
   export HONE_MAX_CYCLES="${MAX_CYCLES}"
   export HONE_THREADS="${THREADS}"
   export HONE_TIMEOUT="${TIMEOUT}"
+  export HONE_KILL_GRACE="${KILL_GRACE}"
   export HONE_STATISTICS
   start=$(date '+%s.%N')
   (
--- a/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
@@ -219,7 +219,13 @@ function rewrite_with_timeout {
   ) &
   watchdog=$!
   wait "${sid}" || code=$?
-  kill "${watchdog}" 2>/dev/null || true
+  # If the deadline fired, the watchdog is mid-escalation (TERM → grace →
+  # KILL) and killing it here would orphan phino again (#727, #863) — so only
+  # stop it when it is still in its initial sleep, i.e. the worker finished
+  # before the deadline.
+  if [ ! -f "${flag}" ]; then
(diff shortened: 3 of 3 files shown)


VasilevNStas:
@yegor256 
please review

yegor256:
Thanks for the contribution, merged!

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (27) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1990](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #886, opened 2026-09-07
#881: keep one copy of the classes, not a copy of the copy

### LINKED ISSUE #881: `classes-before-hone` nests a copy of itself on every run without `clean`
## What happens

`entry.sh` saves the classes before the optimization starts:

https://github.com/objectionary/hone-maven-plugin/blob/master/src/main/resources/org/eolang/hone/scaffolding/entry.sh#L91

```bash
# In order to save them "as is", just in case:
cp -R "${TARGET}/${CLASSES}" "${TARGET}/classes-before-hone"
echo "The binaries before hone are saved in '${TARGET}/classes-before-hone' ($(find ...))"
```

`cp -R src dst` copies the source *into* `dst` when `dst` already exists. Nothing removes
the previous backup, so the second run without `mvn clean` nests one copy inside the
other:

```bash
$ cp -R target/classes target/classes-before-hone
$ find target/classes-before-hone
target/classes-before-hone
target/classes-before-hone/A.class

$ cp -R target/classes target/classes-before-hone     # second run
$ find target/classes-before-hone
target/classes-before-hone
target/classes-before-hone/A.class
target/classes-before-hone/classes
target/classes-before-hone/classes/A.class
```

Running `mvn install` twice without `clean` is ordinary, and by then the classes are the
ones hone already rewrote, so the nested copy is not the "before" state at all. The
directory also grows on every run, and the file count that the line below prints grows
with it.

## What should happen

The backup should be a copy of the current classes and nothing else. Removing the previous
one first, or copying the contents rather than the directory (`cp -R "${TARGET}/${CLASSES}/." "${TARGET}/classes-before-hone"`),
would both do it.

morphqdd:
`cp -R src dst` puts the source inside `dst` when `dst` is already a directory, and nothing removed
the previous backup, so a second `mvn install` without `clean` left
`target/classes-before-hone/classes/...` next to the real one. By then the classes it copied were
the ones hone had already rewritten, so the nested copy was not the "before" state at all, and the
file count the line below prints grew with it.

The previous copy is removed first, so the backup holds the classes and nothing else.

`optimizesTwice` already runs `optimize` twice against the same `target`, which is the case that
produced this, so the check goes there.

Closes #881

### DIFF
--- a/src/main/resources/org/eolang/hone/scaffolding/entry.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/entry.sh
@@ -93,7 +93,10 @@ after the 'compile' phase is finished; this is what is in the '${TARGET}' direct
   fi
 fi
 
-# In order to save them "as is", just in case:
+# In order to save them "as is", just in case. The previous copy goes first,
+# because "cp -R" puts the source inside a directory that already exists, and
+# a second run without "clean" would otherwise nest one backup in the other:
+rm -rf "${TARGET}/classes-before-hone"
 cp -R "${TARGET}/${CLASSES}" "${TARGET}/classes-before-hone"
 echo "The binaries before hone are saved in '${TARGET}/classes-before-hone' ($(find "${TARGET}/classes-before-hone" -print | wc -l | xargs) files)"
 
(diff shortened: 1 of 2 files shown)


yegor256:
Merged. Thanks — removing the previous backup before copying is right, and `optimizesTwice` is the correct place for the check.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6748](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #887, opened 2026-09-07
#884: stop passing jeo options that do not exist

### LINKED ISSUE #884: `entry.sh` passes `-Djeo.assemble.includes` and `excludes`, which jeo does not define
## What happens

`entry.sh` passes the include and exclude globs to both jeo goals:

https://github.com/objectionary/hone-maven-plugin/blob/master/src/main/resources/org/eolang/hone/scaffolding/entry.sh

```bash
if [ -n "${INCLUDES}" ]; then
    disassemble_opts+=("-Djeo.disassemble.includes=${INCLUDES}")
    assemble_opts+=("-Djeo.assemble.includes=${INCLUDES}")
fi
if [ -n "${EXCLUDES}" ]; then
    disassemble_opts+=("-Djeo.disassemble.excludes=${EXCLUDES}")
    assemble_opts+=("-Djeo.assemble.excludes=${EXCLUDES}")
fi
```

`jeo.assemble.includes` and `jeo.assemble.excludes` do not exist. `AssembleMojo` declares
six properties and none of them is a filter:

```
jeo.assemble.sourcesDir
jeo.assemble.outputDir
jeo.assemble.skip.verification
jeo.assemble.xmir.verification
jeo.assemble.disabled
jeo.assemble.debug
jeo.assemble.threads
```

Maven ignores an unknown `-D` silently, so the two lines do nothing. The assemble goal
takes every `.xmir` it finds under its sources directory.

It happens to be harmless while `target/hone/unphi` holds exactly what this run
disassembled, and it stops being harmless as soon as that directory holds anything from an
earlier run with different filters: those files are assembled back into `target/classes`
even though the user excluded them.

## What should happen

Either the two lines should go, since the filtering already happened at disassembly, or
`jeo-maven-plugin` should be asked for the parameters and they should be used once they
exist.

morphqdd:
`jeo.assemble.includes` and `jeo.assemble.excludes` are not parameters of anything. `AssembleMojo`
declares six, and none of them filters:

```
jeo.assemble.sourcesDir
jeo.assemble.outputDir
jeo.assemble.skip.verification
jeo.assemble.xmir.verification
jeo.assemble.disabled
jeo.assemble.debug
```

Maven ignores an unknown `-D` without a word, so the two lines did nothing at all, while reading as
though the assembly were filtered the way the disassembly is.

The filtering happens once, at disassembly, and the assembly takes back whatever that left in the
sources directory. `OptimizeMojoRulesTest` already covers it: with `includes` and `excludes` set,
only the included class comes out optimized.

Closes #884

### DIFF
--- a/src/main/resources/org/eolang/hone/scaffolding/entry.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/entry.sh
@@ -185,13 +185,13 @@ declare -a assemble_opts=(
   "-Djeo.assemble.xmir.verification=false"
   "-Djeo.assemble.skip.verification=true"
 )
+# Only the disassembly is filtered. The assemble goal has no such parameters,
+# and it does not need them: it takes back what the disassembly left behind.
 if [ -n "${INCLUDES}" ]; then
   disassemble_opts+=("-Djeo.disassemble.includes=${INCLUDES}")
-  assemble_opts+=("-Djeo.assemble.includes=${INCLUDES}")
 fi
 if [ -n "${EXCLUDES}" ]; then
   disassemble_opts+=("-Djeo.disassemble.excludes=${EXCLUDES}")
-  assemble_opts+=("-Djeo.assemble.excludes=${EXCLUDES}")
 fi
 if [ "${SKIP_PHINO}" == 'true' ]; then
   echo "Skipping the phino step as requested"


yegor256:
Merged. Thanks — dropping two `-D` flags that no mojo declares.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6760](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #888, opened 2026-09-07
#885: skip only when there is nothing to run with

### LINKED ISSUE #885: `hone.skipWithoutDocker` skips the goal even when the local `phino` could run it
## What happens

`hone.skipWithoutDocker` skips the goal when Docker is missing:

https://github.com/objectionary/hone-maven-plugin/blob/master/src/main/java/org/eolang/hone/AbstractMojo.java

```java
} else if (this.skipWithoutDocker && !new Docker(this.sudo).available()) {
    Logger.info(this, "Execution skipped due to hone.skipWithoutDocker=true");
```

but Docker is no longer the only way to run. `OptimizeMojo` picks the local `phino` when
it is installed and of the expected version:

```java
if (this.alwaysWithDocker || !new Phino().available(this.phino())) {
    this.withDocker();
} else {
    this.withoutDocker();
}
```

So on a machine with `phino` installed and no Docker, the optimization that would have run
perfectly well is skipped instead, and the build silently produces unoptimized bytecode.
The parameter is described as "Skip the execution, if Docker is not available", which is
what it does, and that description predates the no-Docker path.

## What should happen

The goal should be skipped only when neither way of running is available, that is when
Docker is missing and the local `phino` is missing or of the wrong version.

morphqdd:
`hone.skipWithoutDocker` skipped on the absence of Docker alone, and Docker stopped being the only
way to run. All four goals here take the local `phino` when it is installed and of the expected
version:

```java
BuildMojo:    if (!this.alwaysWithDocker && new Phino().available(this.phino()))
PullMojo:     if (this.alwaysWithDocker || !new Phino().available(this.phino()))
RmiMojo:      if (!this.alwaysWithDocker && new Phino().available(this.phino()))
OptimizeMojo: if (this.alwaysWithDocker || !new Phino().available(this.phino()))
```

So on a machine with `phino` and no Docker the optimization that would have run was skipped, and
the build quietly produced unoptimized bytecode.

The flag now skips only when neither way is there: no Docker, and no usable local `phino`. When
`alwaysWithDocker` is set, `phino` is not consulted, because it would not be used anyway.

About the test. What changed only shows on a machine that has no Docker, and both the CI here and
my own machine have it, so any test I could add would take the same branch it takes today and
prove nothing. The `@DisabledWithoutPhino` tests already cover the path that does the running.
Making this deterministic would mean handing `Docker` and `Phino` to the mojo instead of building
them inside it, which is a larger change than this bug asks for. I would rather say so than write a
test that passes either way.

Closes #885

### DIFF
--- a/src/main/java/org/eolang/hone/AbstractMojo.java
+++ b/src/main/java/org/eolang/hone/AbstractMojo.java
@@ -98,7 +98,7 @@ abstract class AbstractMojo extends org.apache.maven.plugin.AbstractMojo {
     private boolean skip;
 
     /**
-     * Skip the execution, if Docker is not available.
+     * Skip the execution, if there is no way to run: no Docker and no local phino.
      * @since 0.22.0
      * @checkstyle MemberNameCheck (6 lines)
      */
@@ -118,7 +118,7 @@ abstract class AbstractMojo extends org.apache.maven.plugin.AbstractMojo {
         StaticLoggerBinder.getSingleton().setMavenLog(this.getLog());
         if (this.skip) {
             Logger.info(this, "Execution skipped due to hone.skip=true");
-        } else if (this.skipWithoutDocker && !new Docker(this.sudo).available()) {
+        } else if (this.skipWithoutDocker && this.helpless()) {
             Logger.info(this, "Execution skipped due to hone.skipWithoutDocker=true");
         } else if (this.skipOnWindows && AbstractMojo.windows()) {
             Logger.info(this, "Execution skipped due to hone.skipOnWindows=true");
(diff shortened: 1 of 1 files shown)


yegor256:
CI is green and the reasoning about the four mojos is right, but this changes when a build silently produces unoptimized bytecode and — as you say yourself — nothing in the suite can exercise the new branch, since both CI and your machine have Docker. I would rather not merge a behaviour change to `skipWithoutDocker` that no test covers.

Assigning to @yegor256.

---

yegor256:
Merged, thanks! The reasoning in the description about why a deterministic test would need `Docker` and `Phino` injected is fair, and stating it beats a test that passes either way.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (26) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6784](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #889, opened 2026-09-07
#882: find every rule and hand them all over

### LINKED ISSUE #882: `RULES` fallback in `entry.sh` finds no `.phr` rules and `rewrite.sh` keeps only the first one
## What happens

When `RULES` is not set, `entry.sh` builds the list itself:

https://github.com/objectionary/hone-maven-plugin/blob/master/src/main/resources/org/eolang/hone/scaffolding/entry.sh#L131

```bash
if [ -z "${RULES}" ]; then
fi
```

Two things go wrong with that list.

First, it looks for `*.yml` only, while the rules directory holds 107 `.phr` files and 2
`.yml` files:

```
$ find src/main/resources/org/eolang/hone/rules -type f | wc -l
109
$ find src/main/resources/org/eolang/hone/rules -name '*.phr' | wc -l
107
```

`Rules.discover()` in the Java side accepts both extensions, so the shell fallback and the
Java side disagree about what a rule is.

Second, `find` prints one path per line, and `rewrite.sh` splits the list with `read`,
which stops at the first newline:

https://github.com/objectionary/hone-maven-plugin/blob/master/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh#L89

```bash
IFS=' ' read -r -a rules <<< "${HONE_RULES}"
```

```bash
$ export HONE_RULES=$(printf '/x/one.yml\n/x/two.yml\n/x/three.yml')
$ IFS=' ' read -r -a rules <<< "${HONE_RULES}"; echo "count=${#rules[@]}"; printf '%s\n' "${rules[@]}"
count=1
/x/one.yml
```

So `phino rewrite` is given a single `--rule=` and the rest are dropped without a word.

The same newline separation makes the count in the log wrong, since it counts spaces:

```bash
printf 'Using the following %d rules:\n\t%b\n' \
  "$(( "$(echo "${RULES}" | grep -o ' ' | wc -l)" + 1))" \
  "${RULES// /\\n\\t}"
```

which prints "Using the following 1 rules" and then lists all of them.

`OptimizeMojo` always exports a space-separated `RULES`, in both the Docker and the
no-Docker path, so this only bites someone who runs the image by hand without `-e RULES`.
It is silent when it does.

## What should happen

The fallback should find `.phr` files as well, and the list should survive the split, for
example by keeping it newline-separated and reading it with `mapfile`/`readarray` instead
of `read`, or by joining it with spaces where it is built. The count in the log should
match the number of rules actually used.

morphqdd:
The fallback list was wrong twice over.

It looked for `*.yml` only, while the rules directory holds 107 `.phr` files and 2 `.yml` ones, and
`Rules.discover()` on the Java side accepts both:

```java
if (!path.endsWith(".phr") && !path.endsWith(".yml")) {
```

And `find` prints one path per line, while `rewrite.sh` splits the list with `read`, which stops at
the first newline, so `phino rewrite` was handed a single `--rule=` whatever the list held:

```
$ RULES=$(find rules \( -name '*.yml' -o -name '*.phr' \) -exec realpath {} \; | sort | tr '\n' ' ')
$ IFS=' ' read -r -a rules <<< "${RULES}"; echo "count=${#rules[@]}"
count=4          # one before this change
```

Both extensions are found now, and the list is separated by spaces the way the `EXTRA` branch a few
lines below already does it with `| sort | tr '\n' ' '`. The trailing space goes, so the count the
script prints stays right.

Closes #882

### DIFF
--- a/src/main/resources/org/eolang/hone/scaffolding/entry.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/entry.sh
@@ -135,7 +135,11 @@ fi
 echo "Using JEO version ${JEO_VERSION}"
 
 if [ -z "${RULES}" ]; then
-  RULES=$(find "${SELF}/rules" -name '*.yml' -exec "${RP}" {} \;)
+  # Both extensions, the way Rules.discover() reads them on the Java side, and
+  # separated by spaces, because rewrite.sh splits this list with "read -a",
+  # which would stop at the first newline:
+  RULES=$(find "${SELF}/rules" \( -name '*.yml' -o -name '*.phr' \) -exec "${RP}" {} \; | sort | tr '\n' ' ')
+  RULES="${RULES% }"
 fi
 for rule in ${RULES}; do
   if [ ! -e "${rule}" ]; then


yegor256:
Merged. Thanks — the `count=4` reproduction made the `read`-stops-at-newline half obvious.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (6) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6772](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #890, opened 2026-09-09
Upgrade qulice to 0.35.2 and fix all warnings

yegor256:
Bumps `qulice-maven-plugin` from `0.30.13` to `0.35.2`.

The new version brings several checks this codebase did not satisfy — 194 violations in total. All of them are fixed here, and no suppression or exclusion was added to the `qulice` profile.

| Check | Count | What was done |
| --- | --- | --- |
| `JavadocEmptyLineBeforeTagCheck` | 140 | Inserted the missing empty Javadoc line between a description and the block of at-clauses below it. |
| `NoJavadocForPrivateMethodsCheck` | 47 | Removed Javadoc from private methods. |
| `ImplicitConstructorCheck` | 3 | Declared the constructor explicitly in the three JUnit extension classes. |
| `SingleUseConstantCheck` | 2 | Inlined `STAGES` and `RACES` in `RandomPipeline`. |
| `ProhibitStaticNestedClassesCheck` | 1 | Moved `Summary.Collector` into its own file. |
| `FullyQualifiedTypeCheck` | 1 | Imported `Locale` in `AbstractMojo`. |

A few of these deserve more than a line.

**Javadoc on private methods.** Most of the 47 blocks only restated the signature in prose (`@param dir The directory`, `@return TRUE if ...`), and those are simply gone — which is what the check exists for. Nine of them carried an explanation that the method's name and body do not give away: why `scrambled()` mixes the seed by SplitMix64's finalizer rather than leaving it to `Random`, why `raceable()` refuses `parallel()` to a counted traversal or a floating-point reduction, why `grepInMatches()` validates the pattern through `grep -E` rather than `java.util.regex.Pattern` (#671). Those explanations stay, as a plain block comment above the method with the at-clauses dropped, so nothing that was worth reading was deleted.

**Single-use constants.** `STAGES` and `RACES` are inlined into their only usage in `RandomPipeline.walk()`. What their Javadoc documented — seven is the largest number of intermediate operations one pipeline gets, one walk in five goes parallel — moves into the comment above `walk()`, because `MethodBodyCommentsCheck` prohibits a comment inside a method body and that is where the two literals now sit.

**`Summary.Collector`.** It becomes `Collector`, package-private, in `src/main/java/org/eolang/hone/Collector.java`. The private `Summary.same()` helper went with it, since `Collector.visitFile()` was its only caller, and `Summary` shed the five imports that only the nested class needed.

## Verification

Locally:

- `mvn clean install -DskipTests -Dinvoker.skip -Pqulice` — `BUILD SUCCESS`, zero violations (was 194 before the fixes, on the same 0.35.2).
- `mvn test` — 72 tests, 0 failures.
- `mvn -Pdeep test` — 92 tests, 0 failures, 21 skipped. The skips are the end-to-end optimize fixtures and the single-rule packs, which need Docker and a local `phino`; neither was available where this ran.
- `mvn javadoc:javadoc` — no warnings, which is what `ImplicitConstructorCheck` is guarding against in the first place.

On this PR, all 19 checks pass, `qulice` and both `mvn` matrix legs among them.

Worth flagging about the 21 local skips: `-Pdeep` is used only by `deep.yml` and `codecov.yml`, and both of those trigger on push to `master` alone, so the `@Tag("deep")` fixtures do not run on this pull request either — they will run once it is merged. What does exercise the pipeline end-to-end here is `smoke` (the plugin against `apache/commons-cli` with a real `phino`) and `make` (the `streams` rule tests), and both are green. Since no `.phr` rule, shell script, or pinned tool version was touched, the risk to the deep fixtures is a Java-side one, and the only behavioural change on that side is `Summary.Collector` moving out into `Collector`, which `SummaryTest` covers.

https://claude.ai/code/session_01UxobLW5wZwyeMCWc71FGSm

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -404,7 +404,7 @@
           <plugin>
             <groupId>com.qulice</groupId>
             <artifactId>qulice-maven-plugin</artifactId>
-            <version>0.30.13</version>
+            <version>0.35.2</version>
             <configuration>
               <excludes>
                 <!--
--- a/src/main/java/org/eolang/hone/AbstractMojo.java
+++ b/src/main/java/org/eolang/hone/AbstractMojo.java
@@ -7,6 +7,7 @@ package org.eolang.hone;
 import com.jcabi.log.Logger;
 import java.io.File;
 import java.io.IOException;
+import java.util.Locale;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.cactoos.io.ResourceOf;
@@ -30,6 +31,7 @@ abstract class AbstractMojo extends org.apache.maven.plugin.AbstractMojo {
 
     /**
      * The "target/" directory of Maven project.
+     *
      * @since 0.1.0
      */
     @Parameter(
@@ -40,6 +42,7 @@ abstract class AbstractMojo extends org.apache.maven.plugin.AbstractMojo {
 
     /**
      * The base directory of the Maven project.
+     *
      * @since 0.1.0
      */
     @Parameter(
@@ -51,12 +54,14 @@ abstract class AbstractMojo extends org.apache.maven.plugin.AbstractMojo {
 
     /**
      * Timings tracker for performance measurements.
+     *
      * @since 0.1.0
      */
     protected Timings timings;
 
     /**
      * Docker image to use.
+     *
      * @since 0.1.0
      */
     @Parameter(property = "hone.image", defaultValue = "yegor256/hone:latest")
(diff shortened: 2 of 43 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (652 >= 100); -8 for way too many hits-of-code (652 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2347](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #908, opened 2026-09-09
Upgrade phino to 0.0.119

### LINKED ISSUE #907: Pinned phino version in `default-phino-version.txt` lags nine releases behind at 0.0.110
The default phino version pinned in `src/main/resources/org/eolang/hone/default-phino-version.txt:1` is still `0.0.110`, while phino has since released up to `0.0.119`. That single line is the only place the version lives: `Phino` compares it against `phino --version` to decide whether the host binary can be used directly, and the `Dockerfile` feeds it to `cabal install phino-${PHINO_VERSION}` as `ARG PHINO_VERSION`, so the pin decides both the host fast path and what lands in the image.

Lagging is wrong here because the gap silently splits contributors into two populations. A developer who installed phino recently gets `0.0.119` on the host, `Phino` sees the mismatch, and every rewrite is pushed through Docker instead of running natively — much slower, and the CLAUDE.md guidance to install phino on the host to skip Docker quietly stops working. The nine skipped releases also carry rule-visible changes (`0.0.117` treats a bare meta-variable differently, `0.0.116` tightened CLI validation and reseeded the synthetic-name generator, `0.0.119` changed how XMIR revisions are read and written), so the longer the pin sits still, the larger the untested jump becomes when it finally moves.

Bump that one line from `0.0.110` to `0.0.119`. No rule or script change is needed: every flag `rewrite.sh` passes (`--input`, `--output`, `--sweet`, `--rule`, `--max-cycles`, `--max-depth`, `--omit-listing`, `--omit-comments`, `--log-level`) still exists in `0.0.119`, so the bump is confined to the pin file.

yegor256:
Bumps `default-phino-version.txt` from `0.0.110` to `0.0.119`, plus a CI workaround the bump turned up.

The pin had drifted nine releases behind. Since it is what `Phino` compares against `phino --version`, anyone who installed phino recently stopped matching it, so their rewrites quietly fell back to Docker instead of running natively on the host. The same line feeds `cabal install phino-${PHINO_VERSION}` in the Dockerfile, so it also decides what ends up in the image.

The rules themselves survive the jump untouched — a full `mvn -Pdeep test` is green at 228 tests, including all 138 single-rule packs and the random-pipeline walk. Every flag `rewrite.sh` passes still exists in 0.0.119.

What did break is CI, and not because of anything here: phino has published no `ubuntu-24.04` binary since 0.0.116, because that one leg of its release workflow keeps failing on GHC setup. `install-phino.sh` downloads exactly that URL, so `make`, `mvn` and `smoke` all died on a bare `curl: (22) ... 403`. The `ubuntu-22.04` build of 0.0.119 does exist and runs fine on 24.04 runners, so this pulls that instead and pins its checksum. I checked it inside `ubuntu:24.04`: download, sha512 and `phino --pin=0.0.119 --version` all pass. That platform switch is a workaround and should be reverted once objectionary/phino#1141 is fixed — the comment in the script says so.

One more thing for anyone running the suite locally: the first Docker test after a bump rebuilds phino from source, and on an emulated `linux/amd64` host that overruns the 600s timeout on `generatesStatisticsWithDocker`. Warm the image once and it passes in 53s. Native CI does not pay that tax. Worth a separate issue, along with `optimizesJustOneLargeJnaClass` having no `@Timeout` at all while its siblings have `@Timeout(180L)`.

Closes #907


yegor256:
@yegor256 mind taking a look when you get a chance? Happy to explain anything about the bump if it is unclear.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2419](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #911, opened 2026-09-10
Upgrade phino to 0.0.133

yegor256:
Moves the pin in `default-phino-version.txt` from 0.0.119 to 0.0.133, the latest release, updates `.github/phino-sha512.txt` to that binary's checksum, and renames the data payload attribute in the four byte-level rules that 0.0.122 broke.

The checksum is of the **ubuntu-22.04** build, which is the one `install-phino.sh` downloads — phino's ubuntu-24.04 release leg is still not publishing (0.0.133 returns 403 on that path), so the workaround from #908 stays as it is.

## The `data` → `φ` rename

0.0.122 renamed the attribute carrying a data object's payload from `data` to `φ` ([phino#1142](https://github.com/objectionary/phino/issues/1142)). Phino still *reads* the old spelling, so a hand-written `Φ.bytes ( data ↦ … )` parses fine — which is what made this look harmless at first. But nothing phino *writes* uses `data` any more, and the pipeline feeds these rules phino's own output, where a literal desugars to `Φ.bytes ( φ ↦ ⟦ Δ ⤍ … ⟧ )`. The old pattern does not match that, so every byte-level rule silently stopped firing.

The gradle fixture caught it on the first push: `No changes in 1/1 App.phi`, `Optimized 0/1 files`, and `AppTest` still reading `int is 33`. Renaming the attribute in the four places that spell it out is the whole fix:

* `rules/33-to-42.yml`
* `src/it/with-extra/src/main/rules/sevens-to-fives.yml`
* the four inline rules in `OptimizeMojoRulesTest`

No rule under `rules/streams/` mentions `bytes` or `data` at all, so the stream pipeline is untouched.

## Everything else

Every flag `rewrite.sh` passes — `--input`, `--output`, `--sweet`, `--max-cycles`, `--max-depth`, `--rule`, `--omit-listing`, `--omit-comments`, `--log-level` — still exists in 0.0.133. `rewrite` also gained `--seed`, which seeds the generator that mints synthetic names; it defaults to `0`, so behaviour is unchanged, but it is a possible future answer to the one-name-per-rule constraint that #777 forced on us.

## Verification

* `mvn -Pdeep test` against the real 0.0.133 binary: **182 tests, 0 failures, 0 errors, 17 skipped**, covering all 84 single-rule packs and the two without-Docker runs that roundtrip through jeo 0.15.3's `disassemble` and `assemble`.
* `mvn clean install -Pqulice`: clean.
* The gradle fixture's own `App` put through the optimize goal on the host: `Optimized 1/1 files`, `AppTest` green on `int is 42`.

The 17 skipped are the `@DisabledWithoutDocker` tests — this container has no Docker daemon — so the Docker-run opcode-count fixtures, `OptimizeMojoRulesTest` and `preservesWhatRandomPipelinesPrint` are left to CI.

One aside for anyone reproducing locally: `phino match` aborts with `hPutChar: invalid argument` under a `C`/`POSIX` locale, which makes all 84 packs fail for a reason that has nothing to do with the version (0.0.119 does the same). Run the suite under a UTF-8 locale.

https://claude.ai/code/session_01DQ9eRrxmR6omp1QGugyVAW


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3376](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #912, opened 2026-09-10
#891: unquote the glob that clears the small-step files

### LINKED ISSUE #891: Stale small-step files are never removed because the glob in `rm -f "${pho}.*"` is quoted
**What happens**

`rewrite.sh` clears the small-step intermediates with:

```bash
rm -f "${pho}.*"
```

The glob is inside double quotes, so the shell does not expand it. The command
looks for a file literally named `out.phi.*` and removes nothing:

```
$ touch out.phi out.phi.01 out.phi.02
$ pho=out.phi
$ rm -f "${pho}.*"
$ ls
out.phi  out.phi.01  out.phi.02
```

So the `.01`, `.02`, ... files written by a previous `smallSteps` run stay on
disk. A run with fewer rules than the one before leaves the tail of the old
sequence next to the new one, which is exactly the state the line was written
to prevent, and it makes the per-rule diffs misleading when debugging a rule.

**What should happen**

The glob should be outside the quotes, so that the intermediates of the
previous run are actually removed.

morphqdd:
`rm -f "${pho}.*"` keeps the glob inside the quotes, so the shell looks for a
file literally named `out.phi.*` and removes nothing. The `.01`, `.02`, ...
files of a previous `smallSteps` run stayed on disk next to the new ones.

The glob is now outside the quotes, so the path is still quoted and the
intermediates are actually removed.

Closes #891

### DIFF
--- a/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
@@ -132,7 +132,7 @@ function rewrite {
   fi
   atomic_write "${phi}" phino rewrite "${phinopts[@]}" --input=xmir --sweet "${xi}"
   verbose "Converted ${idx} XMIR ($(du -sh "${xi}" | cut -f1)) to $(basename "${phi}") ($(du -sh "${phi}" | cut -f1))"
-  rm -f "${pho}.*"
+  rm -f "${pho}".*
   pos=0
   start=$(now)
   if [ "${HONE_SMALL_STEPS}" == "true" ]; then


zaguzovmaksim0-hue (approved):
I reproduced the shell behavior with a path containing spaces. The old `rm -f "${pho}.*"` treats `*` literally and leaves the intermediates; `rm -f "${pho}".*` expands only the suffix glob while keeping the path prefix quoted, so `out.phi.01`, `out.phi.02`, etc. are removed and the base `out.phi` remains intact.

All current CI checks are green, including shellcheck and smoke. The one-line change matches the issue and preserves safe handling of whitespace in the prefix. Approving.

yegor256:
Merged, thanks!

---

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (2) hits-of-code; +4 to give you at least something.  Your running score is [+140](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5909](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #913, opened 2026-09-10
#897: quit before listing the target directory

### LINKED ISSUE #897: `entry.sh` runs `tree` before the `skipIfNoClasses` exit, so a host without `tree` fails the build
**What happens**

When the classes directory is missing, `entry.sh` prints a diagnostic tree
before it decides whether that is fatal:

```bash
if [ ! -e "${TARGET}/${CLASSES}" ]; then
  echo "There is no '${TARGET}/${CLASSES}' directory ..."
  tree "${TARGET}"
  if [ "${SKIP_IF_NO_CLASSES}" == 'true' ]; then
    echo "We don't fail but quit quietly, because of skipIfNoClasses=true"
    exit
```

The script runs with `set -e`. `tree` is installed in the Docker image, but
`OptimizeMojo.withoutDocker()` runs the very same script on the host when a
local `phino` of the right version is found. On a host without `tree` the
command fails, `set -e` ends the script with a non-zero code, and the build
fails at the point where `skipIfNoClasses=true` (the default) says it should
pass quietly.

**What should happen**

The quiet exit should not depend on a diagnostic command. Either the `tree`
call should tolerate being absent, or it should run after the decision to
fail, so that `skipIfNoClasses=true` means the same thing with and without
Docker.

morphqdd:
The script listed the target directory with `tree` before it looked at
`skipIfNoClasses`, so a host without `tree` failed under `set -e` on the path
that was meant to quit quietly. The non-Docker run uses the host's utilities,
where `tree` is not guaranteed.

The quiet exit now comes first, and the listing falls back to `ls -R` when
`tree` is not installed.

Closes #897

### DIFF
--- a/src/main/resources/org/eolang/hone/scaffolding/entry.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/entry.sh
@@ -82,15 +82,19 @@ fi
 if [ ! -e "${TARGET}/${CLASSES}" ]; then
   echo "There is no '${TARGET}/${CLASSES}' directory, which most probably means \
 that the project has not been compiled yet; make sure you use 'hone-maven-plugin' \
-after the 'compile' phase is finished; this is what is in the '${TARGET}' directory:"
-  tree "${TARGET}"
+after the 'compile' phase is finished"
   if [ "${SKIP_IF_NO_CLASSES}" == 'true' ]; then
     echo "We don't fail but quit quietly, because of skipIfNoClasses=true"
     exit
-  else
-    echo "We can't continue and must fail here. Set skipIfNoClasses to 'true' if you need a quiet pass."
-    exit 1
   fi
+  echo "This is what is in the '${TARGET}' directory:"
+  if command -v tree > /dev/null; then
+    tree "${TARGET}"
+  else
+    ls -R "${TARGET}"
+  fi
+  echo "We can't continue and must fail here. Set skipIfNoClasses to 'true' if you need a quiet pass."
+  exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


zaguzovmaksim0-hue (approved):
I reproduced both branches with no `tree` available in the test PATH. With `skipIfNoClasses=true`, the script exits 0 before any diagnostic listing; with it false, the script falls back to `ls -R` and then exits 1 as intended.

This preserves the fatal behavior for the non-skip path while making the documented/default quiet path independent of the optional `tree` utility. All current CI checks are green, including shellcheck and smoke. No correctness issue found. Approving.

yegor256:
Merged, thanks! The `ls -R` fallback is a nice touch beyond what #897 asked for.

---

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (12) hits-of-code; +4 to give you at least something.  Your running score is [+132](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5921](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #914, opened 2026-09-10
#899: join the readers of the output before reporting the code

### LINKED ISSUE #899: `Docker` never joins the threads that read the process output, so log lines arrive late or are lost
**What happens**

`Docker.fire()` starts two threads to copy the process output into the log and
never waits for them:

```java
stdout.start();
stderr.start();
final boolean done;
try {
    done = proc.waitFor(Docker.TIMEOUT, TimeUnit.SECONDS);
...
Logger.info(this, "+ %s -> 0x%04x in %[ms]s", ...);
```

`waitFor` returns when the process exits, not when its pipes are drained, so
whatever is still in the buffers is printed after the `-> 0x0000 in ...` line
that is supposed to close the command, interleaved with whatever the build
does next. On the timeout path it is worse: `destroyForcibly()` is called and
the `IOException` is thrown while both threads are still reading, so the tail
of the output that would say why the command hung is dropped.

For a goal whose whole diagnostic value is the container's output, this is the
part of the log people read.

**What should happen**

Both threads should be joined before the exit code is reported, so that the
command's output is complete and appears before the line that closes it.

morphqdd:
`fire` started two threads to copy the process output into the log and never
waited for them. `waitFor` returns when the process exits, not when its pipes
are drained, so the tail of the output landed after the line that closes the
command, and on the timeout path it was dropped altogether.

Both readers are now joined before the exit code is reported and before the
timeout failure is thrown.

Closes #899

### DIFF
--- a/src/main/java/org/eolang/hone/Docker.java
+++ b/src/main/java/org/eolang/hone/Docker.java
@@ -34,6 +34,13 @@ final class Docker {
     private static final long TIMEOUT = 3600L;
 
     /**
+     * How long to wait for the readers of the output to reach its end, in
+     * seconds. They stop on their own when the pipes close, so this is only
+     * a guard against a child of the command that keeps them open.
+     */
+    private static final long DRAIN = 30L;
+
+    /**
      * Whether to prepend "sudo" to Docker commands.
      */
     private final boolean sudo;
@@ -98,6 +105,17 @@ final class Docker {
         return this.fire(command);
     }
 
+    private void drained(final Thread... pumps) throws IOException {
+        try {
+            for (final Thread pump : pumps) {
+                pump.join(TimeUnit.SECONDS.toMillis(Docker.DRAIN));
+            }
+        } catch (final InterruptedException ex) {
+            Thread.currentThread().interrupt();
+            throw new IOException("Docker output was interrupted while being read", ex);
+        }
+    }
+
     private int fire(final List<String> command) throws IOException {
         final long start = System.currentTimeMillis();
         Logger.info(this, "+ %s ...", String.join(" ", command));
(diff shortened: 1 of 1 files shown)


zaguzovmaksim0-hue (changes requested):
There is still one exit path that leaves both reader threads unjoined: the `InterruptedException` branch around `proc.waitFor(...)` (current `Docker.fire()` lines 118-126). It calls `destroyForcibly()`, restores the interrupt flag, and throws immediately, so `stdout` and `stderr` can outlive `fire()` exactly as they do in the timeout bug this PR is fixing.

Since this patch introduces a dedicated `drained(stdout, stderr)` helper specifically to make reader lifetime match the command lifetime, the interruption branch should use the same cleanup before returning/throwing. Otherwise the invariant is only fixed for normal completion and timeout, not for all exits after the pumps start.

Please drain/terminate the pumps on the interrupted-wait path too.

yegor256:
Merged, thanks!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5933](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (13) hits-of-code; +4 to give you at least something.  Your running score is [+355](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #915, opened 2026-09-10
#898: ask the daemon whether Docker is available

### LINKED ISSUE #898: `Docker.available()` runs `docker --version`, which succeeds while the daemon is down
**What happens**

`Docker.available()` decides that Docker is usable by asking the client for
its version:

```java
boolean available() {
    boolean yes = true;
    try {
        this.exec("--version");
    } catch (final IOException | IllegalStateException ex) {
        ...
```

`docker --version` prints the version of the CLI binary and exits with zero
without ever contacting the daemon. So on a machine where Docker is installed
but the daemon is not running, `available()` returns `true`.

`AbstractMojo.helpless()` then concludes that there is something to run with:

```java
boolean nothing = !new Docker(this.sudo).available();
if (nothing && !this.alwaysWithDocker) {
    nothing = !new Phino().available(this.phino());
}
```

so `hone.skipWithoutDocker=true` does not skip, the local `phino` is never
considered as a fallback, and the build fails later inside `docker run` with
whatever the daemon socket error happens to be.

**What should happen**

The probe should touch the daemon, for example `docker info`, so that
`available()` is false when Docker cannot actually run a container.

morphqdd:
`available()` ran `docker --version`, which prints the version of the client
binary and exits with zero without contacting the daemon. On a machine with
Docker installed but the daemon down, `helpless()` concluded there was
something to run with, `skipWithoutDocker` did not skip, the local `phino` was
never considered, and the build failed later inside `docker run`.

It now runs `docker info`, which fails when the daemon is not there.

Draft: it touches `Docker.java`, which #914 also changes.

Closes #898

### DIFF
--- a/src/main/java/org/eolang/hone/Docker.java
+++ b/src/main/java/org/eolang/hone/Docker.java
@@ -66,14 +66,19 @@ final class Docker {
     }
 
     /**
-     * Docker executable is available?
+     * Docker is available?
+     *
+     * <p>The daemon is asked, not the client: {@code docker --version} prints
+     * the version of the binary and exits with zero without ever reaching the
+     * socket, so a machine with a stopped daemon answered that Docker was
+     * usable and the build failed later inside {@code docker run}.</p>
      *
      * @return TRUE if Docker is here
      */
     boolean available() {
         boolean yes = true;
         try {
-            this.exec("--version");
+            this.exec("info", "--format", "{{.ServerVersion}}");
         } catch (final IOException | IllegalStateException ex) {
             Logger.warn(this, "Docker is not available: %s", ex.getMessage());
             yes = false;


yegor256:
Merged, thanks! Asking the daemon instead of the client is exactly the distinction #898 is about.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (9) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5945](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #901, opened 2026-09-09
`Phino.available()` leaves its reading thread alive after the probe times out


morphqdd:
**What happens**

In `Phino.available()` the reading thread is joined only when the process
finished in time:

```java
final boolean probed = proc.waitFor(3L, TimeUnit.SECONDS);
if (probed) {
    pump.join();
    available = Phino.version(proc, stdout, expected, this);
} else {
    proc.destroyForcibly();
    Logger.info(this, "The 'phino --version' probe timed out, we must use Docker");
}
```

On the timeout branch the thread is left blocked in `InputStream.read` on a
stream whose process has just been destroyed. It is not interrupted, not
joined, and nobody holds a reference to it.

`available()` is not called once per build. `AbstractMojo.helpless()` calls it,
and then `OptimizeMojo.optimize()`, `PullMojo`, `BuildMojo` and `RmiMojo` each
call it again, so a multi-module build with a hanging phino accumulates one
stuck thread per goal per module for the life of the Maven JVM.

**What should happen**

The thread should be interrupted and joined on the timeout branch too, so the
probe leaves nothing behind.

0crat:
@morphqdd Thanks for reporting a new bug! You've earned +12 points. By reporting bugs, you help our project improve its quality. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it. Your running score is [+6624](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #916, opened 2026-09-10
#900: read the error stream of the probe too

### LINKED ISSUE #900: `Phino.available()` does not read stderr, so a phino that writes to it makes the probe time out
**What happens**

`Phino.available()` probes the local executable and reads only its stdout:

```java
final Process proc = new ProcessBuilder(
    this.executable, "--version"
).start();
final ByteArrayOutputStream stdout = new ByteArrayOutputStream();
final Thread pump = new Thread(
    () -> Phino.pump(stdout, proc.getInputStream())
);
pump.start();
final boolean probed = proc.waitFor(3L, TimeUnit.SECONDS);
```

Nothing reads `proc.getErrorStream()`. A process that writes more to stderr
than the pipe buffer holds blocks on the write and never exits, so the
three-second wait expires, `destroyForcibly()` kills it, and the plugin logs
"the probe timed out, we must use Docker".

`phino --version` is quiet today, so this shows up as soon as it starts
warning about anything, for instance a locale or a config file, and the
symptom is that hone silently stops using the local phino and pulls the
Docker image instead. Nothing in the message says the executable was there and
working.

**What should happen**

The error stream should be drained as well, for example with
`redirectErrorStream(true)`, so that the probe measures whether phino answers,
not how much it says.

morphqdd:
The probe read only stdout of `phino --version`. A process that writes more to
stderr than the pipe holds blocks on that write and never exits, so the
three-second wait expired, the process was killed, and the plugin reported a
timeout and fell back to Docker even though the executable was there.

Both streams are read now, so the probe measures whether phino answers and not
how much it says.

Closes #900

### DIFF
--- a/src/main/java/org/eolang/hone/Phino.java
+++ b/src/main/java/org/eolang/hone/Phino.java
@@ -42,6 +42,12 @@ final class Phino {
     /**
      * Is it available?
      *
+     * <p>Both streams of the probe are read. A process that writes more to
+     * stderr than the pipe holds blocks on the write and never exits, so a
+     * phino that warns about anything at all would look like a phino that
+     * hangs, and the plugin would fall back to Docker with a message that
+     * says nothing about the executable it just found.</p>
+     *
      * @param expected This is the expected version
      * @return TRUE if available
      */
@@ -56,9 +62,14 @@ final class Phino {
                 () -> Phino.pump(stdout, proc.getInputStream())
             );
             pump.start();
+            final Thread errs = new Thread(
+                () -> Phino.pump(new ByteArrayOutputStream(), proc.getErrorStream())
+            );
+            errs.start();
             final boolean probed = proc.waitFor(3L, TimeUnit.SECONDS);
             if (probed) {
                 pump.join();
+                errs.join();
                 available = Phino.version(proc, stdout, expected, this);
             } else {
                 proc.destroyForcibly();


zaguzovmaksim0-hue (changes requested):
There is one lifecycle regression in the timeout path. This change adds a second reader thread (`errs`) but the `else` branch after `waitFor(3s)` only calls `destroyForcibly()` and returns; neither `pump` nor the new `errs` thread is joined there. That means a timed-out probe now can leave two blocked reader threads behind instead of one.

This matters especially because #901 / PR #917 is fixing the existing stdout-reader leak, and #917 as currently written only knows about `pump`; if #916 lands first, `errs` remains unhandled.

The simplest composition is the approach already suggested in #900: build the process with `redirectErrorStream(true)` and keep a single pump. Alternatively, both readers need cleanup on the timeout path. Please avoid introducing a second unowned reader thread here.

yegor256:
Merged, thanks! One leftover for later: on the timeout branch the new `errs` thread is not interrupted or joined, the way #901 fixed for `pump`.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (11) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5957](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (11) hits-of-code; +4 to give you at least something.  Your running score is [+359](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #917, opened 2026-09-10
#901: leave no reading thread behind when the probe times out

### LINKED ISSUE #901: `Phino.available()` leaves its reading thread alive after the probe times out

morphqdd:
On the timeout branch the process was killed and the thread reading its output
was left running. `available()` is called by `helpless()` and again by each
goal, so a multi-module build with a hanging phino accumulated one stuck thread
per goal per module for the life of the Maven JVM.

The reader is now interrupted and joined there as well.

Draft: it touches `Phino.java`, which #916 also changes.

Closes #901

### DIFF
--- a/src/main/java/org/eolang/hone/Phino.java
+++ b/src/main/java/org/eolang/hone/Phino.java
@@ -62,6 +62,8 @@ final class Phino {
                 available = Phino.version(proc, stdout, expected, this);
             } else {
                 proc.destroyForcibly();
+                pump.interrupt();
+                pump.join(TimeUnit.SECONDS.toMillis(1L));
                 Logger.info(
                     this,
                     "The 'phino --version' probe timed out, we must use Docker"


yegor256:
Merged, thanks!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5969](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #918, opened 2026-09-10
#892: compare the chain with the XMIR it is made from

### LINKED ISSUE #892: The skip check in `rewrite.sh` ignores the source XMIR, so a changed class keeps its old optimization
**What happens**

The skip check in `rewrite` compares the intermediate files with each other,
but never with the XMIR that the whole chain is derived from:

```bash
if [ -f "${pho}" ] && [ "${pho}" -nt "${phi}" ] && [ -f "${xo}" ] && [ "${xo}" -nt "${pho}" ]; then
  echo "... skipping transformation for ${idx}"
  return
fi
```

`xi` is the input, `phi` is made from it, `pho` from `phi`, and `xo` from
`pho`. The condition only says that the chain is internally consistent.

So on the second build of a project where one class changed:

1. `jeo:disassemble` writes a new `${xi}` for that class.
2. `${phi}`, `${pho}` and `${xo}` are all left over from the first build, and
   each is still newer than the one before it.
3. The condition holds, the file is skipped, and `jeo:assemble` reads the old
   `${xo}`.

The result is that the class is assembled back from the optimization of its
previous version. There is no warning, and the statistics row is not written
either, so nothing in the output says the new code was not looked at.

This is not the same as #837: there the file was half-written by a crashed
phino, and `atomic_write` fixed that. Here every file is complete and
consistent, the missing input is simply not part of the comparison.

**What should happen**

`${phi}` should be considered out of date when `${xi}` is newer than it, so
that a recompiled class is optimized again.

morphqdd:
The skip check compared the intermediates with each other and never with the
XMIR the whole chain is derived from, so on the second build of a project where
one class changed the leftovers of the first build were still consistent with
each other, the file was skipped, and `jeo:assemble` read the optimization of
the previous version of that class. Nothing said so.

The input is now the first link of the comparison.

Draft: it touches `rewrite.sh`, which #912 also changes.

Closes #892

### DIFF
--- a/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
@@ -112,8 +112,8 @@ function rewrite {
   mkdir -p "$(dirname "${phi}")"
   mkdir -p "$(dirname "${pho}")"
   mkdir -p "$(dirname "${xo}")"
-  if [ -f "${pho}" ] && [ "${pho}" -nt "${phi}" ] && [ -f "${xo}" ] && [ "${xo}" -nt "${pho}" ]; then
-    echo "Target $(basename "${pho}") is newer than source $(basename "${phi}") and output $(basename "${xo}") is present; skipping transformation for ${idx}"
+  if [ -f "${phi}" ] && [ "${phi}" -nt "${xi}" ] && [ -f "${pho}" ] && [ "${pho}" -nt "${phi}" ] && [ -f "${xo}" ] && [ "${xo}" -nt "${pho}" ]; then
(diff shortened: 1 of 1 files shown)


yegor256:
Merged, thanks!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5981](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #919, opened 2026-09-10
#893: let the exit code of the worker decide

### LINKED ISSUE #893: A rewrite that finishes as the timeout fires has its result replaced by the unoptimized input
**What happens**

`rewrite_with_timeout` decides what happened by looking at a flag file that
the watchdog creates when the deadline passes:

```bash
(
  sleep "${HONE_TIMEOUT}"
  : > "${flag}"
  ...
) &
watchdog=$!
wait "${sid}" || code=$?
...
if [ -f "${flag}" ]; then
  ...
  cp "${xi}" "${xo}"
```

The worker finishing and the watchdog firing are two independent events. When
the worker writes `${xo}` and exits just as the deadline passes, both are
true: the flag exists and the exit code is zero. The first branch wins, and
`cp "${xi}" "${xo}"` overwrites the optimized XMIR that was just written with
the unoptimized input.

The build stays green and the log says the file timed out, so the only visible
trace is that a class which was in fact optimized ships unoptimized.

The window is `HONE_TIMEOUT`-independent: it is the gap between the worker's
last write and its exit, so it is hit by whatever file happens to finish near
the deadline, which on a big project is not rare.

**What should happen**

The exit code of the worker should decide. When the worker exited normally,
its output should be kept even if the deadline passed in the meantime, and the
`cp` fallback should only run when the worker was actually killed.

morphqdd:
The flag file the watchdog creates was read before the exit code, so a file
that finished just as the deadline passed took the timeout branch and had its
optimized XMIR overwritten with the unoptimized input. The build stayed green
and the log said the file timed out.

The exit code decides now: a worker that exited normally keeps its output, and
the copy runs only when it was actually killed.

Draft: it touches `rewrite.sh`, which #912 and #918 also change.

Closes #893

### DIFF
--- a/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/rewrite.sh
@@ -238,18 +238,22 @@ function rewrite_with_timeout {
     kill "${watchdog}" 2>/dev/null || true
   fi
   wait "${watchdog}" 2>/dev/null || true
-  if [ -f "${flag}" ]; then
+  # The worker finishing and the deadline passing are independent events, so
+  # both can be true for a file that finished just as the watchdog fired. The
+  # exit code decides: a worker that exited normally wrote its output, and
+  # copying the input over it would ship the class unoptimized (#893).
+  if [ "${code}" -eq 0 ]; then
+    rm -f "${flag}"
+  elif [ -f "${flag}" ]; then
     rm -f "${flag}"
     sec=$(perl -E "say int($(now) - ${start})")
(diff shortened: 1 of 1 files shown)


yegor256:
Merged, thanks! Letting the exit code decide, rather than the flag, is the right reading of the race.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5993](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #920, opened 2026-09-10
#894: write a row for every file the pipeline considered

### LINKED ISSUE #894: Skipped files write no row to `hone-statistics.csv`, so `Optimized X/Y` counts the wrong total
**What happens**

`statistics_row` is called at one place in `rewrite`, after the rewriting is
done. Both early returns skip it:

```bash
if [ -f "${pho}" ] && [ "${pho}" -nt "${phi}" ] && ...; then
  echo "... skipping transformation for ${idx}"
  return
fi
...
  if [ "${rc}" -ne 0 ]; then
    cp "${xi}" "${xo}"
    echo "No grep-in match for ${idx} ... skipping"
    return
  fi
```

So `hone-statistics.csv` holds a row only for the files that phino actually
looked at. `OptimizeMojo` then reports:

```java
Logger.info(
    this,
    "Optimized %d/%d files",
    csv.count("Changed", CSV::positive),
    csv.size()
);
```

`csv.size()` is the number of rows, which is the number of files that were
processed, not the number of classes in the project. With the default
`grep-in` most classes never reach phino, so a run over 500 classes where 8
matched and 3 changed prints `Optimized 3/8 files`. The 492 classes that were
copied through are invisible, and the same number is what `SummaryMojo` adds
up across modules.

**What should happen**

Every file that the pipeline considered should get a row, with `Changed` set
to zero for the ones that were skipped, so that the denominator is the number
of classes and the summary across modules means what it says.

morphqdd:
`statistics_row` ran only after a file went through phino, so both early
returns left no row. `Optimized %d/%d files` counts rows, so a run over 500
classes where 8 reached phino and 3 changed printed `Optimized 3/8 files`, and
`SummaryMojo` added up the same number across modules.

A skipped file now writes its row with `Changed` set to zero.

Draft: it touches `rewrite.sh`, which #912, #918 and #919 also change.

Closes #894


yegor256:
Heads up: #918 just landed and rewrote the skip condition this PR adds a row after, so this one needs a rebase before it can go in. The approach looks right for #894; once rebased it should be straightforward.

---

morphqdd:
Rebased on top of #918, the row is now added inside the new skip condition. CI is green. @yegor256, please take another look.

yegor256:
The change is right and #894 is still open, but `src/main/resources/org/eolang/hone/scaffolding/rewrite.sh` conflicts and needs a rebase.

The skip branch you add a row to has grown a stamp check since you branched:

```bash
mark="${xo}.stamp"
if [ -f "${phi}" ] && ... && [ -f "${mark}" ] && [ "$(cat "${mark}")" == "${stamp}" ]; then
```

so your `statistics_row` line has to land inside the new condition rather than the old one. The second one, on the `grep-in` miss, applies as it is.

Three more changes have landed in the same file that you will meet on the rebase: #937 (`kill_tree` in the watchdog cleanup), #957 (the `^>` edit count) and #960 (the single-thread loop). None of them touches your two lines.

The body still says Draft — please say whether it is ready once rebased.

PR #921, opened 2026-09-10
#902: warn instead of failing when the image is not removed

### LINKED ISSUE #902: A failing `docker rmi` fails the build after the optimization has already succeeded
**What happens**

`RmiMojo` removes the image and lets any failure through:

```java
new Docker(this.sudo).exec(
    "rmi",
    "--no-prune",
    this.image
);
```

`Docker.exec` throws `IOException` on a non-zero exit code, and the mojo turns
that into a `MojoExecutionException`. `docker rmi` exits non-zero for reasons
that have nothing to do with the build: the image is used by a container that
is still around, another build removed it first, or it was never pulled
because this module had no classes.

So a build whose bytecode was optimized correctly fails at the very last step,
on a goal whose only purpose is to free disk space. The README recommends
adding `rmi` to the execution, so this is the configuration most users have.

**What should happen**

Failing to delete the image should be a warning, not a build failure. The
optimized classes are already in place by then and nothing about the build
depends on the image being gone.

morphqdd:
`docker rmi` exits non-zero for reasons that have nothing to do with the build:
the image is held by a container, another build removed it first, or it was
never pulled. `Docker.exec` turned that into a `MojoExecutionException`, so a
build whose bytecode was already optimized failed at the last step, on a goal
whose only purpose is to free disk space.

The failure is a warning now.

Closes #902

### DIFF
--- a/src/main/java/org/eolang/hone/RmiMojo.java
+++ b/src/main/java/org/eolang/hone/RmiMojo.java
@@ -37,12 +37,20 @@ public final class RmiMojo extends AbstractMojo {
                 this.image
             );
         } else {
-            new Docker(this.sudo).exec(
-                "rmi",
-                "--no-prune",
-                this.image
-            );
-            Logger.info(this, "Docker image '%s' was removed", this.image);
+            try {
+                new Docker(this.sudo).exec(
+                    "rmi",
+                    "--no-prune",
+                    this.image
+                );
+                Logger.info(this, "Docker image '%s' was removed", this.image);
+            } catch (final IOException ex) {
+                Logger.warn(
+                    this,
+                    "Docker image '%s' was not removed, the build is not affected by it: %s",
+                    this.image, ex.getMessage()
+                );
+            }
         }
     }
 }


zaguzovmaksim0-hue (approved):
I checked the control flow against the goal's purpose. The only operation moved under `try/catch` is `docker rmi`; a successful removal still logs the existing success message, while a non-zero Docker exit now becomes a warning containing both the image and the original error. The optimization/output has already completed before this goal, so cleanup failure should not invalidate the build.

The public method may still throw `IOException` from other future operations; this patch only downgrades the image-cleanup failure requested by #902. No correctness issue found. Approving.

yegor256:
Merged, thanks!

---

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (20) hits-of-code; +4 to give you at least something.  Your running score is [+136](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (20) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6005](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #922, opened 2026-09-10
#904: start the timings file fresh, with a header

### LINKED ISSUE #904: `hone-timings.csv` has no header and keeps the rows of previous runs
**What happens**

`Timings.through` appends a line and nothing ever creates a header or truncates
the file:

```java
Files.write(
    this.path,
    String.format(
        "%s,%d%n", name, System.currentTimeMillis() - start
    ).getBytes(StandardCharsets.UTF_8),
    StandardOpenOption.APPEND, StandardOpenOption.CREATE
);
```

Two consequences. The file has no header, so nothing says which column is the
name and which is the duration, and it cannot be read by the same code that
reads the other CSV this plugin writes. And because `target/` survives between
runs unless `clean` is used, a second `mvn hone:optimize` appends its rows to
the first one's, so the file holds several runs with no way to tell them apart.

The sibling file that `entry.sh` writes for the same purpose does have a
header, and uses a different separator and quoting on top of that:

```bash
echo '"Module";"Mojo";"Time"' > "${timings_csv}"
```

So the plugin produces two timing files in one `target/` directory in two
different formats.

**What should happen**

`hone-timings.csv` should be started fresh for each build and carry a header,
in the same format as the other timings file.

morphqdd:
`hone-timings.csv` was appended to and never started, so it had no header and a
second `mvn hone:optimize` without `clean` mixed its rows with the previous
run's. The other timings file in the same directory has a header and uses
quotes and semicolons.

The file is now replaced with a header when it is older than the JVM, which is
once per build, and the rows take the same quoting and separator as the sibling
file. The goals of one build still append to each other.

Closes #904

### DIFF
--- a/src/main/java/org/eolang/hone/Timings.java
+++ b/src/main/java/org/eolang/hone/Timings.java
@@ -66,16 +66,10 @@ final class Timings {
         }
     }
 
-    /**
-     * Start the file when it does not belong to this build yet.
-     *
-     * <p>The rows of a build that ran before this JVM are of no use here and
-     * "target" survives between runs unless "clean" is used, so a file older
-     * than the JVM is replaced by a fresh one with a header. The goals of this
-     * build write after that moment and append to each other.</p>
-     *
-     * @throws IOException If the file cannot be written
-     */
+    // The rows of a build that ran before this JVM are of no use here and
+    // "target" survives between runs unless "clean" is used, so a file older
+    // than the JVM is replaced by a fresh one with a header. The goals of this
+    // build write after that moment and append to each other.
     private void started() throws IOException {
         final boolean stale;
         if (Files.exists(this.path)) {
--- a/src/main/java/org/eolang/hone/Timings.java
+++ b/src/main/java/org/eolang/hone/Timings.java
@@ -7,6 +7,7 @@ package org.eolang.hone;
 import com.jcabi.log.Logger;
 import java.io.File;
 import java.io.IOException;
+import java.lang.management.ManagementFactory;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


zaguzovmaksim0-hue (approved):
I rechecked the updated revision after the earlier Qulice failures were fixed.

The lifecycle is consistent with the goal of #904: on the first timing write in a new Maven JVM, a missing/stale file is replaced with the same quoted semicolon-delimited header style used by the sibling timings CSV; later goals in that JVM see an mtime newer than the JVM start and append rather than truncate. The timing row itself now uses the same separator/quoting convention. The regression test forces an old mtime and verifies previous-run rows are removed, while the existing two-write test verifies rows from the current run accumulate under one header.

The previous private-method Javadoc and hardcoded newline Qulice violations are gone, and the full current CI is green. No correctness issue found. Approving.

yegor256:
Thanks. Two things stopped me from merging this one:

- It also changes the CSV format from `name,ms` to `"name";ms`. #904 asks only for a header and for the rows of previous runs to go; the separator and the quoting are a separate decision, and the file is a documented artifact that other things may read.
- Staleness is decided by comparing the file's mtime against `RuntimeMXBean.getStartTime()`. That holds for a plain `mvn` run, but not for a reused JVM (mvnd, or an IDE-embedded Maven), where the file of a previous build would look current and its rows would survive — the very thing the issue wants gone.

Assigning to @yegor256 to say whether the format change is wanted and whether the JVM-start heuristic is acceptable.

---

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+128](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #923, opened 2026-09-10
#906: say which file the summary could not read

### LINKED ISSUE #906: `Summary` ignores every walk failure except a symlink loop, so the report is silently partial
**What happens**

The file visitor of `Summary` reports only one kind of failure:

```java
@Override
public FileVisitResult visitFileFailed(
    final Path file, final IOException exc
) {
    if (exc instanceof FileSystemLoopException) {
        Logger.warn(this, "Symlink cycle detected at %s, skipping", file);
    }
    return FileVisitResult.CONTINUE;
}
```

Every other `IOException` - an unreadable directory, a broken symlink, a
permission error on a module's `target/` - is dropped on the floor and the
walk continues.

`SummaryMojo` is an aggregator goal, so the effect is that a module whose
statistics file could not be reached is simply missing from
`hone-statistics.csv`, and the `Optimized X/Y files` line at the end is
computed over the modules that happened to be readable. There is nothing in
the output to say the summary is partial.

**What should happen**

The other failures should be logged too, with the path and the reason, so that
a partial summary is visible as partial.

morphqdd:
The walk logged a symlink cycle and dropped every other failure, so a module
whose statistics file could not be read was simply missing from the summary and
`Optimized X/Y files` was counted over whatever happened to be readable.

Any other failure is now logged with the path and the reason.

Closes #906

### DIFF
--- a/src/main/java/org/eolang/hone/Collector.java
+++ b/src/main/java/org/eolang/hone/Collector.java
@@ -69,6 +69,12 @@ final class Collector extends SimpleFileVisitor<Path> {
                 "Symlink cycle detected at %s, skipping",
                 file
             );
+        } else {
+            Logger.warn(
+                this,
+                "Can't read %s, its statistics are missing from the summary: %s",
+                file, exc.getMessage()
+            );
         }
         return FileVisitResult.CONTINUE;
     }


zaguzovmaksim0-hue (approved):
I checked the visitor behavior against the issue. `visitFileFailed()` previously returned `CONTINUE` for every failure but emitted a warning only for `FileSystemLoopException`, so permission/broken-link/read failures could make the aggregate silently incomplete. This branch preserves the dedicated symlink-cycle warning and adds a warning for every other `IOException`, including both the failed path and exception message, while still continuing the walk as before.

That matches the requested behavior without changing aggregation/control flow. No correctness issue found in the change. Approving.

yegor256:
Merged, thanks!

---

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (6) hits-of-code; +4 to give you at least something.  Your running score is [+144](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (6) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6017](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #924, opened 2026-09-10
#896: Discover phr extra rules

### LINKED ISSUE #896: `entry.sh` looks only for `*.yml`, so every `.phr` rule is invisible to it
**What happens**

The rules of the `streams/` pipeline are `.phr` files. `Rules.discover()`
knows that:

```java
if (!path.endsWith(".phr") && !path.endsWith(".yml")) {
    continue;
}
```

Both places in `entry.sh` that look for rule files on disk look only for
`.yml`:

```bash
if [ -z "${RULES}" ]; then
fi
...
if [ -n "${EXTRA}" ]; then
  e=$(find "${EXTRA}" -name '*.yml' -exec "${RP}" {} \; | sort | tr '\n' ' ')
```

When the plugin runs, `RULES` comes from `OptimizeMojo` and holds the `.phr`
paths, so the first `find` never fires. But the image is documented as
something that can be run on its own, and then `RULES` is empty and the
fallback silently selects only the handful of `.yml` demo rules, so the whole
`streams/` pipeline is skipped and the run reports success with nothing
optimized.

The `EXTRA` scan has the same blind spot for anyone who mounts a directory of
`.phr` rules directly instead of going through `hone.extra`, which is the only
reason it works today: `OptimizeMojo.saveExtra` renames every extra rule to
`.yml` no matter what it was.

**What should happen**

Both `find` calls should accept `.phr` as well as `.yml`, so that a rule file
is picked up by the same set of extensions everywhere.

zaguzovmaksim0-hue:
Closes #896.

The main fallback rule scan already accepts both `.yml` and `.phr` on current `master` (via #882), but the `EXTRA` scan still filters only `*.yml`. This makes directly-mounted extra `.phr` rules invisible to the standalone scaffolding path.

This change makes the `EXTRA` scan use the same extension set as the main rule discovery: `.yml` or `.phr`.

Validation:
- `bash -n src/main/resources/org/eolang/hone/scaffolding/entry.sh`
- exercised the exact `find` expression against a directory containing one `.yml`, one nested `.phr`, and one unrelated `.txt`; only the two rule files were selected.

No other rule ordering or formatting behavior is changed.

### DIFF
--- a/src/main/resources/org/eolang/hone/scaffolding/entry.sh
+++ b/src/main/resources/org/eolang/hone/scaffolding/entry.sh
@@ -152,7 +152,7 @@ for rule in ${RULES}; do
   fi
 done
 if [ -n "${EXTRA}" ]; then
-  e=$(find "${EXTRA}" -name '*.yml' -exec "${RP}" {} \; | sort | tr '\n' ' ')
+  e=$(find "${EXTRA}" \( -name '*.yml' -o -name '*.phr' \) -exec "${RP}" {} \; | sort | tr '\n' ' ')
   if [ -n "${e}" ]; then
     echo "Extra rules found in ${EXTRA}: ${e}"
     RULES="${RULES} ${e}"


yegor256:
Merged, thanks!

---

0crat:
@zaguzovmaksim0-hue Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+315](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #925, opened 2026-09-10
#905: Report temporary cleanup failures

### LINKED ISSUE #905: `Mktemp` ignores the result of `File.delete()`, so a temporary directory can survive silently
**What happens**

`Mktemp` implements `Closeable` so that a temporary directory is removed after
use, and throws away what `delete()` says:

```java
@Override
public void close() throws IOException {
    try (Stream<Path> stream = Files.walk(this.dir)) {
        stream
            .map(Path::toFile)
            .sorted(Comparator.reverseOrder())
            .forEach(File::delete);
    }
}
```

`File.delete()` returns a boolean and never throws. When a file cannot be
removed, the method finishes normally and the caller is told the directory is
gone.

`BuildMojo` and `OptimizeMojo.withoutDocker()` both build their scaffolding in
one of these, so a build leaves a directory of scripts and rule files in the
system temp directory with nothing in the log about it.

**What should happen**

A file that could not be deleted should be reported, so that the leak is
visible instead of silent. `Files.delete` throws and says which path failed.

zaguzovmaksim0-hue:
Closes #905.

`Mktemp.close()` used `File.delete()` inside a stream and discarded its boolean result, so cleanup failures were silently treated as success.

This change materializes the reverse-ordered walk, closes the walk stream, and then deletes every path with `Files.delete()`. A failed deletion now propagates its `IOException` (including the failing path) through the existing `Closeable.close()` contract instead of disappearing. Closing the walk before deletion also avoids keeping traversal handles open while directories are removed.

Validation on the Java 8 production target:
- `javac --release 8` compiles `Mktemp.java`.
- A same-package probe created `a/b/file.txt` under `Mktemp`, called `close()`, and verified the root directory no longer exists.

The change is limited to cleanup semantics; temporary-directory creation and callers are unchanged.

### DIFF
--- a/src/main/java/org/eolang/hone/Mktemp.java
+++ b/src/main/java/org/eolang/hone/Mktemp.java
@@ -5,11 +5,12 @@
 package org.eolang.hone;
 
 import java.io.Closeable;
-import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Comparator;
+import java.util.List;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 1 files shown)


yegor256:
Merged, thanks!

---

0crat:
@zaguzovmaksim0-hue Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+339](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #926, opened 2026-09-10
#903: Record rmi goal timing

### LINKED ISSUE #903: The `rmi` goal is missing from `hone-timings.csv` because it is the only goal not timed
**What happens**

`PullMojo`, `BuildMojo` and `OptimizeMojo` record their duration through
`Timings.through`, and `RmiMojo` is the only goal that does not:

```java
@Override
public void exec() throws IOException {
    if (!this.alwaysWithDocker && new Phino().available(this.phino())) {
        ...
    } else {
        new Docker(this.sudo).exec("rmi", "--no-prune", this.image);
        Logger.info(this, "Docker image '%s' was removed", this.image);
    }
}
```

So `hone-timings.csv` never has a row for `rmi`, while the recommended
execution runs `build`, `optimize` and `rmi` together. Removing an image is
not instant, and the file that is supposed to say where the time went is
missing that part.

**What should happen**

`RmiMojo` should record its timing the same way the other three goals do.

zaguzovmaksim0-hue:
Closes #903.

`rmi` was the only regular Hone goal that bypassed `Timings.through`, so a normal `pull/build/optimize/rmi` run could not account for image-removal time in `hone-timings.csv`.

This wraps the full `RmiMojo.exec()` operation in `timings.through("rmi", ...)`, matching the existing pattern in `PullMojo`, `BuildMojo`, and `OptimizeMojo`. Both the Docker-removal branch and the branch that decides no image was built are therefore represented as execution time of the `rmi` goal.

The existing Docker integration test now also verifies that the generated timings file contains an `rmi,<milliseconds>` row.

Local Maven execution is unavailable in this Android/Termux environment; `git diff --check` passes and repository CI provides Maven/Docker validation.

### DIFF
--- a/src/main/java/org/eolang/hone/RmiMojo.java
+++ b/src/main/java/org/eolang/hone/RmiMojo.java
@@ -30,19 +30,24 @@ public final class RmiMojo extends AbstractMojo {
 
     @Override
     public void exec() throws IOException {
-        if (!this.alwaysWithDocker && new Phino().available(this.phino())) {
-            Logger.info(
-                this,
-                "Docker image '%s' was probably NOT built, that's why not removed either",
-                this.image
-            );
-        } else {
-            new Docker(this.sudo).exec(
-                "rmi",
-                "--no-prune",
-                this.image
-            );
-            Logger.info(this, "Docker image '%s' was removed", this.image);
-        }
+        this.timings.through(
+            "rmi",
+            () -> {
+                if (!this.alwaysWithDocker && new Phino().available(this.phino())) {
+                    Logger.info(
+                        this,
+                        "Docker image '%s' was probably NOT built, that's why not removed either",
+                        this.image
+                    );
+                } else {
+                    new Docker(this.sudo).exec(
+                        "rmi",
+                        "--no-prune",
+                        this.image
+                    );
+                    Logger.info(this, "Docker image '%s' was removed", this.image);
+                }
+            }
+        );
     }
 }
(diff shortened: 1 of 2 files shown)


yegor256:
Heads up: #921 just landed and rewrote the body of `RmiMojo.exec` that this PR wraps in `timings.through`, so this needs a rebase. The fix for #903 itself looks right; after the rebase the `rmi` call should stay inside the new try/catch so a failing `docker rmi` is still only a warning.

---

yegor256:
Not mergeable as it stands: \`src/main/java/org/eolang/hone/RmiMojo.java\` conflicts, and the resolution is not mechanical.

Master has since wrapped the removal in a \`try\`, so a Docker daemon that refuses to drop the image only warns:

\`\`\`java
try {
    new Docker(this.sudo).exec("rmi", "--no-prune", this.image);
    Logger.info(this, "Docker image '%s' was removed", this.image);
} catch (final IOException ex) {
    Logger.warn(
        this,
        "Docker image '%s' was not removed, the build is not affected by it: %s",
        this.image, ex.getMessage()
    );
}
\`\`\`

Your version drops that \`catch\`, so a failed \`docker rmi\` would break the build again. Please rebase on master and keep the \`try\`/\`catch\` inside the \`timings.through("rmi", ...)\` body.

The goal itself is right — #903 is still open and \`rmi\` is still the only untimed goal. Note also that #922 has just landed, so \`hone-timings.csv\` now carries a header and quotes its fields with \`;\` as the separator; the row your test looks for is \`"rmi";<ms>\`, not \`rmi,<ms>\`.

PR #927, opened 2026-09-10
#906: Warn about failed summary visits

### LINKED ISSUE #906: `Summary` ignores every walk failure except a symlink loop, so the report is silently partial

zaguzovmaksim0-hue:
Closes #906.

`Collector.visitFileFailed()` now reports every filesystem walk failure instead of silently continuing for everything except symlink cycles.

Symlink loops keep their existing dedicated warning. Other `IOException`s now log both the path that could not be visited and the exception message, while still returning `CONTINUE`, so the aggregator remains best-effort but no longer presents a partial summary as if it were complete.

`git diff --check` is clean. Maven is not installed in this Termux environment, so repository CI will run the Java quality/test matrix.

### DIFF
--- a/src/main/java/org/eolang/hone/Collector.java
+++ b/src/main/java/org/eolang/hone/Collector.java
@@ -69,6 +69,13 @@ final class Collector extends SimpleFileVisitor<Path> {
                 "Symlink cycle detected at %s, skipping",
                 file
             );
+        } else {
+            Logger.warn(
+                this,
+                "Failed to visit %s, skipping: %s",
+                file,
+                exc.getMessage()
+            );
         }
         return FileVisitResult.CONTINUE;
     }


yegor256:
Thanks for this. #923 landed the same warning in `Collector.visitFileFailed` a moment ago, so #906 is already covered and this one is now a duplicate. Closing it as such would be fine, unless you see something it still adds.

---

yegor256:
Thanks, but #906 was already closed by #923, which landed the same warning in `Collector.visitFileFailed`. This one is now redundant.

---

yegor256:
Issue #906 is already closed and master already warns on every walk failure. \`Collector.visitFileFailed()\` in \`src/main/java/org/eolang/hone/Collector.java\` reads

\`\`\`java
Logger.warn(
    this,
    "Can't read %s, its statistics are missing from the summary: %s",
    file, exc.getMessage()
);
\`\`\`

in the non-\`FileSystemLoopException\` branch, which is what this branch adds, and that is why it conflicts. Please close it.

PR #935, opened 2026-09-14
#895: count the rules the way the list separates them

### LINKED ISSUE #895: `entry.sh` always reports "1 rules" because the count looks for spaces in a newline-separated list
**What happens**

`entry.sh` reports how many rules it is about to use by counting the spaces in
the list:

```bash
printf 'Using the following %d rules:\n\t%b\n' \
  "$(( "$(echo "${RULES}" | grep -o ' ' | wc -l)" + 1))" \
  "${RULES// /\\n\\t}"
```

When `RULES` is not set by the caller, the same script fills it in from
`find`, which separates paths with newlines, not spaces:

```bash
```

There are then no spaces at all, so `grep -o ' ' | wc -l` gives 0 and the line
always reads `Using the following 1 rules:`, however many were found. The
substitution `${RULES// /\\n\\t}` has nothing to replace either, so the whole
list is printed as one blob on a single line.

**What should happen**

The count and the listing should be derived from the same separator the list
actually uses, so that the number matches the number of rule files and each
rule is printed on its own line.

morphqdd:
The rules summary counted spaces, so an empty list still read `Using the following 1 rules:`. The extra rules also came in with a trailing space from `tr`, and were glued to an empty list with a leading one, which added a phantom rule to the count and an empty line to the listing.

Now the count comes from `wc -w`, and the extra rules are trimmed before they are appended.

Closes #895


yegor256:
Merged, thanks! `wc -w` counts the list the way it is actually separated, and the empty case now reads as zero instead of one.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6097](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #951, opened 2026-09-15
#944: count the files, not the directories that hold them

### LINKED ISSUE #944: "Binaries saved" counts directories as files
## What happens

`entry.sh` reports how many class files it saved before the optimization:

```bash
printf 'Binaries saved to %s (%d files)\n' "${TARGET}/classes-before-hone" \
  "$(find "${TARGET}/classes-before-hone" -print | wc -l)"
```

`find -print` prints every entry, including the root directory and one entry per package
directory, so the number is not the number of files. For a normal Java layout with a directory per
package it is two to four times the real class count.

```
$ find cbh -print | wc -l
6
$ find cbh -type f | wc -l
3
```

## What should happen

The count should be `find ... -type f | wc -l`, so the number matches the word "files" next to it.

morphqdd:
`find -print` prints the root and every package directory as well, so the number next to the word "files" was two to four times the real class count.

It now counts `-type f` only.

Closes #944


yegor256:
Merged, thanks! Counting only regular files makes the "binaries saved" number honest.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6137](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #952, opened 2026-09-15
#949: strip whichever extension the rule file has

### LINKED ISSUE #949: Small steps logging prints every rule name with its `.phr` extension
## What happens

In small steps mode the rule name is printed after stripping a `.yml` suffix:

```bash
m=$(basename "${rule}" .yml)
```

109 of the 111 shipped rules are `.phr`, so nothing is stripped and every diagnostic line reads:

```
No changes made by '101-remove-self-reference-labels.phr'
```

```
$ basename 101-remove-self-reference-labels.phr .yml
101-remove-self-reference-labels.phr
```

The rule name is the only handle a user has when bisecting which rule corrupted a class, which is
the whole purpose of `hone.small-steps`.

## What should happen

Whichever extension is present should be stripped, for example `m=$(basename "${rule}"); m=${m%.*}`.

(Same `.yml` only assumption as #896 and #882, in a place they did not touch.)

morphqdd:
`basename "${rule}" .yml` is a leftover from the time rules were YAML only. 109 of the 111 shipped rules are `.phr`, so nothing was stripped and every small steps line read `No changes made by '101-remove-self-reference-labels.phr'`.

Whichever extension is there is stripped now.

Closes #949


yegor256:
Merged, thanks! Stripping whatever extension the rule carries makes the small-steps log usable again for `.phr` rules.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6153](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #956, opened 2026-09-15
#942: floor the elapsed time so a fast file cannot divide by zero

### LINKED ISSUE #942: `rewrite.sh` dies with "Illegal division by zero" when `now` falls back to `date`
## What happens

`now()` in `rewrite.sh` is deliberately written with a fallback, so whole second resolution is an
expected mode:

```bash
perl -MTime::HiRes=time -E 'printf "%.3f\n", time' 2>/dev/null || date '+%s'
```

The rate calculation that uses it has no guard:

```bash
per=$(perl -E "say int(${s_lines} / ($(now) - ${start}))")
```

When the fallback is in use and a file finishes inside the same second, the divisor is `0`, perl
dies with "Illegal division by zero", the assignment fails, `set -e` aborts the worker, and
`rewrite_with_timeout` reports it as a failure and refuses to copy the class through unoptimized.
The whole build then fails on a class that was rewritten correctly.

The fallback is reachable. `eclipse-temurin:25.0.4_7-jre-jammy`, the base image this Dockerfile
uses, ships `perl` but not `Time::HiRes`:

```
Can't locate Time/HiRes.pm in @INC
```

Inside the image it is masked by accident, because installing `parallel` pulls in `perl-modules`.
On the host path, where these scripts run directly, a machine with only `perl-base` (the Debian and
Ubuntu default) has nothing to mask it.

The exact fragment:

```
$ bash -c 'set -e -o pipefail; now(){ date "+%s"; }; start=$(now); s_lines=1234
  per=$(perl -E "say int(${s_lines} / ($(now) - ${start}))")'
Illegal division by zero at -e line 1.
$ echo $?
255
```

## What should happen

The elapsed time should be floored at one second, or the rate left out, when the clock has one
second resolution, so a fast file cannot fail the build.

morphqdd:
`now` is deliberately written with a fallback to `date`, so whole second resolution is an expected mode, but the rate calculation had no guard. A file that finished inside the same second gave a divisor of `0`, perl died, the assignment failed, `set -e` aborted the worker, and `rewrite_with_timeout` reported it as a failure and refused to copy the class through unoptimized. The build then failed on a class that was rewritten correctly.

The fallback is reachable: `eclipse-temurin:25.0.4_7-jre-jammy`, the base this Dockerfile uses, ships `perl` but not `Time::HiRes`. Inside the image it is masked by accident, because installing `parallel` pulls in `perl-modules`; on the host path nothing masks it.

The elapsed time is floored now, and the values go to perl through `@ARGV` so the shell cannot eat them.

Closes #942


yegor256:
Merged, thanks! Passing the numbers as arguments and flooring the interval also removes the interpolation into the perl program, which is a nice bonus.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6169](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #957, opened 2026-09-15
#945: count a replaced line once, not from both sides

### LINKED ISSUE #945: "N lines changed" counts every replaced line twice and can exceed the file length
## What happens

The number of changed lines is counted from both sides of the diff:

```bash
changed=$(diff "${phi}" "${pho}" | grep -cE '^[><]')
```

A replaced line shows up twice, once as `<` and once as `>`, so it counts as two. The value is then
printed against the real line count of the output:

```
${changed}/${s_lines} lines changed
```

so a rewrite that replaces every line reports two hundred percent. Two three line files differing
in all three lines give `changed=6` against `s_lines=3`.

The same number goes into the `Changed` column of `hone-statistics.csv`, which `CSV.count()` sums
for the summary report, so the aggregated figure is inflated as well.

## What should happen

Only one side should be counted, `grep -cE '^>'`, so the number can never exceed the line count it
is printed against.

(#708 fixed this same counting in `hone-it.sh`. `rewrite.sh` still has it, in three places.)

morphqdd:
`grep -cE '^[><]'` counts both sides of every hunk, so one replaced line counted as two. The value is printed against the real line count of the output, so a rewrite that replaced every line reported two hundred percent, and the same number went into the `Changed` column of `hone-statistics.csv`, which `CSV.count()` sums for the summary.

Only the `>` side is counted now, in all three places.

(#708 fixed the same counting in `hone-it.sh`.)

Closes #945


yegor256:
CI is green and the double counting in #945 is real, but I would rather not merge this one without a second opinion.

Counting only `^>` fixes the replaced line, which `diff` prints on both sides, but it also makes a pure deletion count as zero. A rule that only removes lines would then print `0 lines changed`, although `cmp -s` has already established that the file did change — a wrong number in the other direction. Something like the larger of the two counts, or `diff --unchanged-line-format`, would cover both shapes.

Assigning to @yegor256 to decide which number the line is meant to report.

---

PR #958, opened 2026-09-15
#946: copy phino without a pipe that exits early

### LINKED ISSUE #946: The image build fails at random with code 141 because `find | head -1` runs under `pipefail`
## What happens

The builder stage sets:

```dockerfile
SHELL ["/bin/bash", "-e", "-c", "-o", "pipefail"]
```

and later copies the built binary with:

```dockerfile
RUN find /root -name phino -type f 2>/dev/null | head -1 | xargs -I {} cp {} /usr/local/bin/phino
```

`head -1` exits after the first match while `find` is still walking the cabal store. The next write
`find` attempts gets EPIPE, `find` dies with SIGPIPE, and `pipefail` turns that into exit code 141,
so the `RUN` fails even though the copy succeeded. Whether it fires depends on how many `phino`
files cabal leaves under `/root` and on when `find` writes, so the image build is red at random.

```
$ bash -c 'set -o pipefail -e; find / -name "*" | head -1 >/dev/null'; echo $?
141
```

## What should happen

The early exit pipe should be avoided, for example
`cp "$(find /root -name phino -type f | sort | tail -1)" /usr/local/bin/phino`, or `find ... -quit`,
so a successful copy cannot fail the build.

morphqdd:
The builder stage runs under `SHELL ["/bin/bash", "-e", "-c", "-o", "pipefail"]`, and `head -1` exits after the first match while `find` is still walking the cabal store. The next write `find` attempts gets EPIPE, `find` dies with SIGPIPE, and `pipefail` turns that into exit 141, so the `RUN` failed even though the copy had succeeded. Whether it fires depends on how many `phino` files cabal leaves behind and on when `find` writes, so the image build was red at random.

The copy now takes the last match of a completed `find`, with no early exit pipe.

Closes #946


yegor256:
The `head -1` under `pipefail` really is the cause of the random 141 in #946, and replacing the pipe with a command substitution is the right shape.

Two things stop me from merging it unreviewed. `sort | tail -1` does not just avoid the pipe, it also changes which binary is picked when `find` returns more than one, from the first encountered to the last in sort order — worth stating that this is intended. And `showsThatSuchAPipeReallyFails` walks the whole filesystem with `find /` on every build to demonstrate a property of `pipefail`; the same thing can be shown against a directory the test creates.

Assigning to @yegor256 for a look.

---

PR #960, opened 2026-09-16
#950: run a task with its own input, through env bash

### LINKED ISSUE #950: The single thread task loop hardcodes `/bin/bash` and hands the child its own task list
## What happens

The single thread branch of the task loop runs each task like this:

```bash
while IFS= read -r cmd; do
  ...
  /bin/bash -c "${cmd}"
done < "${tasks}"
```

Two problems.

`/bin/bash` is hardcoded, while both scripts start with `#!/usr/bin/env bash` precisely so they do
not depend on that path (the concern behind #867). On a host without `/bin/bash` this line fails
once per file with "No such file or directory", while the parallel branch works fine.

The loop's standard input is the task file, and the child inherits it. Any command in the chain
that reads standard input consumes the task lines that were supposed to run next, so classes are
silently left unoptimized and the loop ends early.

## What should happen

The task should run with its standard input detached, `bash -c "${cmd}" </dev/null`, resolved
through `env bash` rather than a hardcoded path, and the loop should read the task file on its own
file descriptor.

morphqdd:
The single thread branch ran each task with `/bin/bash`, which the shebang of the script deliberately avoids, and let the child inherit the task list on standard input. Now the loop reads the task file on its own descriptor and the task runs with standard input detached.

Closes #950


morphqdd:

PR #961, opened 2026-09-16
#947: set only a locale the host actually has

### LINKED ISSUE #947: The scripts force `en_US.UTF-8` on the host without checking that it exists
## What happens

Both scripts force a locale without checking that it exists:

```bash
if [ "${LANG}" != 'en_US.UTF-8' ]; then
  LANG=en_US.UTF-8
  export LANG
  LC_ALL=en_US.UTF-8
  export LC_ALL
  ...
fi
```

Inside the image that locale is generated, so it works. But the same code runs on the user's own
machine when there is no Docker, and minimal Debian, Ubuntu, Alpine and NixOS have only `C.UTF-8`.
Every child process then warns and falls back to C:

```
$ docker run --rm ubuntu:24.04 bash -c 'LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 perl -e "print 1"'
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings: ...
```

`now()` forks perl at least twice per file, so a module with a thousand classes prints thousands of
four line warning blocks into the build log.

## What should happen

The locale should be set only when `locale -a` lists it, with `C.UTF-8` as the fallback.

morphqdd:
Both scripts forced `en_US.UTF-8` without checking that the host has it. Outside the image most systems have only `C.UTF-8`, and every forked perl then warns. Now the locale is taken from what `locale -a` lists, with `C.UTF-8` as the fallback, and nothing is exported when neither is there.

Closes #947


morphqdd:

PR #962, opened 2026-09-16
#943: invalidate the skip when a rule or an option changes

### LINKED ISSUE #943: Editing a rule changes nothing without `clean`, because the skip check ignores the rules
## What happens

The skip check in `rewrite.sh` compares only the file chain:

```bash
if [ -f "${phi}" ] && [ "${phi}" -nt "${xi}" ] && [ -f "${pho}" ] && [ "${pho}" -nt "${phi}" ] \
   && [ -f "${xo}" ] && [ "${xo}" -nt "${pho}" ]; then
```

Nothing in it depends on the rules or on the configuration. On a build without `clean`, a user who
edits a `.phr` rule, adds an extra rule through `hone.extra`, or widens `hone.grep-in` gets
"Output ... is newer than input ... skipping transformation" for every unchanged class, and ships
the bytecode the previous run produced. The rule edit appears to do nothing.

The same path defeats a narrowed `grep-in` in the other direction. A class optimized in the first
run is skipped here, before the grep-in test further down ever runs, so it stays optimized after
the user excluded it.

## What should happen

The skip should also compare against the newest modification time of the rule files, and be
invalidated when the options that change the result (`hone.rules`, `hone.grep-in`,
`hone.small-steps`, `hone.max-cycles`, `hone.max-depth`) differ from the ones the output was
produced with, for example through a stamp file written next to the output.

(#892 added the input XMIR to this chain. The rules and the options are still outside it.)

morphqdd:
The skip check compared only the file chain, so an edited rule, an extra rule or a changed option left every class as the previous run produced it. The rules and the options that change the result are now written into a stamp file next to the output, and the skip happens only when that stamp still matches.

Closes #943


morphqdd:

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5700](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #963, opened 2026-09-16
Remove stale summary reports

### LINKED ISSUE #938: `hone:summary` leaves a stale report when no statistics files are found
## What happens

`Summary.collect()` returns the previous `hone-statistics.csv` unchanged when the current tree contains no statistics files. A summary from an earlier build can therefore be presented as if it described the current build.

## How to reproduce

1. Run the `hone:summary` goal once with a child module that produces `hone-statistics.csv`; keep the generated report.
2. Remove the child statistics files (or run against a clean tree) while leaving the old report at the configured target path.
3. Run `Summary.collect()` again.

`src/main/java/org/eolang/hone/Summary.java` only calls `Summary.reduce(found).flush(destination)` inside `if (!found.isEmpty())`. When `found` is empty, the old destination is neither removed nor replaced.

## Expected result

A run with no input statistics should remove the old report or write an explicit empty report, and the goal should report that no current data was collected.

## Evidence

The returned path still points to the existing file, so downstream consumers read stale rows without any failure or warning.

gemshrine:
Closes #938

### Problem
`hone:summary` left the previous `hone-statistics.csv` in place when the current reactor contained no module reports. Consumers could therefore read statistics from an earlier build as if they belonged to the current one.

### Solution
Delete the destination report when collection finds no input files. A current run can no longer expose stale rows after module statistics disappear.

### Verification
- Added a regression test that creates a report, removes all source reports, and verifies the destination is removed.
- `mvn -q -Dtest=SummaryTest test` passes.
- `git diff --check` passes.


yegor256:
The fix looks good and CI is green, but this branch has merge conflicts with `master` and cannot be merged as it stands. Could you merge `master` in and resolve them? I will merge afterwards.

yegor256:
Master already does this. \`Summary.collect()\` has carried

\`\`\`java
if (found.isEmpty()) {
    Files.deleteIfExists(destination);
} else {
    Summary.reduce(found).flush(destination);
}
\`\`\`

for a while now (\`src/main/java/org/eolang/hone/Summary.java\`), which is the same behaviour your \`Summary.remove()\` adds, and that is why the branch no longer merges. Issue #938 is still open only because nobody closed it.

If \`SummaryTest\` gained a case that master lacks, it is worth keeping: rebase and send the test alone. Otherwise please close this one.

PR #964, opened 2026-09-16
Preserve spaces in extra rule paths

### LINKED ISSUE #940: `hone:optimize` splits extra rule paths containing spaces

gemshrine:
Closes #940

### Problem
Extra rule paths were serialized as a space-separated list and then split by the shell. A directory such as `src/rules/a few` was consequently passed as two paths and its rules were not loaded.

### Solution
Serialize rule paths one per line and read them as complete lines in both scaffolding scripts. The generated and extra rule lists now preserve spaces through environment variables, validation, and phino invocation.

### Verification
- Updated the extra-rule integration fixture to use a directory containing a space.
- `bash -n` passes for both changed scaffolding scripts.
- `mvn -q -DskipTests package` passes.
- `git diff --check` passes.


yegor256:
CI is green here and the fix is correct in itself, but I am not confident enough to merge it without a second pair of eyes: it changes the wire format of `HONE_RULES` between `entry.sh` and `rewrite.sh` from space-separated to newline-separated, and that list is the scheduling mechanism for the whole rewrite pipeline. `mapfile`/`<<<` also make `rewrite.sh` strictly bash-only, where the previous `read -a` was not.

Assigning to @yegor256 for a manual look.

---

yegor256:
Not mergeable: `src/main/resources/org/eolang/hone/scaffolding/entry.sh` conflicts, in both places you touch.

#935 rewrote the rule listing in the meantime, and master now counts with

```bash
printf "Using the following %d rules:\n\t%b\n" \
  "$(printf "%s" "${RULES}" | wc -w)" \
  "${RULES// /\\n\\t}"
```

which is the very word-splitting #940 is about, so your newline-separated list still has a place — it just has to be rebased onto the new listing rather than onto the old one. The extra-rules block below it conflicts for the same reason.

Worth checking while you rebase: `rewrite.sh` has also moved (#937, #957, #960), and #959 added a `rm -rf "${TARGET}/hone/..."` line to `entry.sh` just above the disassembly options.

PR #965, opened 2026-09-16
Escape paths in statistics CSV rows

### LINKED ISSUE #939: rewrite statistics CSV becomes malformed when a class path contains a quote
## What happens

The optional `hone-statistics.csv` output is assembled by interpolating file paths into a quoted CSV line without escaping embedded double quotes. POSIX permits `"` in file names, so one class path can break the record and make the CSV unreadable by the plugin summary.

## How to reproduce

1. Create or compile a class whose source path contains a double quote, for example `src/main/java/a"b/Example.java`, so the generated XMIR path contains that character.
2. Run the optimization with `-DHONE_STATISTICS=true` (the `HONE_STATISTICS=true` scaffold setting).
3. Inspect `target/hone-statistics.csv` and parse it with a CSV reader.

`src/main/resources/org/eolang/hone/scaffolding/rewrite.sh` writes rows as `"${idx},\"${phi}\",\"${pho}\",${changed},${per}"` and never doubles quotes in `phi` or `pho`.

## Expected result

Every path should be escaped according to CSV rules before it is appended, so paths containing quotes, commas, or newlines remain one field and `hone:summary` can merge the report.

## Evidence

The generated line has an unescaped quote inside a quoted field. Apache Commons CSV in `CSV.java` then interprets the remainder of the row as syntax/data in the wrong columns, silently corrupting statistics or failing the summary.

gemshrine:
Closes #939

### Problem
The rewrite scaffold interpolated PHI paths directly into quoted CSV fields. A path containing a double quote therefore terminated the field early and produced malformed statistics for `hone:summary`.

### Solution
Build each statistics row from separate fields and double embedded quotes before writing the quoted path values. Commas and newlines remain protected by the surrounding CSV quotes.

### Verification
- `git diff --check` passes.
- The row writer now receives each field separately, so paths are escaped before `printf` writes them.


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3520](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #966, opened 2026-09-16
Kill the rewrite session on timeout

### LINKED ISSUE #931: `hone.timeout` never stops a hanging rule pass: the watchdog kills the worker's first child, which has already exited
**What happens**

`hone.timeout` doesn't stop a phino rule pass that runs too long. The watchdog fires on time, but it signals a process that has already exited, so the hanging phino runs to its natural end. The file is then still reported as a timeout and copied through unoptimized.

I ran `rewrite.sh rewrite_with_timeout` from `7fce962` in `ubuntu:24.04` (util-linux `setsid`, procps `pgrep`) with `HONE_TIMEOUT=5` and `HONE_KILL_GRACE=1`. A stub `phino` answers `--input=xmir` after 1 second, the way a real XMIR-to-PHI conversion takes a moment, and then the rule pass sleeps for 60 seconds. I added one `echo` to print what the watchdog picked:

```
  watchdog armed: worker sid=320, leader=340 (bash /tmp/bin/phino rewrite --input=xmir)
conversion xmir->phi pid=340 (takes 1s)
rule pass pid=361 (hangs for 60s)
Modified 1/1 A.phi (4.0K): 1/0 lines changed, 0 lps
Timeout in 1/1 A.xmir (4.0K) after 61 seconds
run 2: returned after 61s with HONE_TIMEOUT=5
```

It's the same on all three runs: 62s, 61s, 61s. In the other two runs `leader` was an even shorter-lived child (its `ps` line was already empty). `ps` during the run shows that the worker is the session leader itself, `PID=PGID=SID`, running `bash`.

**Why**

`rewrite_with_timeout` (`rewrite.sh:209-232`) starts the worker with `"${SETSID}" --wait "${0}" rewrite "$@" &` and then treats the first child of `$!` as the process group to kill:

```bash
leader=$(pgrep -P "${sid}" 2>/dev/null | head -n 1) || true
...
kill_tree TERM "${leader}"
kill -TERM "-${leader}" 2>/dev/null || true
```

A background job of a non-interactive script isn't a process group leader, so util-linux `setsid` doesn't fork. It calls `setsid()` and `exec`s the worker in place. So `$!` is already the new session and group leader (the bash worker), and `pgrep -P` returns the worker's first child. That is `mkdir`/`dirname`, or the XMIR-to-PHI `phino`, never the rule pass that comes later. By the time the deadline comes, that PID is gone. `kill_tree` finds nothing, and `kill -TERM -<pid>` names a group that doesn't exist. The worker and the phino it's waiting on aren't touched.

When the timeout does work in a quick test, it's only because the poll happened to catch the rule pass as the first child. That doesn't happen when the conversion takes longer than the 50 ms polling step.

**How to fix**

Signal the session that `setsid` created, which is `$!` itself:

```bash
"${SETSID}" --wait "${0}" rewrite "$@" &
sid=$!
...
kill -TERM "-${sid}" 2>/dev/null || true
sleep "${HONE_KILL_GRACE:-10}"
kill -KILL "-${sid}" 2>/dev/null || true
```

`kill_tree` can stay as a fallback, but it should start from `${sid}` rather than from a child picked once at start. If you want to support a `setsid` that does fork, use `ps -o pgid= -p <pid>` to find the group instead of guessing it. A test can reuse the stub above: with `HONE_TIMEOUT=5` and a rule pass that sleeps 60s, `rewrite_with_timeout` must return in about `5 + HONE_KILL_GRACE` seconds, and no `sleep` may stay behind.

gemshrine:
Closes #931

### Problem
The timeout watchdog selected the first child of the `setsid` worker as the process group to kill. That child can exit during XMIR conversion, leaving the long-running rule pass alive and allowing a timed-out rewrite to run for its natural duration.

### Solution
Use the process group belonging to the `setsid` worker itself and terminate the worker tree and group during timeout escalation. The watchdog now targets the session leader rather than a transient child.

### Verification
- `git diff --check` passes.
- The timeout path now signals the full session before the grace-period escalation.


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3532](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #967, opened 2026-09-16
Stop rewrite workers after a parallel failure

### LINKED ISSUE #932: After one file fails with threads > 1, the rewrites of other files keep running as orphans after `rewrite.sh` exits
**What happens**

When one file fails in a multi-threaded `optimize`, `rewrite.sh` exits right away, but the rewrites still running for other files keep going in the background. Their `rewrite` workers and `phino` processes are reparented to init and outlive the script, and when hone runs without Docker they also outlive the Maven build.

I ran `rewrite.sh` from `7fce962` in `ubuntu:24.04` with GNU Parallel, `HONE_THREADS=2` and two XMIR files. A stub `phino` fails the conversion of `B.xmir` and hangs for 120s on the rule pass of `A.xmir`:

```
Starting to rewrite 2 file(s) in 2 thread(s)...
conversion of /tmp/in/B.xmir fails
Failure (exit code 1) in 2/2 B.xmir (4.0K) after 0 seconds; refusing to copy it through unoptimized
parallel: This job failed:
/tmp/rewrite.sh rewrite_with_timeout 2/2 /tmp/phi/B.phi /tmp/pho/B.phi /tmp/in/B.xmir /tmp/out/B.xmir
rewrite.sh exited after 1s
processes left behind after rewrite.sh exited:
 3491     1  3491       3 bash /tmp/rewrite.sh rewrite 1/2 /tmp/phi/A.phi /tmp/pho/A.phi /tmp/in/A.xmir /tmp/out/A.xmir
 3566  3491  3491       2 sleep 120
```

The worker for `A` (PPID 1, its own session 3491) and its phino are still there after `rewrite.sh` and GNU Parallel have returned. They'll go on burning CPU and will later write `phi-optimized/A.phi` and `unphi/A.xmir` into `target/hone`, after the build has already failed.

Inside the Docker image this is hidden: the container stops when `entry.sh` exits, and that kills everything. But `OptimizeMojo.optimize()` runs `withoutDocker()` whenever a local `phino` of the right version is available (`OptimizeMojo.java:452-455`), and then nothing reaps them. The `Mktemp` directory holding `rewrite.sh` is even deleted under them.

**Why**

Each job is `rewrite.sh rewrite_with_timeout ...`, and that starts the real work in a new session, `"${SETSID}" --wait "${0}" rewrite "$@" &`, plus a watchdog subshell (`rewrite.sh:211` and `220-230`). When `--halt=now,fail=1` fires (`rewrite.sh:369-372`), GNU Parallel signals the jobs it started, which are the `rewrite_with_timeout` shells. `rewrite_with_timeout` has no `trap`, so it dies without passing the signal on. Because of `setsid` the worker is in another session and process group, so the signal to the job's group doesn't reach it. That separation is what the timeout code relies on (#727), and here it keeps the worker alive.

**How to fix**

Give `rewrite_with_timeout` a trap that tears down what it started, reusing the timeout path:

```bash
trap 'kill -TERM "-${sid}" 2>/dev/null; kill "${watchdog}" 2>/dev/null; sleep "${HONE_KILL_GRACE:-10}"; kill -KILL "-${sid}" 2>/dev/null; exit 143' TERM INT HUP
```

It has to be set before `"${SETSID}" ... &` starts. Signal `-${sid}`, the session `setsid` created, and not the first child (see #931). A test can reuse the stub: with two threads, one failing file and one hanging rule pass, no `rewrite`, `phino` or `sleep` process may remain once `rewrite.sh` has returned.

gemshrine:
Closes #932

### Problem
When GNU Parallel stopped the rewrite batch after one file failed, the interrupted `rewrite_with_timeout` process did not forward the signal to the `setsid` worker it had started. That worker and its phino process could survive the Maven build and continue writing output.

### Solution
Install a signal handler around each timed worker. On termination it stops the worker process group and watchdog, waits through the configured grace period, escalates to `KILL`, and exits with the interrupted status.

### Verification
- `git diff --check` passes.
- The cleanup path targets both the worker session and the watchdog process.


yegor256:
@gemshrine conflict here

gemshrine:
@yegor256 the conflict is resolved. I rebased the branch onto the current master and pushed the updated commits.

PR #968, opened 2026-09-16
Reject streams optimization for old bytecode

### LINKED ISSUE #930: `streams/*` rules emit Java 16 `mapMulti` into classes compiled for Java 8 or 11, which then fail with `NoSuchMethodError`
**What happens**

The `streams/*` rules rewrite a pipeline into a `mapMulti` call no matter which Java version the class was compiled for. `mapMulti` and its primitive siblings (`Stream.mapMulti`, `mapMultiToInt`, `mapMultiToLong`, `mapMultiToDouble`, `IntStream.mapMulti`, ...) exist only since Java 16. A class compiled with `--release 8` or `11` keeps its class file version after `hone:optimize`, but now links against a method that its target runtime doesn't have. On a Java 8 to 15 JVM, the first call of the optimized method fails with `NoSuchMethodError`.

The plugin's own integration test already produces such a class. `src/it/bench/pom.xml` compiles with

```xml
<maven.compiler.source>1.8</maven.compiler.source>
<maven.compiler.target>1.8</maven.compiler.target>
```

and `src/it/bench/verify.groovy` requires the fused call to be there:

```groovy
String log = new File(basedir, 'target/hone/phi-optimized/org/eolang/bench/App.phi').text;
assert log.contains("mapMulti")
```

So `App.class` stays major version 52 (Java 8) and calls a Java 16 API. The test passes only because CI runs the result on a modern JDK. A library that targets Java 8 or 11 and is published after `hone:optimize` breaks for every user on an older runtime.

**Why**

Nothing in the rules or in the mojo looks at the class version. No rule uses `𝑒-class-version` anywhere except to copy `version ↦ 𝑒-class-version` through unchanged, and the rules that produce `mapMulti` have no condition on it at all: `4xx/451-fuse-distill-unbox`, `455-flatMap-to-mapMulti`, `456`/`457`/`458` (the primitive `flatMapTo*`), `461-fuse-mapMulti`, `5xx/501-distill-to-mapMulti`, `502-distill-state-to-mapMulti`, and the `6xx` group that lowers the `Φ.hone.mapMulti` pragma into the real `Stream.mapMulti` call. `OptimizeMojo` and `rewrite.sh` don't filter classes by version either. The README only says Java 11+ is needed to build the plugin, and doesn't say optimized code needs Java 16 at runtime.

**How to fix**

Any of these would close it:

1. skip the fusion rules for a class whose major version is below 60 (Java 16), by a `where` condition on `𝑒-class-version` in the rules that emit `mapMulti`, or by leaving such classes out before `rewrite.sh`;
2. or refuse to run `streams/*` when `maven.compiler.release` or `target` is below 16, with a clear message;
3. at the very least, document that the optimized bytecode requires a Java 16+ runtime, and change the `bench` IT to `release 16` or add a negative IT with `release 8` that checks the class is left alone.

gemshrine:
Closes #930

### Problem
The `streams/*` rules emit calls to `Stream.mapMulti` and its primitive variants, which were added in Java 16. Applying those rules to classes compiled for Java 8 or 11 leaves bytecode that fails with `NoSuchMethodError` on the target runtime.

### Solution
When streams rules are selected, inspect the class-file major versions before rewriting. The scaffold now stops with a clear diagnostic if any target class is older than version 60 (Java 16), preventing incompatible optimized bytecode from being emitted.

### Verification
- `bash -n src/main/resources/org/eolang/hone/scaffolding/entry.sh` passes.
- `git diff --check` passes.
- The guard reports the offending class path and version before optimization starts.


0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (38) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3544](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #980, opened 2026-09-17
Fuse method references whose return the SAM unboxes

### LINKED ISSUE #974: dropWhile, primitive mapMulti and shape-crossing chains are never fused, and the README does not say so
Three sink-free shapes never fuse today, each ruled out by a rule header rather than by a limit of the approach. They are worth tracking together, because the README and the fusion story both read as though every sink-free operation is covered.

`208-lambda-to-dropwhile.phr` handles the object `java/util/stream/Stream` with a `static` predicate whose target returns `Z`. Its header names what that excludes: primitive streams, captured predicates, and method references to `Boolean`-returning methods, which is the shape `305` already handles for `filter`. So `IntStream.dropWhile(i -> i < n)` and a `dropWhile` over a captured variable both stay unfused.

`461-fuse-mapMulti.phr` merges adjacent `mapMulti` stages only in their reference form: "The primitive `mapMultiToInt/Long/Double` variants and capturing lambdas are left for a follow-up." A user who has already written `mapMultiToInt(...).mapMultiToInt(...)`, which is what the plugin's own output looks like, gets no merge.

There is also a shape gap with no target to aim at. `IntStream` has no `mapMultiToObj`, so a chain that crosses from a primitive stream to an object one, such as `IntStream.range(0, n).mapToObj(Integer::toString).filter(s -> !s.isEmpty())`, has no single `mapMulti` form to collapse into. Whatever the eventual answer is, splitting at the boundary or lowering through `boxed()`, it is not written down anywhere today.

Lifting the `static`-only and `Z`-only guards in `208` to reuse the `305` object-predicate path is the smallest of the three and would close the most common `dropWhile` case. The other two are larger and can stay as follow-ups, but they belong in the README's list of what is and is not fused, next to the existing parallel and short-circuit caveats. Issue #635 covers the related case of operations taking a functional object instead of a lambda.

yegor256:
A new rule, `108-static-methodref-unbox-to-lambda`, desugars a static
method reference whose target returns a wrapper where the functional
interface wants the primitive. `X::small` for `static Boolean
small(Integer)` behind a `Predicate` becomes a synthesised
`lambda$hone$108N` wrapper with the `booleanValue()` spelled out, and
from there it is an ordinary lifted lambda that 111 and the rest of the
pipeline fuse.

`LambdaMetafactory` inserts that unboxing itself, so the invokedynamic's
instantiated type and its target signature differ, and 104 declines
exactly that difference. Until now the reference was a fusion barrier:
the operations on either side of it each fused into their own
`mapMulti` and the native reference sat between them. The new fixture
`methodref-unbox.yml` measures it — three pipelines lose an
invokedynamic each — and covers both the `dropWhile` case the issue
names and the same reference behind a `filter` and a `mapToInt`.

A return needing widening on top of the unboxing (an `Integer`-returning
method behind a `ToLongFunction`, which wants `intValue()` and then
`i2l`) still stays native, and the fourth pipeline in the fixture plus a
second phino pack pin that refusal — build the wrapper anyway and the
class dies at link time. The other gaps the issue lists (capturing
`dropWhile`, primitive-stream `dropWhile`, adjacent primitive `mapMulti`
and shape-crossing chains) are now written down in README with examples,
as the issue asked, and stay as follow-ups.

Closes #974


yegor256:
@yegor256 could you take a look when you get a moment? Happy to explain any part of the new rule or the fixture counts if something reads oddly.

PR #986, opened 2026-09-17
fix(#984): fuse the mapToObj crossing from a primitive stream

### LINKED ISSUE #984: A chain crossing from IntStream to Stream cannot fuse, because there is no mapMultiToObj
`IntStream.range(0, n).mapToObj(Integer::toString).filter(s -> !s.isEmpty())` fuses on neither side of the crossing. No rule under `src/main/resources/org/eolang/hone/rules/streams/` names `mapToObj` at all — `grep -rl mapToObj` on that tree is empty — so the `Φ.hone.lambda` carrying it is never recognised and the operations before and after it stay two independent runs.

Recognition alone would not be enough, which is what makes this different from the other gaps. The 5xx emitters lower a distill into `Stream.mapMulti`, `Stream.mapMultiToInt` and their siblings, and `IntStream` has no `mapMultiToObj`, so a distill that consumes ints and produces objects has no single JDK call to become. `IntStream.range(0, n).mapToObj(...)` is the everyday foreach-with-index idiom, so the barrier shows up often.

The smallest fix is a 2xx rule that rewrites `IntStream.mapToObj(f)` into `IntStream.boxed()` followed by `Stream.map(f)`, which turns the crossing into a reference chain of the shape the boxed() rules (`106-remove-boxed-label.phr` and `209-unbox-box-to-distill.phr`) already handle, with no new emitter needed. For the record, #538 asked for `mapToObj` recognition and was closed in May 2026 without a rule landing.

yegor256:
Closes #984.

`IntStream.range(0, n).mapToObj(f).filter(p)` fused on neither side of the crossing. No rule named `mapToObj`, and — as the issue points out — recognising it alone would not have been enough: the 5xx emitters lower a distill into `Stream.mapMulti` and its siblings, and there is no `IntStream.mapMultiToObj` for a distill that consumes ints and produces objects to become.

## What it does

**`2xx/208-mapToObj-to-boxed-map`** moves the crossing instead of emitting it. The operator becomes two pragmas that already exist: the `Φ.hone.box` of a `boxed()`, and a plain `Φ.hone.map` carrying the original mapper — one whose `bridge-input` is the wrapper the box produces while its `accepted` is still the primitive the desugared `lambda$` declares. That mismatch is not a lie the rule has to resolve: it is the shape `231` already mints for a primitive operator split by `211`, and the rules that read it (`241`–`243` between two pragmas, `421` at the head of a fused run) are the ones that decide, once the chain around the crossing has settled, whether an unwrapping is needed. The map lands on the reference side, so `281` gives it the dup a following filter eats, `306` folds it, `401` fuses it and `501`/`511` lower the run to one `mapMulti`. Nothing downstream learns a new shape.

Letting them decide is the point, because whether the box survives depends on what sits in front of the crossing and the rule cannot see it:

* nothing fusable in front — the box is alone, `602` emits the `boxed()`, and `421` prepends the `intValue()` at the head of the run;
* a primitive operator in front (`LongStream.map(g).mapToObj(f)`) — `221` cancels `211`'s trailing unbox against this box, and the element never leaves the primitive domain;
* a `mapToX(lambda$)` in front (`list.stream().mapToInt(g).mapToObj(f)`) — `209` folds the pair into a boxing distill, which puts a `Wrapper.valueOf` in front of the mapper.

**`2xx/244-transform-object-to-primitive-after-distill`** is the sibling for that last case, the distill-headed twin of `243`: `243`'s `𝜏-first` is spelled out as a `Φ.hone.map`, so it never looks at the one pragma in the 2xx phase that is already a distill.

**`5xx/510-distill-parked-frame`** fixes a latent defect the crossing exposes. A capture peel (`316`, `314`) parks the item mid-body with `astore 1`, while every frame in the `503`/`504`/`510` family names that slot from `bridge-input`; once anything ahead of the peel retypes the item, the two disagree and the class fails verification with *"Type 'java/lang/Double' ... is not assignable to 'java/lang/Integer'"*. It reproduces with no crossing involved —

```java
Stream.of(1, 2, 3).map(n -> n / 2.0).map(d -> d * frac)
  .peek(...).filter(d -> d > 1.0)
```

— and seed 53 of `RandomPipeline` (`PROSE.chars().peek(...).mapToObj(n -> n / 2.0).map(d -> d * frac)...`) is what found it. The parked slot is widened to `java/lang/Object`, which is all the merge needs; the item on the stack is left exactly as the frame writer named it.

## Tests

* `src/test/phino/208-mapToObj-to-boxed-map.yml`, `244-...yml`, `510-...yml` — one pack per rule.
* `optimize/streams/maptoobj-crossing.yml` — the `Int`/`Long`/`DoubleStream` crossings plus the non-head `mapToInt(...).mapToObj(...)` shape; invokedynamic 13 → 6.
* `optimize/streams/parked-retyped-item.yml` — the frame defect, reached both with and without a crossing; invokedynamic 8 → 2.
* `optimize/streams/all-ops.yml` — `after` counts re-measured: its `mapToInt/mapToObj/mapToLong` run now collapses further, invokedynamic 14 → 12.
* README: the `#984` entry is removed from "Which Sink-Free Shapes Are Not Fused Yet" (four → three), and stage 2 gains the crossing and the `241`–`244` conversion family.

## Verification

**`deep.yml` triggers only on push to `master`**, so the fixture packs and the `RandomPipeline` fuzz — the tests that actually exercise these rules — do not run on this PR. They were replicated by hand instead, against the pinned `phino` 0.0.119 on the host (this environment has no Docker daemon):

* **all 60 packs under `optimize/streams/`** pass: every class prints exactly what it printed before optimization, and every declared opcode count matches. This is what caught the `all-ops` drift above.
* **all 120 `RandomPipeline` seeds** compile, optimize, run and print exactly what they printed before — 120 rewritten, 0 broken. Earlier revisions of this change broke 4 seeds and then 1; each rejection drove a design change here, and the last of them is what surfaced the `510` defect.
* `mvn test` (non-deep): 80 tests, 0 failures.
* the three new rule packs match their expected patterns.

https://claude.ai/code/session_01HHd1T6kjcBJL83kZrGYQ8q


yegor256:
The full fixture replay promised in the description has finished: **all 60 packs under `optimize/streams/` pass** — every class prints exactly what it printed before optimization, and every pack's declared opcode counts match.

It found one real drift, already pushed as 45e30f0: `all-ops.yml` has `.mapToInt(Integer::intValue).mapToObj(Integer::valueOf).mapToLong(Integer::longValue)`, and with `208` recognising the middle operator that run collapses further than it used to — invokedynamic 14 → 12, with the invokestatic and invokevirtual the extra fusion emits. That pack says of itself that its counts are a tally to re-measure rather than a design to defend.

Two packs report a missing log line, `closure-map-filter` and `rand`, both for `"No grep-in match for …"`. That is an artifact of how I replayed them rather than a finding: neither declares a `grep-in:` key, so the real test runs them under `OptimizeMojo.DEFAULT_GREP_IN` (which deliberately skips a class carrying neither `map` nor `filter`), while my replay forced `.*` and so never skipped anything. `rand.yml` exists precisely to assert that skip, and CI will exercise it properly.

Worth noting for whoever reviews: **`deep.yml` triggers only on push to `master`**, so the fixture packs and the `RandomPipeline` fuzz — the tests that actually exercise these rules — do not run on this PR. That is why the replay above was done by hand against the pinned `phino` 0.0.119 on the host, alongside all 120 fuzz seeds (120 rewritten, 0 broken) and `mvn test` (80 tests, 0 failures). All 18 CI checks are green on the current head.

---

PR #1009, opened 2026-09-18
Exclude pre-Java-16 classes instead of ending the build

### LINKED ISSUE #1007: The Java 16 guard in `entry.sh:194` exits the build on the first old class, so `self` fails on every master commit
The guard #968 added at `src/main/resources/org/eolang/hone/scaffolding/entry.sh:187-197` walks every class under the target directory and calls `exit 1` on the first one whose major version is below 60. The `self` profile selects `streams/*` at `pom.xml:387`, so the guard inspects this project's own classes, and `HelpMojo.class`, which `maven-plugin-plugin` generates at Java 8 level, is version 52. The `self` workflow has failed on every master commit since `bad0b8a4` landed on 2026-09-17.

Refusing to emit `mapMulti` into an old class is right, but refusing to build anything beside it is not: one class the rules cannot serve now stops the whole project, and `HelpMojo.class` is generated boilerplate nobody chose. The same guard aborts `src/it/bench` and `src/it/modular`, which set `maven.compiler.target` to 1.8 (`src/it/bench/pom.xml:15`, `src/it/modular/pom.xml:20`) while selecting `streams/*`. Nothing in CI reports those two, because `.github/workflows/mvn.yml:39` runs with `-Dinvoker.skip`.

`entry.sh` already has a way to keep a class out of the pipeline: `EXCLUDES` reaches jeo as `-Djeo.disassemble.excludes` at line 238. Adding an offending class to that list, and naming it in the log, would keep the guarantee #930 asked for without ending the build.

yegor256:
The bytecode guard added in #968 walks every class under the target directory and calls `exit 1` on the first one below major version 60. The `self` profile selects `streams/*` over this project's own classes, and `maven-plugin-plugin` generates `HelpMojo.class` at Java 8 level, so the whole build died on one generated class nobody chose. `self` has failed on every master commit since `bad0b8a4`. The same guard aborts `src/it/bench` and `src/it/modular`, both of which set `maven.compiler.target` to 1.8 while selecting `streams/*`.

Refusing to emit `mapMulti` into an old class is right; refusing to build anything beside it is not. The guard now gathers the offending classes instead of exiting, appends them to `EXCLUDES` (keeping whatever the user already set), logs each one with its version, and carries on. `EXCLUDES` already reaches jeo as `-Djeo.disassemble.excludes` further down the same script, so those classes are never disassembled and never get a `mapMulti` — what #930 asked for still holds — while everything else is rewritten as before.

I reproduced it first with fixture classes at versions 52 and 69: the old guard printed its refusal and exited, so the two modern classes were never optimized. The new one was then exercised on four cases — an old class present, a user-supplied `EXCLUDES` already set, no old classes at all, and two old classes. The pattern format seemed worth checking rather than assuming, since `DisassembleMojo` hands these to `GlobFilter`, which compiles each with `getPathMatcher("glob:" + pattern)`: a literal absolute path matches itself, does not match a sibling in the same package, and an inner class `$` is not a glob metacharacter.

Two things for you to weigh. `.github/hone-it.sh` still has the whole-module `supports_streams` skip that came in with #968, which now overlaps with per-class exclusion — a module with one old class is skipped entirely there, where this would have excluded just that class. I left it alone to keep the change small, but the two policies may be worth collapsing. And there is a `@todo #1007:30min` puzzle noting that `mvn.yml` passes `-Dinvoker.skip`, so neither integration project can turn a build red; that is why this went unnoticed.

Closes #1007


yegor256:
@yegor256 could you take a look when you have a moment? Happy to explain any of it, particularly the call to leave the module-level skip in hone-it.sh alone rather than folding it into this.

Issue #635, opened 2026-06-01
we don't fuse operations with objects as predicates


yegor256:
For example, this doesn't fuse:

```java
Stream.of(1, 2, 3).map(new Foo()).count();
```

VasilevNStas:
@yegor256 since this is assigned to you and nothing has moved yet, here is the analysis we did on this problem a while ago — hopefully it saves you some digging. We'd be happy to help implement it if you want a hand.

## The gap

Every `2xx` recognition rule (`201`-`208`) matches a `Φ.hone.lambda` formation (produced by `111` from an `invokedynamic`) followed by the stream call. An **object predicate** — `Stream.of(1,2,3).map(new Foo())` — compiles to:

```
new Foo
dup
invokespecial Foo.<init>()V
invokeinterface java/util/stream/Stream.map (Ljava/util/function/Function;)Ljava/util/stream/Stream;
```

There is no `invokedynamic`, hence no `Φ.hone.lambda`, and the operation is invisible to the whole pipeline. Nothing in `streams/` matches a `new`/`invokespecial` today (grep over all `.phr` files), so the gap is total, not partial.

## Proposed design

A front-door rule in `1xx` — the free slot `115-` (between `114` and `118`), mirroring `111` — that matches the four-opcode run `new + dup + invokespecial <init>` + `invokeinterface Stream.map|filter` and abstracts it into a `Φ.hone.map` / `Φ.hone.filter` pragma. Because the object's type is unknown, synthesise the lambda signature as `(Ljava/lang/Object;)Z` / `(Ljava/lang/Object;)Ljava/lang/Object;` and let the existing sed-fallbacks (`201`'s `𝑒-java-object-pattern`) carry the `Object` element type downstream. From there the pragma rides the same `3xx` → fusion → `mapMulti` machinery.

## Pitfalls we found

1. **`closures.yml` pins the `named` case as "survive untouched"** — `filter(new Good()).map(new Doubler())` (line ~147), `after` `invokedynamic 27`. Fusing it adds a mapMulti dispatch lambda, so `invokedynamic` becomes `28`; the fixture must be updated with the feature.
2. **`RandomPipeline` cannot generate object predicates** — every stage production in the grammar is a lambda/method-ref, and `quarantined()` is empty. A production like `map(new Doubler())` needs a `Function`/`Predicate`-implementing class in the walk template (today it only has `Pair`).
[...]

PR #1011, opened 2026-09-18
Recognize a map or a filter whose object argument captures

### LINKED ISSUE #1008: 215-recognize-named-map.phr:40-49: Fuse a map whose...
The puzzle `635-8c54bfc9` from #635 has to be resolved: 

https://github.com/objectionary/hone-maven-plugin/blob/d808969c638f6d3362cf6e03d56e5020c9611190/src/main/resources/org/eolang/hone/rules/streams/2xx/215-recognize-named-map.phr#L40-L49

The puzzle was created by @yegor256 on 18-Sep-26. 

Estimate: 60 minutes,  role: DEV. 

yegor256:
Closes #1008, the puzzle `635-8c54bfc9` that #635 left behind.

The two `215-` recognisers pinned the constructor descriptor to `"()V"` and took the constructor run to be exactly four adjacent opcodes, so a `stream.map(new Scaler(2))` was invisible to the whole pipeline: no pragma, no fusion, the operation just stayed native. They now bind the run of pushes between the `dup` and the `invokespecial` as a block and the descriptor as an expression, and carry both on the pragma; the two `317-*-to-distill` rules splice them back between the `new` and the `<init>` of the state append.

One thing worth arguing about before you read the diff: nothing is peeled. The ticket suggested peeling the pushes the way `114` peels a capturing lambda's, and I think that is the wrong treatment. A lambda's captures are handed to `lambda$` on every element, which is why they have to go into the shared list one at a time. A constructor argument is evaluated once, when the pipeline is built, and `502` relocates the whole append back into the very method the pushes came from, where their locals are still in scope. So carrying the run whole is both simpler and the semantically right thing. Binding the object name on both the `new` and the `<init>` owner is what now proves the two belong together, and it settles a nested `new Foo(new Bar())` for free, since the split that would pair `new Foo` with `Bar.<init>` cannot bind one name to two.

The second choice is the `when` guard, which bounds the arguments to two JVM slots. That is a stack budget, not a limit of the shape. `502` bumps the caller's max-stack by a fixed 6, sized for 308's append; this one peaks at 4 with no arguments, so two more slots land exactly on what is already guaranteed and a third would blow it — the class would die at load with "Exceeded max stack size" rather than fail a rule, so the guard stands the rule down instead. One reference, one int, one long, one double or two one-slot arguments all pass. Widening it means teaching `502` to bump by the append it actually relocated instead of by a constant, which is a change to `502` and not to these rules.

I reproduced it first: the capturing input yields no pragma against master's own `215` and one against the branch's. The guard was checked against real descriptors rather than reasoned about — `()V`, `(I)V`, `(J)V`, `(Ljava/lang/String;)V` and `(II)V` are recognised, `(III)V`, `(JI)V` and `(JD)V` are not. The `215`-to-`317` chain produces `dup; new Scaler; dup; bipush 2; invokespecial Scaler.<init>(I)V; List.add; pop`. Three new packs, four existing ones updated for the two new pragma fields; `mvn -Pdeep test` on the pack test, `RulesTest` and `OptimizeMojoRuleNameTest` gives 135 tests and no failures, against 132 on master.

A narrower `@todo #1008:45min` replaces the old `@todo #635`. The old puzzle bundled in the operator that arrives as a method argument or a field read (`filter(p)`, the refArg case in `streams/closures.yml`), but that is a different shape rather than a wider one — there is no `new` to anchor on and the lone `aload` cannot be told apart from the receiver stream's own, so the anchor has to come from the descriptor of the stream call. The over-two-slot constructor is named in the same puzzle. The README's "not fused yet" list is updated to match: the entry that said a capturing object argument is not fused now says a constructor wider than two slots is not.


yegor256:
@yegor256 could you take a look when you have a moment? The two calls I would most like a second opinion on are relocating the constructor pushes instead of peeling them, and the two-slot guard; happy to explain either in more detail if anything reads oddly.

Issue #835, opened 2026-08-14
The 82 phino rule packs are never executed by any CI workflow


VasilevNStas:
## Problem

The single-rule verification packs in `src/test/phino/*.yml` (82 files) are never actually executed by any CI workflow. They depend on three conditions that no workflow satisfies together:

1. **The test is `@Tag("deep")`** — `OptimizeMojoTest.appliesPhinoRulesAsSpecifiedInYamlPack` (`src/test/java/org/eolang/hone/OptimizeMojoTest.java:167-172`). The default surefire config excludes `deep` (`excludedGroups=deep`), so `mvn clean install` in `mvn.yml:41` (without `-Pdeep`) skips it.
2. **The test requires phino of exactly the pinned version** — `@DisabledWithoutPhino` checks `new Phino().available(expected)` where `expected` comes from `default-phino-version.txt` (currently `0.0.106`) and `Phino.available()` does an exact string comparison (`DisabledWithoutPhinoCondition.java:38`). `mvn.yml:29` downloads **`phino-latest`**, which is not `0.0.106`, so even the version probe would disable the test.
3. **The `deep` workflow installs no phino at all** — `deep.yml` runs `mvn clean install -Pdeep` (`deep.yml:72`) but its steps are only checkout, setup-java, install-maven and cache (`deep.yml:38-71`). Unless the EC2 AMI (`secrets.AWS_AMI`) happens to carry phino `0.0.106`, `@DisabledWithoutPhino` disables every pack.

Net result: the 82 packs (plus the no-Docker deep tests guarded by `@DisabledWithoutPhino`) run only if the undocumented EC2 AMI happens to contain the right phino binary. No workflow guarantees it. The only workflow that correctly pins the phino version — `make.yml:26-29` (`phino-${version}`) — is itself a no-op (see #834).

## Expected behavior

The workflow that runs `-Pdeep` must install the pinned phino first, mirroring `make.yml:26-29`:

```sh
version=$(xargs < src/main/resources/org/eolang/hone/default-phino-version.txt)
sudo wget -q -O /usr/bin/phino "http://phino.objectionary.com/releases/ubuntu-24.04/phino-${version}"
sudo chmod +x /usr/bin/phino
phino --pin="${version}" --version
```

so that `@DisabledWithoutPhino` actually enables the packs on the deep runner. Alternatively, `mvn.yml` could run the packs by pinning phino and invoking `mvn test -Pdeep` with the pack test selected.

## Details

[...]

PR #1043, opened 2026-09-19
Install GNU Parallel on the deep runner and match the timings row it writes

### LINKED ISSUE #1042: The `deep` runner has no GNU Parallel, so every host-phino test dies in `rewrite.sh:417`
The `deep` workflow runs on an EC2 runner whose AMI carries neither Maven nor phino, so `.github/workflows/deep.yml:52` and `.github/workflows/deep.yml:81` install them explicitly. Nothing installs GNU Parallel. Once phino sits on the PATH, every test that optimizes without Docker takes the host path, and `src/main/resources/org/eolang/hone/scaffolding/rewrite.sh:417` exits 1 when `threads` is above one and `parallel --version` fails; `src/main/resources/org/eolang/hone/scaffolding/entry.sh:239` prints the same `parallel: command not found` a moment earlier.

The result is that 79 tests error out in [run 35433203833](https://github.com/objectionary/hone-maven-plugin/actions/runs/35433203833/job/105871585357), and the same message appears in every recent `deep` run, so the suite has not been green since #835 put phino on that runner. The GitHub-hosted runners behind `mvn.yml` and `smoke.yml` ship GNU Parallel, which is why only `deep` is red. The refusal in `rewrite.sh` is deliberate, as #409 settled, so the runner has to supply the tool rather than the script work around its absence.

Add an `install-parallel` step to `.github/workflows/deep.yml`, alongside `install-maven` and `install-phino`, that runs `sudo apt-get install --yes parallel` when `command -v parallel` finds nothing.

yegor256:
The `deep` job installs GNU Parallel on the EC2 runner now, and the JNA test looks for the timings row in the shape the plugin actually writes.

Both are what made [run 35433203833](https://github.com/objectionary/hone-maven-plugin/actions/runs/35433203833/job/105871585357) red. Once #835 put phino on that runner, every test that optimizes without Docker started rewriting on the host, and `rewrite.sh` exits there when GNU Parallel is missing, which it is on that AMI — 79 tests died with `parallel: command not found`. The single test that got past that point, `optimizesJustOneLargeJnaClass`, asked for `optimize,123` in `hone-timings.csv`, but #922 changed the row to `"optimize";123`, so it failed too.

I ran that JNA test locally against Docker and it passes with the new pattern; the workflow step I could only lint. One thing left over from the same run: `generatesStatisticsWithDocker` timed out after 600s because the cold `docker build` on a fresh instance ate the whole budget. It passed in the runs before and after, so I left it alone, but the margin there is thin.

Closes #1042


yegor256:
@yegor256 mind taking a look when you get a chance? Happy to explain anything here.

PR #1049, opened 2026-09-19
Keep both statistics tests inside their budgets on the deep runner

### LINKED ISSUE #1048: Both tests of `OptimizeMojoStatisticsTest` hang the deep build, for two unrelated reasons
The deep suite is red again in
[run 35438436667](https://github.com/objectionary/hone-maven-plugin/actions/runs/35438436667):
`generatesStatisticsWithoutDocker` times out after 180s and
`generatesStatisticsWithDocker` after 600s, while every other test in the run
passes. Two separate defects meet in that one class.

`generatesStatisticsWithoutDocker`
(`src/test/java/org/eolang/hone/OptimizeMojoStatisticsTest.java:40`) sets
`debug=true`, which runs `entry.sh` and `rewrite.sh` under `set -x`. That
trace lands on the script's stderr, and `OptimizeMojo.java:714` runs the
script through `Jaxec.exec()`, which does not read the child's pipes until the
child has exited (yegor256/jaxec#165). Once the trace fills the 64 KiB pipe
buffer the script blocks in `write()` and neither side ever moves again. The
fixture is right on the edge: its two classes produce 61,875 bytes of trace,
and 239 of those bytes' worth of lines name the project directory, so the
runner's longer checkout path is enough to cross the line. On a real project
`debug=true` would cross it far earlier.

`generatesStatisticsWithDocker` (same file, line 50) has `@Timeout(600)` but
pays for a cold `docker build` whenever it is the first test in the run to
force Docker. `BuildMojo.java:68` returns early when host `phino` is on the
PATH, so every test that does not set `alwaysWithDocker` skips the image
build, and the image is built from scratch — GHC and `cabal install phino` —
by whichever forcing test surefire happens to schedule first. In the earlier
attempt of the same run that test was
`OptimizeMojoRulesTest.optimizesJustOneLargeJnaClass`, which carries no
`@Timeout` at all, took 858.3s and left the statistics Docker test to finish
in 12s. In the later attempt the statistics class ran first and the same build
blew its 600s budget.

Give `generatesStatisticsWithDocker` the 1200s budget the other forced-Docker
test already has (`OptimizeMojoTest.java:131`), and keep
`generatesStatisticsWithoutDocker` under the pipe limit by dropping
`debug=true`, which none of its assertions need, until jaxec is fixed.

yegor256:
Two small changes to `OptimizeMojoStatisticsTest`. The host-path test no
longer passes `debug=true`, and the Docker test's timeout goes from 600s
to 1200s, the same budget the other forced-Docker test already has.

Both tests died in the last deep run and they died for different reasons.
With `debug=true` the scaffolding traces itself with `set -x`, which is
over 60 KiB of stderr for the two classes in that fixture, and jaxec
reads none of a child's output until the child has exited
(yegor256/jaxec#165) — so the script fills the 64 KiB pipe, blocks in
`write()`, and the build hangs forever. The measured trace was 61,875
bytes on my laptop, just under the limit, and the runner's longer
checkout path is what pushes it over. The Docker test, meanwhile, pays
for a cold `docker build` whenever surefire happens to schedule it before
`optimizesJustOneLargeJnaClass` — every other test skips the image build
now that phino sits on the host PATH — and building GHC and phino from
source does not fit in ten minutes.

The `debug=true` removal is temporary and carries a puzzle: once jaxec is
fixed and bumped here, the option goes back, since nothing else covers
debug mode on the host path. Worth knowing that the same deadlock is
waiting for any user who turns `<debug>true</debug>` on for a real
project.

Closes #1048


yegor256:
@yegor256 could you have a look when you get a chance? Glad to explain either half of it in more detail.

yegor256:
Updated: jaxec 0.6.1 is out with the fix, so this now bumps the dependency instead of turning debug off, and the puzzle is gone.

